PR #6, opened 2015-02-26
Update ListBox.java

mearing:
Added a setSelectedValue(String value) method to ListBox. (Copied from gwt-boostrap).


tbroyer:
Please use Gerrit for contributions to GWT: http://www.gwtproject.org/makinggwtbetter.html

PR #8, opened 2015-04-05
Prevents multiple loading of script from the same url in ScriptInjector

64BitAsura:


dankurka:
Please use gerrit for contributions to GWT source, see: http://www.gwtproject.org/makinggwtbetter.html

user/src/com/google/gwt/core/client/ScriptInjector.java
```diff
@@ -265,6 +272,7 @@ function clearCallbacks() {
       }
     }
     scriptElement.onload = $entry(function() {
+      scriptRegistry[scriptElement.src] = scriptElement;
```
JamesXNelson: Pretty sure you would want to set this before it finishes loading, otherwise there will be a race condition if you add the same script again before the first one finishes loading.

user/src/com/google/gwt/core/client/ScriptInjector.java
```diff
@@ -122,6 +122,7 @@ public FromString setWindow(JavaScriptObject window) {
     private boolean removeTag = false;
     private final String scriptUrl;
     private JavaScriptObject window;
+    private static JavaScriptObject scriptRegistry = JavaScriptObject.createObject();
```
JamesXNelson: Make this final

JamesXNelson:
As Daniel said, use Gerrit.  This is just a mirror for people who just want to check out and build.

Also, although the changes look fine, please consider that some people may actually want to inject the same script multiple times.  One example would be a JSON-P callback; maybe I will add `<script src="/context/getUpdates" />` whenever I want to check for updates after performing some other operation.

Thus, the prevention of duplication needs to be explicit and optional; changing default behaviour like this will almost always break somebody's code.

PR #9211, opened 2015-11-06
Implement SafeUriString#toString

aabdagic:
Implement SafeUriString#toString because if a Mockito verify or an assertion fails, you'd want to get an actionable message. Also, it's a good idea to have toString anyways.

[Review on Reviewable](https://reviewable.io/reviews/gwtproject/gwt/9211)

### DIFF
--- a/user/src/com/google/gwt/safehtml/shared/SafeUriString.java
+++ b/user/src/com/google/gwt/safehtml/shared/SafeUriString.java
@@ -75,4 +75,9 @@ class SafeUriString implements SafeUri {
   public int hashCode() {
     return uri.hashCode();
   }
+
+  @Override
+  public String toString() {
+    return asString();
+  }
 }


jDramaix:
We use Gerrit for code contributions: http://www.gwtproject.org/makinggwtbetter.html#contributingcode
Would you mind sending the patch there?

user/src/com/google/gwt/safehtml/shared/SafeUriString.java
```diff
@@ -75,4 +75,9 @@ public boolean equals(Object obj) {
   public int hashCode() {
     return uri.hashCode();
   }
+
+  @Override
+  public String toString() {
+    return asString();
```
tbroyer: Please don't make `toString` a synonym for `asString`; rather copy the behavior from `SafeHtmlString`.

(but as said by Julien, please don't update the PR and send a patch on Gerrit instead).

PR #9481, opened 2017-01-21
SimplePager fastForward is sometimes clickable when disabled

francescoforesti:
The setFastForwardDisabled method in SimplePager changes the style of the button without actually calling setDisabled on the button. The result is that sometimes the button is greyed out, but can still be clicked.

---
This change is [Reviewable](https://reviewable.io/reviews/gwtproject/gwt/9481)


tbroyer:
Thank you for your contribution but we don't use GitHub pull requests: http://www.gwtproject.org/makinggwtbetter.html

Before you go further with this change, I think it should also add non-regression tests to `SimplePagerTest`, maybe also add tests for the "last page" button.

PR #9516, opened 2017-04-21
Deleted unnecessary line/comment and added comment

thapapinak:
I just added additional feature and ###### Known workarounds(#) in the template issues . Deleted unnecessary lines present.

---
This change is [Reviewable](https://reviewable.io/reviews/gwtproject/gwt/9516)


tbroyer:
OK, so, I reported the account to GitHub for abuse, and will block it from the organization if we see a fourth PR (with the same or different changes).

PR #9575, opened 2017-11-29
Fix 'GWT' headline not being rendered properly

sebkur:
Okay, I get it, you're not using PRs. Anway, I would like to avoid setting up the whole toolchain for submitting a patch to the project for fixing a single broken whitespace in the Readme file. Would be nice if this could be fixed one way or another. Thanks

---
This change is [Reviewable](https://reviewable.io/reviews/gwtproject/gwt/9575)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-## GWT
+## GWT
 
   [![nightly](https://img.shields.io/jenkins/s/http/build.gwtproject.org/gwt.svg?label=nightly)](http://build.gwtproject.org/job/gwt/)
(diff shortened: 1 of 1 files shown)


sebkur:
To fix this yourself, just delete the space in between '#' and 'GWT' on the first line and insert a new space character there

niloc132:
Well the good news is that you don't need the toolchain (except for the CLA and the single commit hook) just to clone from git and push to gerrit ;) but I get your point - I'll submit the change directly.

PR #9580, opened 2017-12-13
Return all resources that match a default suffix.

chris-malloy-mri:
This is more of a question, so I'm creating a pull request here just to get some feedback.

I'm thinking that this code should return all resources that match a default suffix. Most of the built in generators already check that the length returned by this function is exactly 1, since they are expecting to embed only a single resource. However, for building custom resource loaders that need to determine the proper suffix at runtime (i.e. mp3 vs ogg), we would need to make the attached change.

Is there another way to accomplish what I'm trying to do, or is this a good change?

---
This change is [Reviewable](https://reviewable.io/reviews/gwtproject/gwt/9580)

### DIFF
--- a/user/src/com/google/gwt/resources/ext/ResourceGeneratorUtil.java
+++ b/user/src/com/google/gwt/resources/ext/ResourceGeneratorUtil.java
@@ -432,6 +432,7 @@ public final class ResourceGeneratorUtil {
 
     if (resourceAnnotation == null) {
       if (defaultSuffixes != null) {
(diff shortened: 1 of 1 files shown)


niloc132:
That does look reasonable to me - as an issue it could be okay to discuss here, but at least at present, we'll have to go through http://gwt-review.googlesource.com/ to get the source.

I'll leave it open for the moment as discussion, but please create an account and complete the CLA?

chris-malloy-mri:
OK in that case I will get set up on gerrit.

PR #9636, opened 2018-10-23
Create LICENSE FILE

berroteran:
The license file is very important.

---
This change is [Reviewable](https://reviewable.io/reviews/gwtproject/gwt/9636)

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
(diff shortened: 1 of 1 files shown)


jnehlmeier:
Thankful contributing but we don't use GitHub pull requests: http://www.gwtproject.org/makinggwtbetter.html

GWTs license isn't on the root of the project but in 
https://github.com/gwtproject/gwt/blob/master/distro-source/core/src/COPYING 
which is used to build the SDK.

Issue #9756, opened 2022-06-22
Migrate GWT source control to GitHub


niloc132:
With the GWT 2.10.0 release finalized, and no longer publishing artifacts on with a `com.google.*` groupId, we're able to finish moving the GWT 2 toolkit development entirely to GitHub. Issues have been managed here for several years, and while builds haven't been publicly visible for some time, they have been running until the release process began. This ticket is to track and discuss remaining work to migrate to GitHub.

- [x] File a ticket with github support to disable the "Public Mirror".
- [x] Enable Github Actions (and set up a build badge on the readme) for nightly builds, pull requests.
  - [x] Set up snapshots for nightly builds (GitHub packages are not a viable option, as this requires authentication to read, Sonatype Snapshots are probably best?)
  - [ ] Optional: Always add PR comments instead of diff annotations in the case of checkstyle violations (see https://github.com/gwtproject/gwt/issues/9768 for this)
- [x] Update pull request template with basic guidelines for developing on github first.

niloc132:
I'm intending to close this, with a separate follow-up ticket for posting PRs about style issues, unless anyone has any other thoughts on tasks that need to be done to finish migration? I think we've covered everything listed at https://docs.google.com/document/d/10Cn0Z7SDv0ZW5OVZdoIdw5EAEDUbtK6vracOOnxj7po/edit.

tbroyer:
👍 for closing this.

Wrt the doc, did Google already delete (or ask Sonatype) the snapshots? If not, in what timeframe do we want to have it done? Should we open an issue to track it?

niloc132:
The HEAD-SNAPSHOT builds in google-snapshots have not yet been deleted, though I have requested this, as they are going to be stale soon and not reflect other changes being made.

Closing this issue, https://github.com/gwtproject/gwt/issues/9768 will follow up for the remaining task.

niloc132:
HEAD-SNAPSHOTs for com.google artifacts have been deleted.

PR #9762, opened 2022-07-14
Enable GitHub Actions for pushes, nightly builds, and PRs

niloc132:
This introduces two builds, intended to fairly closely reflect the
builds that used to be run on the build.gwtproject.org Jenkins instance.

Nightly builds run most tests, and provide both a simple visualization
of the test summary as well as access to download all test reports, and
in the case of successful builds, the sdk zip can be downloaded. Maven
artifacts are not yet available. This full build can also be run by
pushing to main or release branches, and can be run by creating branches
that end with "-test", which may be helpful for forks to run their own
actions.

On each Pull Request, only checkstyle and apicheck will be run, in
addition to verifying that the SDK and samples can be built, but the ~2
hours of tests are not run. Any checkstyle errors that overlap with the
git diff will be posted as comments or annotations to the diff.

Partial #9756

### DIFF
--- a/.github/workflows/quick-check.yml
+++ b/.github/workflows/quick-check.yml
@@ -47,10 +47,11 @@ jobs:
           REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           set -eux
+          shopt -s extglob
           cd gwt
-          for xml in $(find build/out | grep checkstyle); do
-            echo $xml
-            reviewdog -f=checkstyle -filter-mode=diff_context -reporter=github-pr-review -level=info < $xml
+          for f in build/out/**/checkstyle*.xml ; do
+            echo $f
+            reviewdog -f=checkstyle -filter-mode=diff_context -reporter=github-pr-review -level=info < $f
           done
       - name: Upload checkstyle xml for manual review
         uses: actions/upload-artifact@v2
--- a/.github/workflows/quick-check.yml
+++ b/.github/workflows/quick-check.yml
@@ -1,0 +1,61 @@
+# Run simple checks to make sure that the build is sane when any branch is pushed or pull
+# request created. Comments or annotations will be added to a pull request in the case of
+# style guide violations. Presently, tests are not run on pull requests.
+name: Quick smoke test
+on: [push, pull_request]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout GWT itself into one directory
+        uses: actions/checkout@v2
+        with:
+          path: 'gwt'
+          # we need depth=2 to see which style violations overlap with the current changes
+          fetch-depth: 2
+      - name: Checkout GWT tools into a sibling directory
+        uses: actions/checkout@v2
+        with:
+          repository: 'gwtproject/tools'
+          path: 'tools'
+      - name: Set up JDK 8
+        # GWT presently requires Java8 to build, but can run on newer Java versions
+        uses: actions/setup-java@v2
+        with:
+          java-version: '8'
+          distribution: 'temurin'
+      - name: Set up reviewdog for easier checks on the PR's checkstyle output
+        uses: reviewdog/action-setup@v1
+        with:
+          reviewdog_version: latest
+
+      - name: Build, style/api checks, produce docs
+        # Presently this runs no tests at all, but could run quick tests
+        run: |
+          set -eux
+          cd gwt
+          # Set env vars to ensure we get the build we expect
+          export \
+            TZ=America/Los_Angeles \
+            ANT_OPTS=-Dfile.encoding=UTF8 \
+            ANT_OPTS=-Xmx2g
+          ant clean dist doc checkstyle apicheck
+
+      - name: Create pull request comments/annotations for checkstyle, even on failure
+        if: always() && github.event_name == 'pull_request'
+        env:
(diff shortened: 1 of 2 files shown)


niloc132:
Example of pull request comments: https://github.com/niloc132/gwt/pull/1. Note that without the "write" permission, pull requests will see annotations on the diff instead of comments. This can be dealt with securely by producing an artifact with the diff and checkstyle XMLs, and creating an additional job which runs on workflow_run events, and picks up those artifacts. Alternatively, a label could be required to run the quick build at all, and only permit maintainers to add that label once they visually inspect the build. See https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ for more discussion.

Example of a full build in a fork, on a `*-test` branch: https://github.com/niloc132/gwt/actions/runs/2639485313.

.github/workflows/quick-check.yml
```diff
@@ -0,0 +1,61 @@
...
+        if: always() && github.event_name == 'pull_request'
+        env:
+          REVIEWDOG_GITHUB_API_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          set -eux
+          cd gwt
+          for xml in $(find build/out | grep checkstyle); do
+            echo $xml
+            reviewdog -f=checkstyle -filter-mode=diff_context -reporter=github-pr-review -level=info < $xml
+          done
```
tbroyer, line 54: Maybe use `-name 'checkstyle*.xml'` in `find` command rather than pipe to `grep`?

BTW, maybe use `find`'s `-execdir` too rather than a loop? (and use reviewdog's `-tee` rather than the separate `echo $xml`)
```suggestion
          find build/out -name 'checkstyle*.xml' -execdir sh -c 'reviewdog -f=checkstyle -filter-mode=diff_context -reporter=github-pr-review -level=info -tee < {}' \;
```

tbroyer (approved):

.github/workflows/quick-check.yml
niloc132: Good suggestions - I find `find` to be a bit too much rolled into one for my brain when I am thinking in bash, I prefer smaller building blocks for my code, but I think I agree it would be more readable.

I believe that `reviewdog`'s `-tee` flag does not do what you are expecting here - I am writing (`echo`, not `cat`) the name of the file for debugging purposes, then passing `$xml` to stdin here, but `-tee` will write to stdout with the changes it would like to suggest _in addition to_ passing them to the reporter. Among other things, `-tee` will effectively `cat` the file, which is not what I was hoping to do, but to let us see what files are seen by this loop, in what order, and to see what output each produced from `reviewdoc`. See https://github.com/reviewdog/reviewdog#debugging

So from that perspective, of wanting to put both `echo` and `reviewdog` in the `-execdir`, maybe the loop does make more sense? 

Note that `-tee` is very verbose, since our checkstyle files are gigantic, and they are made available as a build artifact, so catting them shouldn't be necessary, but knowing what each command _might_ provide benefit.

tbroyer: Oops, yes indeed I mixed up `echo` and `cat` :tired_face: (I'll blame the heat wave :hot_face:)

BTW, would that `echo` be really useful when using `set -x`? Wouldn't the shell then print the file name on the reviewdog command line? Maybe use `-exec` rather than `-execdir` (even though it's discouraged) to have the path in the `{}` in addition to the file name?

How about then:
```suggestion
          find build/out -name 'checkstyle*.xml' -exec bash -eux -c 'reviewdog -f=checkstyle -filter-mode=diff_context -reporter=github-pr-review -leve=info < "{}"' \;
```
or possibly:
```suggestion
          shopt extglob
          for f in build/out/**/checkstyle*.xml ; do
            reviewdog -f=checkstyle -filter-mode=diff_context -reporter=github-pr-review -leve=info < "$f"
          done
```

Just suggestions, but the current version is good as-is.

niloc132: At least as written, the `... < $xml` does not get logged with the `set -x`, see the sample output at https://github.com/niloc132/gwt/runs/7258451989?check_suite_focus=true#step:7:20 (note that this still had `-tee` in it):
```
 + for xml in $(find build/out | grep checkstyle)
+ echo build/out/dev/checkstyle_tests_log.xml
+ reviewdog -tee -f=checkstyle -filter-mode=diff_context -reporter=github-pr-review -level=info
build/out/dev/checkstyle_tests_log.xml
<?xml version="1.0" encoding="UTF-8"?>
<checkstyle version="6.0">
<file name="/home/runner/work/gwt/gwt/gwt/dev/core/test/com/google/gwt/core/ext/linker/ArtifactSetTest.java">
</file>
```

I haven't debugged further - I typically avoid redirections in my scripts, but it seems that `reviewdog` can only read from redirect. I guess I could do `cat $xml | reviewdog` (and I often would if no one else would look at it, but quite a few reviewers have found that at least mildly offensive in the past...). Since it even cuts off the `<` in the above log output, even with the `find` it wouldn't work (without resorting to cat and a pipe). With that all said, I'm inclined to stick with the explicit loop.

`shopt extglob` - did you want `shopt -s extglob` to set it there, or gathering debug info? or `-u` to unset it? I'm not very familiar with it, but from some quick reading suspect you intended `-s`. 

I'll make a manual commit with the -s added (assuming that to be correct), will fix the `-level`, and give it a test run on my test PR to confirm we still get output as expected.

tbroyer (approved):

PR #9763, opened 2022-07-24
Deploy nightly builds to sonatype, using github secrets

niloc132:
Configures nightly scheduled builds to push to sonatype snapshots, as
used to happen with the Google-managed Jenkins job. GitHub Actions
for the account is assumed to have SONATYPE_USERNAME and
SONATYPE_PASSWORD secrets configured. To avoid forks attempting to
deploy to org.gwtproject from their own actions, this is also set to
only run on the gwtproject organization's repository.

Partial #9756

### DIFF
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -80,7 +80,7 @@ jobs:
       - name: Set up sonatype credentials
         # Using the same java version as above, set up a settings.xml file
         uses: actions/setup-java@v3
-        if: github.event_name == 'schedule'
+        if: github.event_name == 'schedule' && github.repository_owner == 'gwtproject'
         with:
           java-version: '8'
           distribution: 'temurin'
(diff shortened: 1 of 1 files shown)


niloc132:
Note that these secrets are not yet set - I propose creating a specific Sonatype account for this purpose, to limit potential damage to other groupIds if the secrets somehow did leak. If there is a concern about that, I can also use my own credentials.

Testing was performed using my own fork and enabling this action without the repository check, see results at https://github.com/niloc132/gwt/runs/7485939306?check_suite_focus=true#step:11:23 and https://oss.sonatype.org/content/repositories/snapshots/org/gwtproject/gwt/ etc. Note that the com.google dependencies still use `project.version`, which will be an error in the future, but happens to still resolve before Google deletes their snapshots (which we will request, after snapshots are usable).

tbroyer (approved):

niloc132:
New sonatype account has been created and added to the gwtproject org (though only the gwt project is presently configured to use it). I have only validated that these credentials work to log in, but haven't tried to deploy - but sonatype's ticket confirms they should work. 

I did one more "nightly" test in my fork to confirm that a fork would not attempt to deploy, even with valid credentials.

PR #9766, opened 2022-07-25
Fix gzip encoding of responses in the CodeServer

### LINKED ISSUE #9765: Using GWT 2.10.0 and super dev mode the cache.js files are uncompressed
**GWT version: 2.10.0** 
**Browser (with version): all browsers** 
**Operating System: tested in Windows and Macos** 

---

##### Description
After upgrading from GWT 2.8.2. to 2.10.0 the cache.js files in Super Dev Mode are no longer served as gzip files. They are uncompressed instead.
This leads to higher load times of the application in SDM.

##### Steps to reproduce
Start Super Dev Mode in a GWT 2.10.0 project and check the network tab in your browser. Although the browser allows gzip, the file is transferred uncompressed. See the response in the browsers network tab.

We discussed this on gitter and @tbroyer shared this link: https://github.com/eclipse/jetty.project/blob/jetty-9.4.44.v20210927/jetty-servlets/src/main/java/org/eclipse/jetty/servlets/GzipFilter.java

tbroyer:
Also remove sessions' handling as they're not used there.

Fixes #9765

### DIFF
--- a/dev/codeserver/java/com/google/gwt/dev/codeserver/WebServer.java
+++ b/dev/codeserver/java/com/google/gwt/dev/codeserver/WebServer.java
@@ -30,7 +30,6 @@ import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
 import org.eclipse.jetty.servlet.ServletContextHandler;
 import org.eclipse.jetty.servlet.ServletHolder;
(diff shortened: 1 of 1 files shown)


niloc132:
Tested and confirmed locally - @nbartels can you double check this, and we'll land it?

zbynek (approved):
Works for me (Windows, Chrome)

PR #9767, opened 2022-07-26
Finish moving away from Google groupIds

niloc132:
Hardcodes the com.google groupIds for GWT to 2.10.0 going forward, and
switches from google snapshot repository to general sonatype snapshots
now that we have nightly builds deployed there.

### DIFF
--- a/maven/poms/gwt/pom-template.xml
+++ b/maven/poms/gwt/pom-template.xml
@@ -29,7 +29,7 @@
             <dependency>
                 <groupId>com.google.gwt</groupId>
                 <artifactId>gwt-user</artifactId>
-                <version>${project.version}</version>
+                <version>2.10.0</version>
             </dependency>
             <dependency>
                 <groupId>org.gwtproject</groupId>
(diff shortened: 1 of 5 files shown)


niloc132:
According to the pre-2.10 release checklist, this is the final set of changes to make before we consider gwt open for development on github.

tbroyer (commented):
Sounds good. I'd have to check there are no other locations that'd need to be changed that way. Feel free to merge if you're certain.

niloc132:
I searched for `google-snapshots` to ensure there were none (_except_ in the push-gwt.sh script, which likely should be removed eventually, along with the google-specific poms), and `groupId>com.google` to ensure they all a) occurred in pom.xml files, b) outside of maven/google-poms, with c) an explicit version set in the next few lines.

```
git ls-files | xargs grep google-snapshots
```
```
git ls-files | grep -v maven/google-poms/ | grep xml | xargs grep -B0 -A3 groupId\>com.google
```

tbroyer:
Just a FYI, those commands can be simplified:
> ```
> git ls-files | xargs grep google-snapshots
> ```

```
git grep google-snaphots
```

> ```
> git ls-files | grep -v maven/google-poms/ | grep xml | xargs grep -B0 -A3 groupId\>com.google
> ```

```
git grep -B0 -A3 groupId\>com.google -- ':!maven/google-poms/' '*xml*'
```
(I'd probably change it to pass `-F` to `git grep` so the `.` in the pattern is matched literally, and use `'*.xml'` rather than `'*xml*'` to only match XML files, possibly adding in `*.xmlsrc` to match the webAppCreator template POM)

PR #9769, opened 2022-07-29
Correctly write Long.toBinaryString as an unsigned value

### LINKED ISSUE #9764: Long.toBinaryString is broken in GWT 2.10.0
**GWT version:**  2.10.0
**Browser (with version):** Chrome 103.0.5060.134
**Operating System:** Windows 10

---

##### Description

In previous GWT versions the method `Long.toBinaryString` was broken for negative values, but now seems to be broken for values higher than `Integer.MAX_VALUE`.

##### Steps to reproduce

Try to call `Long.toBinaryString(3218270003L)` and you get a **negative** binary number.

niloc132:
A previous fix to negative longs incorrectly allowed some positive longs
that would be represented as negative ints to be printed as negative
binary values, instead of as unsigned values. This patch fixes that case
and adds tests to confirm that these edge cases are well represented.
This bug was introduced by 7616ceb5.

The binary, hex, and octal tests all fail before this change, the
toString test was just for completeness, but wasn't actually broken.

Additionally, getHighBits could be expensive, and a range check against
two constants (that easily fit in a "small long") should be cheaper at
runtime. This is not required for the fix, and I will validate this
performance assumption.

Fixes #9764

### DIFF
--- a/user/super/com/google/gwt/emul/java/lang/Long.java
+++ b/user/super/com/google/gwt/emul/java/lang/Long.java
@@ -237,7 +237,7 @@ public final class Long extends Number implements Comparable<Long> {
 
     int highBits = LongUtils.getHighBits(value);
     if (highBits == 0) {
-      return Integer.toString((int) value, radix);
+      return Integer.toUnsignedString((int) value, radix);
     }
 
     final int mask = radix - 1;
--- a/user/test/com/google/gwt/emultest/java/lang/LongTest.java
+++ b/user/test/com/google/gwt/emultest/java/lang/LongTest.java
@@ -254,6 +254,14 @@ public class LongTest extends GWTTestCase {
     assertEquals(
         "1111111111111111111111111111111111111111111111111111111111111111",
         Long.toBinaryString(-1));
+
+    // Test int MAX+1 and int -1, as both are positive longs, but negative ints
+    assertEquals(
+        "10000000000000000000000000000000",
+        Long.toBinaryString(((long) Integer.MAX_VALUE) + 1));
+    assertEquals(
+        "11111111111111111111111111111111",
+        Long.toBinaryString(4294967295L));
   }
 
   public void testToHexString() {
@@ -262,6 +270,8 @@ public class LongTest extends GWTTestCase {
     assertEquals("1234500000000", Long.toHexString(0x1234500000000L));
     assertEquals("fff1234500000000", Long.toHexString(0xFFF1234500000000L));
     assertEquals("ffffffffffffffff", Long.toHexString(-1));
+    assertEquals("80000000", Long.toHexString(((long) Integer.MAX_VALUE) + 1));
+    assertEquals("ffffffff", Long.toHexString(4294967295L));
   }
 
   public void testToOctalString() {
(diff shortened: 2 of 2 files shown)


niloc132:
I made a quick performance test of the range check vs high bit check, and found that while chrome saw small performance improvements (or at worst saw the same performance most of the time), firefox saw significantly worse performance by making this change, for longs that don't fit in int32. The losses by firefox seemed (from a small sample size) to be worse than the gains seen in chrome, so at least for now I'm voting for keeping the status quo.

In this sample GWT2 app, v1 is the current check that is done, and v2 is the attempted improvement. My theory was that it was cheaper to compare against constants rather than shift, even for large numbers, but that seems to not universally be true.

```
public class AppEntryPoint implements EntryPoint {
    static final int COUNT = 1_000_000;
    @Override
    public void onModuleLoad() {
        // small positive numbers
        makeTestButton("small positive", 1, Integer.MAX_VALUE);
        // medium positive numbers
        makeTestButton("medium positive", Integer.MAX_VALUE, 1L << 53);
        // large positive numbers
        makeTestButton("large positive", 1L << 55, Long.MAX_VALUE);
        
        // small negative numbers
        makeTestButton("small negative", Integer.MIN_VALUE, -1);
        // medium negative numbers
        makeTestButton("medium negative", -1L << 53, Integer.MIN_VALUE);
        // large negative numbers
        makeTestButton("large negative", Long.MIN_VALUE, -1L << 55);
    }

    private void makeTestButton(String buttonName, long min, long max) {
        Element button = document.createElement("button");
        button.append(buttonName);

        button.onclick = e -> {
            // generate the numbers
            long[] numbers = generate(min, max, COUNT);
            boolean[] results = new boolean[COUNT];

            double[] perf = new double[2];

            Promise.resolve((IThenable<Object>) null)
                    .then(ignore -> {
                        double start = performance.now();
                        for (int i = 0; i < COUNT; i++) {
                            results[i] = fitsInInt32v1(numbers[i]);
                        }
                        perf[0] = performance.now() - start;
                        return null;
                    })
                    .then(ignore -> {
                        double start = performance.now();
                        for (int i = 0; i < COUNT; i++) {
                            results[i] = fitsInInt32v2(numbers[i]);
                        }
                        perf[1] = performance.now() - start;
                        return null;
                    }).then(ignore -> {
                        document.body.append(buttonName, "v1: ", perf[0] + "ms ", "v2:", perf[1] + "ms ", Arrays.toString(results).length() + "");
                        document.body.append(document.createElement("br"));
                        return null;
                    });
            return null;
        };

        document.body.append(button);
    }

    private long[] generate(long min, long max, int count) {
        long[] result = new long[count];
        Random random = new Random();
        for (int i = 0; i < count; i++) {
            result[i] = min + (random.nextLong() % (max - min));
        }
        return result;
    }

    public static boolean fitsInInt32v1(long value) {
        return getHighBits(value) == 0;
    }

    public static int getHighBits(long value) {
        return (int) (value >>> 32);
    }

    public static boolean fitsInInt32v2(long value) {
        // test if no high bits are set, this is the largest int that can be
        // represented with no high bits set
        long negativeOneIntAsLongBits = (1L << 32) - 1;//in int, this is -1

        // verify that no high bits are set by comparing only using "small" longs
        return 0 <= value && value <= negativeOneIntAsLongBits;
    }
}
```

I'll roll back that attempted improvement, and mark this pull request as ready for review.

zbynek (approved):

PR #9773, opened 2022-09-21
Update chrome/firefox globals to newest versions

### LINKED ISSUE #9771: GWT generates
**GWT version:** 2.8.2, 2.9, 2.10
**Browser (with version):** Chrome 104
**Operating System:** MacOs

---

##### Description
Chrome has added new short keyword to window namespace: `Ink`. It's short enough that likely compiler might use it as a name. Last time globals were updated with Firefox 60 and Chrome 60, so ~4 years ago in 2018.

##### Steps to reproduce
Have project big enough to generate `Ink` variable by obfuscator, then run app in runtime and expect exception. Whatever is expected under `Ink`, browser class will be used.

##### Known workarounds

We can set XML property for additional reserved keywords:

```
<set-configuration-property name="js.identifier.blacklist" value="Ink" />
```

Previously issue happened with u2f in Firefox in 2018.

piotrl:
### Implementation detail
This change-set contains all updates of globals, both added and removed experimental classes based on [globals txt files](https://github.com/gwtproject/gwt/tree/main/dev/core/src/com/google/gwt/dev/js/globals).

Last time it was done in 2018, so the changes are significant. I manually smoke-tested removed symbols to indeed, not exist in window namespace anymore.

* Firefox 104.0.1
* Chrome 105.0.5195.125

##### Issue context
Chrome has added new short keyword to window namespace: `Ink`. It's short enough that likely compiler might use it as a variable name. This variable in runtime will be non-deterministically throw exceptions from different places, changing from compilation to compilation. Really hard to detect, since SymbolMaps will show very-much valid code.

Last time globals were updated with Firefox 60 and Chrome 66, so ~4 years ago in 2018.

##### Steps to reproduce
Have project big enough to generate `Ink` variable by obfuscator, then run app in runtime and expect exception. Whatever is expected under `Ink`, browser class will be used.

##### Known workarounds

```
<set-configuration-property name="js.identifier.blacklist" value="Ink" />
```

This workaround is fine for local project, given the issue will be recognised by the contributors.

Fixes #9771


tbroyer (changes requested):
> I manually smoke-tested removed symbols to indeed, not exist in window namespace anymore.

I wouldn't remove anything. You smoke-tested on only 2 browsers and on their latest version. On the other hand, GWT 2.10.0 is still supporting IE 11.

I don't think anything has ever been removed from these files, and the cost of keeping legacy globals is close to zero.

piotrl:
@tbroyer sure, for backward compatibility reasons I've left only new symbols

there are few lines reordered as Chrome sorting mechanism has been changed, but nothing is removed.

niloc132:
(authored but not sent before you had replied, agreeing with Thomas) Agreed, please don't remove strings, there are firefox ESR (extended support release) versions to worry about too (and for chrome, countless embedded installations...)

It is possible we could introduce bugs even in modern browsers (for sufficiently old apps/libraries) by removing those as well - for example a polyfill might avoid creating a global property if it already exists, but if it doesn't exist it will create it. For old enough browsers before the feature was an idea, and for new enough browsers to reject the idea, the symbol might not exist, so GWT should continue to not create that symbol even if it no longer exists in general, or we might confuse such cases.

niloc132 (approved):
Adding my approval - any other thoughts before we merge this?

tbroyer (approved):

MCMicS (approved):

niloc132:
Sorry for sitting on this, I don't like writing commit messages for other people's code :)

Proposed commit message:
```
Update chrome/firefox globals to latest (#9773)

This change-set contains all updates of added globals, to match Firefox
104.0.1 and Chrome 105.0.5195.125.

Last time it was done in 2018, so the changes are significant. Manually
smoke tested to ensure that conflicting `Ink` no longer is present in
compiled output.
 
Fixes #9771
 ```

PR #9775, opened 2022-09-24
Response-Streaming GWT RPC; Fixes #7987

### LINKED ISSUE #7987: feature request: Appendable-based ServerSerializationStreamWriter
Originally reported on Google Code with ID 7990

```
Currently, ServerSerializationStreamWriter builds a string in memory by successive concatenation,
this is wasteful if the resulting output ends up being written to an output stream.

Would it be feasible to add an alternate implementation which takes an Appendable object
on creation, and then writes the output to that directly?

Here's some related code - the SerializationStreamWriter interface is pretty simple,
but the layers on top of that add a lot of complications such as a maximum array length
workaround for IE6/7:

https://code.google.com/p/google-web-toolkit/source/browse/trunk/user/src/com/google/gwt/user/server/rpc/impl/ServerSerializationStreamWriter.java
https://code.google.com/p/google-web-toolkit/source/browse/trunk/user/src/com/google/gwt/user/server/rpc/impl/ServerSerializationStreamWriter.java
https://code.google.com/p/google-web-toolkit/source/browse/trunk/user/src/com/google/gwt/user/client/rpc/SerializationStreamWriter.java

The RPC encodeResponse{,For{Success,Failure}} methods would need an alternate API taking
an Appendable and using that instead of the toString method:
https://code.google.com/p/google-web-toolkit/source/browse/trunk/user/src/com/google/gwt/user/server/rpc/RPC.java#596

Are others interested in this as well? Does it sound feasible?
```

Reported by `klausw@google.com` on 2013-02-04 20:25:06

axeluhl:
This PR changes the serialization such that instead of reversing the order of the payload tokens they are streamed in forward order. If payload-intercepting methods are provided in RemoteServiceServlet subclasses they will still be provided the full payload. Otherwise, the payload will only be streamed to the HTTP response stream and will not exist as a full byte[] or char[] or StringBuffer in the server's memory anymore.

This saves approximately 10% CPU time for response payloads in the 10s of MB range, and it eliminates a seven-times memory footprint during assembling and sending the response payload, reducing memory issues on RPC servers with many concurrent clients making requests that lead to large responses.

The PR makes incompatible changes to a few static public methods and to the processCall method where a new Writer parameter is introduced. Migration of subclasses overriding this method should, however, be straightforward.

Fixes #7987

### DIFF
--- a/user/src/com/google/gwt/user/client/rpc/SerializationStreamReader.java
+++ b/user/src/com/google/gwt/user/client/rpc/SerializationStreamReader.java
@@ -32,6 +32,8 @@ public interface SerializationStreamReader {
 
   int readInt() throws SerializationException;
 
+  int readIntFromEnd() throws SerializationException;
+
   long readLong() throws SerializationException;
 
   Object readObject() throws SerializationException;
--- a/user/src/com/google/gwt/user/client/rpc/impl/AbstractSerializationStream.java
+++ b/user/src/com/google/gwt/user/client/rpc/impl/AbstractSerializationStream.java
@@ -42,7 +42,15 @@ public abstract class AbstractSerializationStream {
   /**
    * The newest supported RPC protocol version.
    */
-  public static final int SERIALIZATION_STREAM_MAX_VERSION = 8;
+  public static final int SERIALIZATION_STREAM_MAX_VERSION = 9;
+  
+  /**
+   * The first version that is streaming content in forward order from
+   * server to client, passing through a {@link Writer} for writing the
+   * serialization payload to the HTTP response with close to no additional
+   * memory footprint for buffering and converting.
+   */
+  public static final int SERIALIZATION_STREAM_FORWARD_STREAMING_VERSION = 9;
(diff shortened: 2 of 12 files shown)


axeluhl:
BTW, both processCall flavors have the following in their Javadoc:

   * This is public so that it can be unit tested easily without HTTP.

It seems these methods were never meant to be overridden by regular subclasses, letting it seem fair to change their contract between minor versions.

niloc132:
Thanks for your patience on the review, I hope to do some testing this week and give you proper feedback.

One patch that will need to be landed first is our apichecker to spot api breakage, but you've taken care to avoid any such problems, so this shouldn't be a roadblock.

axeluhl:
https://github.com/gwtproject/gwt/pull/9779 has an API-compatible version. Closing this one.

axeluhl:
> Thanks for your patience on the review, I hope to do some testing this week and give you proper feedback.

Any news so far?

PR #9780, opened 2022-10-26
Java9+ support for doctool

### LINKED ISSUE #9683: Update gwt build to run under Java 11
As of 2.8.2, GWT supports being run under Java7+, but the build itself only works cleanly in Java 7. As of GWT 2.9, this changed to support Java8+ (via https://gwt-review.googlesource.com/c/gwt/+/22320), but Java 9+ still do not work to actually produce the GWT distribution. This issue is a placeholder to track the problems faced.

- [x] Internal build for Java 7 vs "modern" can be removed
- [x] `SwingLoggerPanel.java` uses improper generics for Java 9+. Previously, the `postorderEnumeration()` returned a raw `Enumeration`, but that changed in Java 9 https://docs.oracle.com/javase/9/docs/api/javax/swing/tree/DefaultMutableTreeNode.html#postorderEnumeration--
- [x] `CloseableJarHandler.java` uses classes in the java.base module that are not exported. An extra compiler arg can either make that accessible, or it is possible that this class is dead and can be removed
- [ ] The GWT doctool project uses `com.sun.javadoc` packages `com.sun.tools.doclets` package, deprecated in java9, gone in 11. Finding a cross-version way to support this could be difficult (having not tried at all yet), it may be easier to find a different way to support the custom documentation present in GWT.

niloc132:
Update doctool to run on Java 11, thus allowing either Java 8 or 11 to build GWT itself. Because the new javadoc types are not available in Java 8, the ant wiring has been configured to just skip docs when building with Java 8. CI is updated to build with both 8 and 11. 

The old Doclet is entirely removed, so some link tags are "broken" now, that were automatically fixed by the old implementation. Follow-up should be done to replace `@link` with actual `<a href>` tags, as well as to correct other javadoc errors that have persisted over time.

Note that Java 17 is not supported yet, the EztDoclet must be rewritten next to accommodate that. Building with Java >=13 will give us search functionality, so this is something we should aim for in the near future, even if we don't require Java 17 for other features.

Testing:
* There are no changes in the actual output of the ezt fragment (only the `$isInstance` methods are extra, and will be removed in a future fix).
* Javadoc can be compared at https://www.gwtproject.org/javadoc/latest/ vs https://colinalworth.com/javadoc/index.html?overview-summary.html to force frames (note that java 11 has removed frames by default, but I've enabled them here until we have search).
   * https://www.gwtproject.org/javadoc/latest/com/google/gwt/cell/client/AbstractCell.html vs https://colinalworth.com/javadoc/com/google/gwt/cell/client/AbstractCell.html for several usages of `@example`
   * https://colinalworth.com/javadoc/com/google/gwt/user/client/ui/FormPanel.SubmitCompleteEvent.html#getResults() vs https://www.gwtproject.org/javadoc/latest/com/google/gwt/user/client/ui/FormPanel.SubmitCompleteEvent.html#getResults-- for an example of `@tip` - not widely used, but has been updated to copy the text.
   * https://colinalworth.com/javadoc/com/google/gwt/i18n/client/Dictionary.html vs https://www.gwtproject.org/javadoc/latest/com/google/gwt/i18n/client/Dictionary.html for an example of `@gwt.include`, inlining JS and XML files as part of the example docs.

This patch depends on https://github.com/gwtproject/gwt/pull/9777 landing first, which removes unused files for generation of an intermediate doc format, which was historically used for googlecode wiki files.

Fixes #9683


vegegoku:
this link from the PR description is broken
https://colinalworth.com/javadoc-update/com/google/gwt/cell/client/AbstractCell.html

Also here https://colinalworth.com/javadoc/com/google/gwt/user/client/ui/FormPanel.SubmitCompleteEvent.html#getResults()
The tip is different from the original one even though there is no changes in the FormPanel in the commit, could this be updated with another PR?

niloc132:
Fixed the link, the new AbstractCell docs should be https://colinalworth.com/javadoc/com/google/gwt/cell/client/AbstractCell.html

That `@tip` content is definitely wrong. I think my sample size was too small to assume that AbstractTaglet.getText will only ever contain a single node. It looks like `<code>null</code>` will end up introducing a new node in the tree, which has to safely be copied.

vegegoku:
LGTM

niloc132:
Before this can merge, https://github.com/gwtproject/tools/pull/26 needs to land.

niloc132:
This has now been updated to fix broken tests, but those tests won't build without the above tools PR. See https://github.com/niloc132/gwt/actions/runs/4519854767/jobs/7960588154 for a full passing build on Java 8 and 11 using that PR.

niloc132:
Any more thoughts on this? Especially with the other Java 9-11 fixes landing, I think we should finalize this.

PR #9782, opened 2022-10-26
Add 2.10 apichecker config

niloc132:
Depends on https://github.com/gwtproject/tools/pull/25, before that lands this will fail due to the jars being missing.


tbroyer (commented):
Shouldn't that PR add the gwt210_211userapi.conf file?

niloc132:
Yes, I'll add that and update the branch.

PR #9784, opened 2022-10-29
Fixes #9783

### LINKED ISSUE #9783: Current Version does not handle body replace/refresh from DOM while fetching rootPanel
**GWT version:**  Current Version , also replicable on previou versions as well
**Browser (with version):**  All browsers
**Operating System:**  All OS

---

##### Description

When we do RootPanel.get() , it fetches us the rootPanel with body of the application as it's element . We also store this rootPanel in a map with key as "null". Now , in some cases/application , the body gets replaced/refreshed from DOM on certain actions like page navigation. On calling RootPanel.get() from new page , it returns the cached rootPanel with wrong body reference as it's element .

##### Steps to reproduce

Consider any application where body gets deleted and added again on page navigation . 
Navigate from one page to another , try calling RootPanel.get() , you will get a rootPanel with old body reference as it's element

##### Known workarounds

In RootPanel.java class , in get function , all cases are handled apart from the case when id and elem are null and the body is replaced , handling this should solve this .

nish28:
Consider a scenario where on page navigation , application removes the body from DOM and add it again . on calling RootPanel.get() from new page will return a existing rootpanel cached on previous page with id equals null . Here the element of root panel is nothing but the DOM's body. The cached rootPanel will have reference to the old body and all the actions would try to be performed on wrong body reference.  Putting id and elem null check will handle this scenario well and will not touch the other positive scenarios .

### DIFF
--- a/user/src/com/google/gwt/user/client/ui/RootPanel.java
+++ b/user/src/com/google/gwt/user/client/ui/RootPanel.java
@@ -183,7 +183,7 @@ public class RootPanel extends AbsolutePanel {
         // current body to handle scenarios where body gets replaced/refreshed
         // from DOM on certain actions like page navigation.
         if(id == null && elem == null) {
-          rp.setElement((Element) natGetBodyElement());
+          rp.setElement((Element) getBodyElement());
         }
         // There's already an existing RootPanel for this element. Return it.
         return rp;
--- a/user/src/com/google/gwt/user/client/ui/RootPanel.java
+++ b/user/src/com/google/gwt/user/client/ui/RootPanel.java
@@ -182,7 +182,7 @@ public class RootPanel extends AbsolutePanel {
         // if id and elem are null, set the element of existing rp with the 
         // current body to handle scenarios where body gets replaced/refreshed
         // from DOM on certain actions like page navigation.
-        if(id == null && elem == null) {
+        if(id == null) {
           rp.setElement((Element) getBodyElement());
         }
         // There's already an existing RootPanel for this element. Return it.
--- a/user/src/com/google/gwt/user/client/ui/RootPanel.java
+++ b/user/src/com/google/gwt/user/client/ui/RootPanel.java
@@ -179,6 +179,12 @@ public class RootPanel extends AbsolutePanel {
       // for any reason, return a new RootPanel rather than the existing one (
       // see issue 1937).
       if ((elem == null) || (rp.getElement() == elem)) {
+        // if id and elem are null, set the element of existing rp with the 
+        // current body to handle scenarios where body gets replaced/refreshed
+        // from DOM on certain actions like page navigation.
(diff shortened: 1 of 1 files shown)


user/src/com/google/gwt/user/client/ui/RootPanel.java
```diff
@@ -179,6 +179,12 @@ public static RootPanel get(String id) {
       // for any reason, return a new RootPanel rather than the existing one (
       // see issue 1937).
       if ((elem == null) || (rp.getElement() == elem)) {
+        // if id and elem are null, set the element of existing rp with the 
+        // current body to handle scenarios where body gets replaced/refreshed
+        // from DOM on certain actions like page navigation.
+        if(id == null && elem == null) {
```
tbroyer, line 185: If `id` is null, then `elem` is necessarily null as well (`elem` is only assigned at line 172 when `id` is not null); and conversely, if `elem` is null at that point then `id` is null (because at lines 172–173, when `id` is not null and `elem` is null, then the method returns).

user/src/com/google/gwt/user/client/ui/RootPanel.java
```diff
@@ -179,6 +179,12 @@ public static RootPanel get(String id) {
...
+          rp.setElement((Element) natGetBodyElement());
+        }
```
tbroyer, line 187: This should probably rather avoid entering this `if` block when `rp.getElement() != getBodyElement()` and thus eventually return a new `DefaultRootPanel`. Using `setElement` on an existing widget that already has an element (like a `DefaultRootPanel`) is plain wrong (and will throw an assertion error in dev mode).

(not to mention that this code doesn't even compile as `natGetBodyElement()` doesn't exist)

tbroyer (changes requested):

user/src/com/google/gwt/user/client/ui/RootPanel.java
nish28: Resolved , placed a single (id == null) check only

user/src/com/google/gwt/user/client/ui/RootPanel.java
nish28: My bad ! I have replaced with getBodyElement() . 
The if check at line 181 first check for (elem == null) which is true and does not validate if element matches with the body or not . So added this check to keep the rootPanel same with the correct body reference

niloc132: If `elem` is non-null, then the `if` will continue with the `rp.getElement() == elem` check, and then if `id` happens to be null, the rootpanel will get its element replaced... with its current element. Even though that is the same element, in dev mode `setElement()` will still fail.

niloc132:
Closing this as stale.

Issue #9752, opened 2022-06-10
Please remove protobuf pom from gwt-servlet


YSavanier:
Hello

Since there is vulnerability CVE-2021-22569 (see #9659 & #9749) on protobuf 2.5.0 which is detected into gwt-servlet by sonarqube while this is a false positive and no protobuf package is present into the jar (only internal classes), could you please remove the whole folder META-INF\maven\com.google.protobuf please ?

I can't see why there is any need to keep this and it's removal would solve this problem instantly.

Thank you very much.

B.R.

niloc132:
Thanks for the report - yes, we can look into tweaking the packaging scripts to keep this out. Is that something you'd be interested in looking into fixing for the next release?

It looks like the gwt-servlet.jar packaged in the distribution zip contains both the maven marker and the classes, though the classes are "rebased", with a different package than usual - this is produced by the servlet/build.xml file.

It also appears that the gwt-servlet.jar deployed to maven still includes those classes. It isn't clear at a glance what gwt-servlet needs with those classes - I know that protobuf is necessary for legacy dev mode, but I don't presently see what purpose these serve for a production server.

tbroyer:
IIRC protobuf is also used for source maps, possibly also the streaminghtmlparser.

YSavanier:
Hi, thank you for your answer, 

Actually the fix is quite simple, in fact this protobuf meta-inf folder content is provided by the unzipping of [protobuf-java-rebased-2.5.0.jar](https://github.com/gwtproject/tools/blob/master/lib/protobuf/protobuf-2.5.0/protobuf-java-rebased-2.5.0.jar) located in the tools lib folder at https://github.com/gwtproject/tools/tree/master/lib/protobuf/protobuf-2.5.0.

But while in the readme it is clearly stated at step 3 :

```
3) Remove the META-INF directory from protobuf-java-rebased-2.5.0.jar. You can do this
   by expanding the jar, deleting the META-INF directory, and rebuilding the jar with the
   -M (no MANIFEST) switch.
```

This step has clearly not been applied to the protobuf-java-rebased-2.5.0.jar file which still contains this META-INF folder.

So to me, the solution to me appear just to update a new version of this protobuf-java-rebased-2.5.0.jar with the META-INF folder removed at https://github.com/gwtproject/tools/blob/master/lib/protobuf/protobuf-2.5.0/protobuf-java-rebased-2.5.0.jar

Do you want me to send you the jar file stripped of its META-INF folder ?

[...]

PR #9785, opened 2022-11-11
Fixes #9778

### LINKED ISSUE #9778: gwt-servlet misidentified as protobuf-java by static security analyzers
**GWT version:** 2.10.0
**Browser (with version):** n/a
**Operating System:** n/a

---

##### Description

Some static security analyzers misidentify gwt-servlet as being protobuf-java, and thus report vulnerabilities.

While it is true that protobuf-java is bundled within gwt-servlet, it is also repackaged and AFAICT is not used for serialization/deserialization, only for protobuf-generated value objects.

At a minimum, gwt-servlet should not contain protobuf-java's POM in its META-INF. The same is probably true for other GWT jars and other bundled-dependencies. This could either be done in gwtproject/tools, removing the files from the JARs that will later be bundled, or in gwtproject/gwt when bundling.

##### Steps to reproduce

Run Dependency-Check on a project with gwt-servlet dependency.

##### Known workarounds

Most tools using heuristics can be configured to override bad heuristic findings.

https://github.com/jeremylong/DependencyCheck/issues/4912

rjeeb:
Excluding the `pom.xml` file from rebased `protobuf-java-rebased-2.5.0.jar`.

The fix excludes the pom file from the jar to ensure that dependency check libraries don't deal with it as usual `protobuf-java` dependency. It was tested on a project with [jeremylong dependency check maven plugin](https://jeremylong.github.io/DependencyCheck/dependency-check-maven/) and the result attached has no vulnerability on `protobuf-java`.

[dependency-check-report.zip](https://github.com/gwtproject/gwt/files/9992838/dependency-check-report.zip)


niloc132:
Thanks for this patch.

I've deployed a build of this to https://repo.vertispan.com/gwt-snapshot/ with version 2.11.0-fix-9778-SNAPSHOT. This uses the new groupIds, org.gwtproject:gwt-servlet:2.11.0-fix-9778-SNAPSHOT.

For example, see https://repo.vertispan.com/gwt-snapshot/org/gwtproject/gwt-servlet/2.11.0-fix-9778-SNAPSHOT/ to get the gwt-servlet jar.

FrankHossfeld (approved):

zelinskiyvs (approved):

niloc132:
If anyone can [report back that they've tested this and confirmed it removes the false positive](https://github.com/gwtproject/gwt/issues/9752#issuecomment-1723299005), we can see about merging it. That said, another change is in progress to update the sourcemap support so that gwt-servlet.jar doesn't use protobuf at all anyway, so these issues will be resolved in 2.12.

aragonedgar (approved):

niloc132:
Closed, for GWT 2.12 this will be handled by #9936. This could be backported for a potential 2.11.x bugfix release, should that happen to continue Java 8 support for the compiler and dev tools.

PR #9796, opened 2023-01-29
Update installScriptEarlyDownload.js to remove old workaround

### LINKED ISSUE #9725: CSP error on Chromium-based browsers
**GWT version:** 2.9.0
**Browser (with version):** Any Chromium-based browser (Chromium, Chrome, Chromium-based Edge)
**Operating System:** Tested on Linux and Windows

---

##### Description
The following error is triggered by the browser:

```
Some messages have been moved to the Issues panel.
VM11:1 Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "default-src 'self' 'unsafe-inline'".

    at <anonymous>:1:1
    at m (Geot.nocache.js:9)
    at Geot.nocache.js:10
    at HTMLDocument.d (Geot.nocache.js:6)
```

What is really strange with this error is that it disappears if the project is compiled with -Dgwt.style=PRETTY.
When this problem occurs (i.e. when gwt.style has its default value of OBFUSCATED), the following code is responsible for the error:
chrome_csp_gwt

In this obfuscated code, we can see that there is at line 203 a test to detect if we are using a Chrome-based user. If this is the case, the CSP error is triggered at line 214. At this line, we can see that the 'g' variable is a script tag containing a huge "eval(...)" function. This is this unsafe eval that triggers the CSP error.
On the other hand, if we use another browser (e.g. firefox), the other part of the "if/else" block is executed, and in this part the 'g' variable holds a script tag without any "eval()" function.

As I'm not using any "eval()" function in my code, and I don't even have some code that tests the name of the browser, I suppose that this function is directly generated by GWT and I don't have any control on it.

Is it a known problem and is there any workaround available (except lowering the CSP protection level by accepting unsafe eval functions?).
Thanks for your answers.

bones-bones:
The eval behavior in GWT is no longer needed since that chrome bug ([#90707](https://bugs.chromium.org/p/chromium/issues/detail?id=90707)) was fixed (it looked like the bug got lost in the weeds at some point.) Removing that behavior improves the performance of bootstrapping. It also coincidently fixes the referenced CSP issue.

Considered unifying the logic between browsers, but Safari took a bit of a perf hit when trying to insert a single large text.

Fixes https://github.com/gwtproject/gwt/issues/9725

### DIFF
--- a/dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptEarlyDownload.js
+++ b/dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptEarlyDownload.js
@@ -20,7 +20,7 @@ function installScript(filename) {
     var docbody = doc.body;
     var script;
     // for sourcemaps, we inject the code as a single string for Chrome
-    if (navigator.userAgent.includes("Chrome")) {
+    if (navigator.userAgent.indexOf("Chrome") > -1) {
       var codeString = "";
       for (var i = 0; i < code.length; i++) {
         codeString += code[i];
--- a/dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptEarlyDownload.js
+++ b/dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptEarlyDownload.js
@@ -21,19 +21,13 @@ function installScript(filename) {
     var script;
     // for sourcemaps, we inject the code as a single string for Chrome
     if (navigator.userAgent.indexOf("Chrome") > -1) {
-      var codeString = "";
-      for (var i = 0; i < code.length; i++) {
-        codeString += code[i];
-      }
       script = doc.createElement('script');
-      script.language = 'javascript';
-      script.text = codeString;
+      script.text = code.join('');
       docbody.appendChild(script);
       removeScript(docbody, script);
     } else {
       for (var i = 0; i < code.length; i++) {
         script = doc.createElement('script');
-        script.language='javascript';
         script.text = code[i];
         docbody.appendChild(script);
         removeScript(docbody, script);
--- a/dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptEarlyDownload.js
+++ b/dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptEarlyDownload.js
@@ -19,23 +19,15 @@ function installScript(filename) {
     var doc = getInstallLocationDoc();
     var docbody = doc.body;
     var script;
-    // for sourcemaps, we inject textNodes into the script element on Chrome
-    if (navigator.userAgent.indexOf("Chrome") > -1 && window.JSON) {
-      var scriptFrag = doc.createDocumentFragment()
-      // surround code with eval until crbug #90707 
-      scriptFrag.appendChild(doc.createTextNode("eval(\""));
+    // for sourcemaps, we inject the code as a single string for Chrome
+    if (navigator.userAgent.includes("Chrome")) {
+      var codeString = "";
       for (var i = 0; i < code.length; i++) {
-        // escape newlines, backslashes, and quotes with JSON.stringify
-        // rather than create multiple script tags which mess up line numbers, we use 1 tag, multiple text nodes
-        var c = window.JSON.stringify(code[i]); 
-        // trim beginning/end quotes
-        scriptFrag.appendChild(doc.createTextNode(c.substring(1, c.length - 1)));
+        codeString += code[i];
       }
-      // close the eval
-      scriptFrag.appendChild(doc.createTextNode("\");"));
       script = doc.createElement('script');
-      script.language='javascript';
-      script.appendChild(scriptFrag);
+      script.language = 'javascript';
+      script.text = codeString;
       docbody.appendChild(script);
       removeScript(docbody, script);
     } else {


dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptEarlyDownload.js
```diff
@@ -19,23 +19,15 @@ function installScript(filename) {
     var doc = getInstallLocationDoc();
     var docbody = doc.body;
     var script;
-    // for sourcemaps, we inject textNodes into the script element on Chrome
-    if (navigator.userAgent.indexOf("Chrome") > -1 && window.JSON) {
-      var scriptFrag = doc.createDocumentFragment()
-      // surround code with eval until crbug #90707 
-      scriptFrag.appendChild(doc.createTextNode("eval(\""));
+    // for sourcemaps, we inject the code as a single string for Chrome
+    if (navigator.userAgent.includes("Chrome")) {
```
zbynek, line 23: this [breaks ](https://caniuse.com/es6-string-includes) IE11 compatibility

bones-bones: I am ashamed, reverted to indexOf

dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptEarlyDownload.js
```diff
@@ -19,23 +19,15 @@ function installScript(filename) {
...
+    // for sourcemaps, we inject the code as a single string for Chrome
+    if (navigator.userAgent.indexOf("Chrome") > -1) {
+      var codeString = "";
       for (var i = 0; i < code.length; i++) {
-        // escape newlines, backslashes, and quotes with JSON.stringify
-        // rather than create multiple script tags which mess up line numbers, we use 1 tag, multiple text nodes
-        var c = window.JSON.stringify(code[i]); 
-        // trim beginning/end quotes
-        scriptFrag.appendChild(doc.createTextNode(c.substring(1, c.length - 1)));
+        codeString += code[i];
```
niloc132, line 26: Why not simply `script.text = code.join();`?

bones-bones: TBH I mostly looked at the recent answers on https://stackoverflow.com/questions/7296594/array-join-vs-string-concat

niloc132: Methodology matters a lot - remember for the answers that do loop(`push`)+`join` and compare to loop(`+=`) that we don't even have a loop - the array already exists, we're only doing `join` and comparing it against loop(`+=`). https://stackoverflow.com/a/65402300/860630 seems to test this directly, while https://stackoverflow.com/a/67840218/860630 tests the combined "make the array" and "join it"). I'm not sure there's a clear winner in that light?

That said, if it is close, I'd personally vote for concise/clarity.

dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptEarlyDownload.js
```diff
@@ -19,23 +19,15 @@ function installScript(filename) {
     var doc = getInstallLocationDoc();
     var docbody = doc.body;
     var script;
-    // for sourcemaps, we inject textNodes into the script element on Chrome
-    if (navigator.userAgent.indexOf("Chrome") > -1 && window.JSON) {
-      var scriptFrag = doc.createDocumentFragment()
-      // surround code with eval until crbug #90707 
-      scriptFrag.appendChild(doc.createTextNode("eval(\""));
+    // for sourcemaps, we inject the code as a single string for Chrome
```
tbroyer, line 22: Shouldn't we do this for all browsers as sourcemaps are now supported everywhere?

dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptEarlyDownload.js
tbroyer: > Why not simply `script.text = code.join();`?

+1

Use `.join("")` though, as `.join()` without argument uses an implicit `","` separator

dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptEarlyDownload.js
```diff
@@ -19,23 +19,15 @@ function installScript(filename) {
...
-        // trim beginning/end quotes
-        scriptFrag.appendChild(doc.createTextNode(c.substring(1, c.length - 1)));
+        codeString += code[i];
       }
-      // close the eval
-      scriptFrag.appendChild(doc.createTextNode("\");"));
       script = doc.createElement('script');
-      script.language='javascript';
-      script.appendChild(scriptFrag);
+      script.language = 'javascript';
```
tbroyer, line 29: This has been deprecated for years, I think we can safely delete those `.language` lines

tbroyer (changes requested):

dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptEarlyDownload.js
bones-bones: Can I do that in a later PR? I'd want to perf test in safari and ff since it's a change in how many chunks the scripts are evaluated in

dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptEarlyDownload.js
bones-bones: Thanks for the pointers, moved to `join('')`

dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptEarlyDownload.js
niloc132: It seems likely that with the general demise of IE that we don't even need these chunks any more, and can default to no limit, so that all browsers get this as a single string (without even a need for the join).

bones-bones: Oh so that's IE related? Neato. I found [this review](https://github.com/gwtproject/gwt/commit/4814ffae4a81d396a2650d46f0b3c120bb69b926) but I've been having trouble getting more context around the change, is there something I can read up on?
Can that also be in another PR that I'll do? I'd want to do a bunch more testing around that change.

tbroyer: The commit you link to shows that this was an IFrameLinker thing that had been ported to xsiframe. Looking at the IFrameLink history, we can find https://github.com/gwtproject/gwt/commit/7bcb57ae4fd79ffd0eba43dda335fda8c8c0458f (nearly 14 years old):

> This patch modifies the IFrame linker to use multiple script tags instead of one giant one.  This improves parsing time on Firefox, and it should improve overall loading latency on any browser whenever the network latency is significant.

I'm not sure how this applies to xsiframe though, as the chunks aren't streamed over the network (contrary to HTML in the iframe linker), unless it was part of an experiment at Google at the time to somehow use a .cache.html with an xsiframe subclass?

zbynek:
Works for me both in SuperDevMode and compiled.  FTR I used the following CSP:
```
default-src 'self' 'unsafe-inline' http://localhost:9876/; img-src 'self' data:
```

niloc132:
@bones-bones can you file a follow-up for the aspects of the review you're deferring, and then I think we land this as-is?

bones-bones:
https://github.com/gwtproject/gwt/issues/9818 
Opened this and will start on it soon.

niloc132 (approved):

PR #9800, opened 2023-02-10
Fixed set- and clearTimeout to make debugging working with Firefox

keinhaar:
This fixes a problem that only occurs in firefox while debugging.
Without the fix SDBG Users would be forced to super-source the Impl class.
After a few weeks of testing i did'nt find any negative effects on runtime, or in other browsers.
Has been discussed in #9787.
Fixes #9787

### DIFF
--- a/user/src/com/google/gwt/core/client/impl/Impl.java
+++ b/user/src/com/google/gwt/core/client/impl/Impl.java
@@ -393,7 +393,7 @@ public final class Impl {
   }-*/;
 
   private static native void watchdogEntryDepthCancel(int timerId) /*-{
-    $wnd.clearTimeout(timerId);
+    clearTimeout(timerId);
   }-*/;
 
   private static void watchdogEntryDepthRun() {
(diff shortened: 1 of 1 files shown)


zbynek (changes requested):
this PR is based on the patch that @niloc132 provided in the issue as a workaround with a comment

> I don't think this is a reasonable fix to make to GWT itself, but it does look like we could remove the watchdog entirely, or only conditionally enable it.

so I think it would be better implement that instead.

keinhaar:
> so I think it would be better implement that instead.
Maybe, but nobody is working on removing the watchdog, and this is a very simple fix. 
In all of my projects it works without any problems.
So i propose to do this now, and later on, somebody could remove the watchdog.

niloc132:
Sorry for the delay on this. I finally simplified the bug report and filed it at https://bugzilla.mozilla.org/show_bug.cgi?id=1833637. 

The watchdog timer isn't needed at all to see invalid behavior. Unfortunately, this shows that this patch doesn't go far enough - not only can the watchdog timer keep running, but any other timer can too - we have to fix any other call to `$wnd.setTimeout` to instead reference the local window/self, or we might hit other issues with this while a debugger is open. With the exception of RichTextAreaImplStandard and ScrollImpl, every call I see in GWT to setTimeout goes through $wnd, so will have this problem. Elemental2's DomGlobal.setTimeout will have this issue as well.

keinhaar:
The same problem occurs in automated tests using selenium with the latest htmlunit-driver.
The proposed fix does resolve this too.

PR #9802, opened 2023-02-24
Added basic LocalDate support

keinhaar:
Because LocalDate is used very often in these days, it would be great if it was supported per default, instead of beeing super sourced in every project.
See https://github.com/gwtproject/gwt/issues/611
This implementation is very basic, but has been used in production level code for some time.
I would be happy to extend the implementation add to add more emulated classes like GregorianCalendar if there is any interest.

### DIFF
--- a/user/super/com/google/gwt/emul/java/time/LocalDate.java
+++ b/user/super/com/google/gwt/emul/java/time/LocalDate.java
@@ -1,0 +1,100 @@
+/*
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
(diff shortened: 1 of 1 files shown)


niloc132:
This is pretty rough, and some of the basic methods that don't even rely on other java.time types that are missing: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/time/LocalDate.html

Projects shouldn't need to emulate their own version, there are existing emulation libraries out there like https://github.com/foal/gwt-time, with much more complete emulation of this type: https://github.com/foal/gwt-time/blob/master/time/src/main/java/org/jresearch/threetenbp/gwt/emu/java/time/LocalTime.java

Another impl, looks like it came from the same original source https://github.com/adrianmsmith/gwt-time/, https://github.com/adrianmsmith/gwt-time/blob/master/src/main/resources/com/google/gwt/emul/java/time/LocalTime.java

keinhaar:
OK, didn't know that.
But could you please explain why this isn't in the main GWT branch?

niloc132:
The short answer is a mix of "no one has made a pull request with that code", but the long answer has to touch on a few more things: 
 * Is the code safe to relicense, is the contributor confident that the original code is all safe to distribute like this?
 * Are there enough tests to put this into GWT itself?
 * Is the implementation complete enough?
 * Extra work of formatting the code, ensuring it builds and tests make sense in various Java versions.

keinhaar:
Closing because there are better solutions (see above), even though they will not make it into GWT

PR #9814, opened 2023-03-27
Avoid document write in single script linker loading

### LINKED ISSUE #9797: Replace document.write('<script ...>') with document.createElement('script')
**GWT version:** 2.10.0
**Browser (with version):** Chrome 
**Operating System:** Version 109.0.5414.119 (Official Build) (x86_64)

---

##### Description

I'm using `SingleScriptLinker` to produce a single Javascript file, and the generated JS code contains:

```
    $doc_0.write('<script id="' + markerId + '"><\/script>');
```

Which causes errors when the GWT output JS is used in a Chrome Extension:

> Failed to execute 'write' on 'Document': It isn't possible to write into a document from an asynchronously-loaded external script unless it is explicitly opened.

According to the latest documentation about `document.write`:
https://developer.mozilla.org/en-US/docs/Web/API/Document/write

> Use of the document.write() method is strongly discouraged.

I'm wondering if it's a good idea to replace the following two lines:
```
    $doc.write('<script id="' + markerId + '"></script>');
    markerScript = $doc.getElementById(markerId);
```
https://github.com/gwtproject/gwt/blob/main/dev/core/src/com/google/gwt/core/linker/SingleScriptTemplate.js#L96-L97

with:

```
    markerScript = $doc.createElement('script');
    markerScript.id = markerId;
    $doc.body.appendChild(markerScript);
```

If this makes sense, there are a few other script templates needing similar changes.
https://github.com/gwtproject/gwt/search?q=%24doc.write%28%27%3Cscript&type=code

##### Steps to reproduce

Inspect the generated GWT Javascript code, you will see `doc.write` in it.

##### Known workarounds

I've tested the new template in a custom linker, which avoided `document.write` in the generated Javascript, hence eliminated such errors in Chrome Extension.

Mivr:
Avoid document write method in single script linker.

Fixes #9797

### DIFF
--- a/dev/core/src/com/google/gwt/core/linker/SingleScriptTemplate.js
+++ b/dev/core/src/com/google/gwt/core/linker/SingleScriptTemplate.js
@@ -89,21 +89,6 @@ function __MODULE_FUNC__() {
   // base url.
   //
   function computeScriptBase() {
-    var thisScript
-    ,markerId = "__gwt_marker___MODULE_NAME__"
-    ,markerScript;
-
-    $doc.write('<script id="' + markerId + '"></script>');
-    markerScript = $doc.getElementById(markerId);
-
-    // Our script element is assumed to be the closest previous script element
-    // to the marker, so start at the marker and walk backwards until we find
-    // a script.
-    thisScript = markerScript && markerScript.previousSibling;
-    while (thisScript && thisScript.tagName != 'SCRIPT') {
-      thisScript = thisScript.previousSibling;
-    }
-
     // Gets the part of a url up to and including the 'path' portion.
     function getDirectoryOfFile(path) {
       // Truncate starting at the first '?' or '#', whichever comes first. 
@@ -119,6 +104,7 @@ function __MODULE_FUNC__() {
       return (slashIndex >= 0) ? path.substring(0, slashIndex + 1) : '';
     };
 
+    var thisScript = $doc.currentScript;
     if (thisScript && thisScript.src) {
       // Compute our base url
       base = getDirectoryOfFile(thisScript.src);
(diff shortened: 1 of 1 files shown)


Mivr:
[jnehlmeier](https://github.com/jnehlmeier) or anyone from GWT, any feedback on this PR?

dev/core/src/com/google/gwt/core/linker/SingleScriptTemplate.js
```diff
@@ -119,6 +104,7 @@ function __MODULE_FUNC__() {
       return (slashIndex >= 0) ? path.substring(0, slashIndex + 1) : '';
     };
 
+    var thisScript = $doc.currentScript;
```
jnehlmeier, line 107: According to caniuse.com `currentScript` is not supported by IE 11. So you can not use it for now unless a decision has been made to drop IE 11 in the next GWT release.

Mivr: Are you OK to make the code try currentScript and if fails to try the old implementation?

niloc132: That seems reasonable to me - put an explicit note so that when IE 11 is finally truly dead we know that this can be simplified out?

jnehlmeier: > Are you OK to make the code try currentScript and if fails to try the old implementation?

Yes that would be fine.

niloc132: @Mivr do you want to make this change so we can see about merging?

Mivr: @niloc132 Yes, sorry for the delay I am just way too busy and warning are not the highest priority

niloc132: Yep no problem, just checking in with various open PRs to see what we can land before trying to do a release. No rush, we've got other pending PRs, but this one is small enough we could ship without it if you don't get the chance to get to it.

niloc132:
Closing in favor of https://github.com/gwtproject/gwt/pull/9847

PR #9817, opened 2023-04-04
docs: adjust readme for git system

sparsick:
The release instruction related to a SVN system. I adjust it so that it match the GIT system.

### DIFF
--- a/maven/README.txt
+++ b/maven/README.txt
@@ -1,9 +1,9 @@
 push-gwt.sh packages and deploys GWT to a local or remote Maven
 repository
  
-To build GWT from trunk and deploy to your local repo:
+To build GWT from main and deploy to your local repo:
 
-> cd trunk
+> git checkout main
 > ant clean dist-dev
 > maven/push-gwt.sh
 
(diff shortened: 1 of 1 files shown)


maven/README.txt
```diff
@@ -1,9 +1,9 @@
 push-gwt.sh packages and deploys GWT to a local or remote Maven
 repository
  
-To build GWT from trunk and deploy to your local repo:
+To build GWT from main and deploy to your local repo:
 
-> cd trunk
+> git checkout main
 > ant clean dist-dev
 > maven/push-gwt.sh
```
niloc132, line 8: One more request (and this is also my fault for changing it), can you adjust the push-gwt script name to be `maven/push-gwtproject.sh`? The old push-gwt will only push com.google.gwt artifacts, which we don't expect to release again. I failed to update this correctly when I made that change for the last release.

maven/README.txt
```diff
@@ -1,9 +1,9 @@
 push-gwt.sh packages and deploys GWT to a local or remote Maven
 repository
  
-To build GWT from trunk and deploy to your local repo:
+To build GWT from main and deploy to your local repo:
 
-> cd trunk
+> git checkout main
```
tbroyer, line 6: nit: `git switch main`

```suggestion
> git switch main
```

maven/README.txt
sparsick: no problem

maven/README.txt
sparsick: Thanks for the hint. I adjusted it.

niloc132 (approved):

Issue #9813, opened 2023-03-26
Ensure Java 8/9/10/11 tests are run


niloc132:
There is some odd wiring around Java 8/9/10/11 tests in the user tests, and it prevents them from running as expected. Comments in the classes suggest that they should be run even on older Java versions thanks to supersource, and the tests are set to not run if the gwt `-sourceLevel` is set to too old of a version. However, the Java 8 test for some reason requires Java 9 as a minimum, and besides, the test wiring somehow isn't ending up with the GWT sourceLevel being 11 when run on Java 11.

Likely breakdown of things that need exploration/fixing:
- [x] Fix Java8Test compilation.
- [x] Investigate if `-sourceLevel auto` actually works with Java>8 (I suspect `Utility.versionCompare()` doesn't work for newer Java versions).
- [ ] Explore setting sourceLevel to max supported when testing rather than relying on `java.specification.version` from the testing jvm. *(Default is `auto` to enable legacy dev mode to still function)*
- [x] Don't skip Java8Test when running Java 8 (or rather, sourceLevel 8).
    - [x] It looks like this test was modified to also run Java 9 tests, those should be moved to their own test class

niloc132:
Update, `auto` does work, sort of, but it apparently hasn't been the default in years (default is `1.8`, but Java8 test still required Java 9, see above). The comparison is still broken though, since `java.specification.version` uses integer values since Java 9, so those versions and alt versions will need to be corrected.

niloc132:
I misclicked and didn't close, but it turns out I misclicked, which is good because the issued isn't resolved. We're still using the default value for `-sourceLevel` which is 1.8, rather than auto. Selected tests seem to be passing locally with that in place, running the full nightly build now https://github.com/niloc132/gwt/actions/runs/7013207631.

PR #9819, opened 2023-04-07
Migrate java9 tests to their own class, run java 8 on java 8

niloc132:
Also corrects a compile error that went unnoticed in the Java 8 test, since these tests were never run even under Java 8. 

Once #9780 is merged, we can either switch to `-sourceLevel auto` or `-sourceLevel 11`, depending on what legacy dev mode breakage is acceptable and what Java versions we run tests on.

Partial fix for #9813

### DIFF
--- a/user/test-super/com/google/gwt/dev/jjs/super/com/google/gwt/dev/jjs/test/Java9Test.java
+++ b/user/test-super/com/google/gwt/dev/jjs/super/com/google/gwt/dev/jjs/test/Java9Test.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2023 GwtProject contributors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


tbroyer (approved):
LGTM after adding the license header to Java9Test.java

user/test-super/com/google/gwt/dev/jjs/super/com/google/gwt/dev/jjs/test/Java9Test.java
```diff
@@ -0,0 +1,99 @@
+/*
+ * Copyright 2023 GwtProject contributors
```
niloc132, line 2: Odd that it doesn't mind this header.

Google has no interest in this code any more, as we no longer need a CLA. I'm happy to change this PR for the sake of landing it now, but any thoughts on relaxing this requirement, allow any apache2 code, encourage a more general copyright notice?

tbroyer: +1 on relaxing the requirement. How about changing everything to "The GWT Project authors" and adding an AUTHORS file?
(see https://opensource.google/documentation/reference/releasing/authors; that's also what I did in the few "modularized gwt-user" modules, populating the AUTHORS file from the Git history)

PR #9820, opened 2023-04-07
Correct Java versions numbers to support -sourceLevel auto

niloc132:
Starting in Java 9, the value of the system property java.specification.version has been an integer, with no "1." prefix. As a result, using -sourceLevel auto with any version 9 and above has automatically used the latest version of Java that GWT supports, rather than an actual matching version.

This doesn't matter a lot today for Java 9 and 10, but with coming Java 17 support while 8 and 11 are still allowed, it may make a difference.

The altStringValues have been updated to still use the 1.x versions for backward compatibility.

Partial fix for #9813

### DIFF
--- a/dev/core/src/com/google/gwt/dev/util/arg/SourceLevel.java
+++ b/dev/core/src/com/google/gwt/dev/util/arg/SourceLevel.java
@@ -24,9 +24,9 @@ import com.google.gwt.util.tools.Utility;
 public enum SourceLevel {
   // Source levels must appear in ascending order for the default setting logic to work.
   JAVA8("1.8", "8"),
-  JAVA9("1.9", "9"),
-  JAVA10("1.10", "10"),
-  JAVA11("1.11", "11");
+  JAVA9("9", "1.9"),
+  JAVA10("10", "1.10"),
+  JAVA11("11", "1.11");
 
   /**
    * The default java sourceLevel.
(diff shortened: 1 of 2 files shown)


tbroyer (commented):
How about changing the parsing logic to trim a leading `1.` before comparing versions? (or the reverse, make sure it starts with `1.`)
That way we wouldn't have to change the values of `getStringValue()` and `getAltStringValue()`.

niloc132:
As I understand it, changing the string constants makes more sense - alt string value was always meant to mean "people might type this by mistake instead of the real version", and the string value was intended instead to mean "the value of the `java.specification.version` property.

That it "worked" until now is just an accident that, by the version compare code, "9", "10", and "11" are all greater than "1.11", so no matter what string was returned by `System.getProperty("java.specification.version")`, if you were running Java >=9, `-sourceLevel auto` meant 11.

I can't see any way that those values are used outside of these version comparison checks (string value used against `java.specification.version`, and both checked against user provided values for `-sourceLevel`), and this code isn't available to linkers/generators or user code, so I'm not sure I see the downside.

Changing the parsing logic in that way seems very dirty to me, but don't see a way that it could result in incorrect code. But sticking with a bad decision because we did it in the past doesn't seem like a good reason.

tbroyer (approved):

PR #9824, opened 2023-04-12
Update ezt tool so GWT build can run under Java 17

### LINKED ISSUE #9821: Update ezt tool so GWT build can run under Java 17
As of https://github.com/gwtproject/gwt/pull/9780 being merged, GWT can be built with Java 8-11, and runs successfully on Java 8-latest (though legacy dev mode may not work with newer versions of Java, or may require additional JVM arguments to circumvent JPMS checks). 

It looks like rewriting the Ezt doc tool is the last remaining piece to support Java 17. That is, providing a new mechanism to generate the contents of https://www.gwtproject.org/doc/latest/RefJreEmulation.html, the HTML contents after the header of https://github.com/gwtproject/gwt-site/blob/main/src/main/markdown/doc/latest/RefJreEmulation.md?plain=1.

The classes responsible for this are `com.google.doctool.JreDocTool`, the main() invoked by the target `emul-ezt` in `doc/build.xml`, a builder class `com.google.doctool.JreDocToolFactory`, and the class `com.google.doctool.custom.EztDoclet` - apparently used as a doclet to emit a plain HTML file when run.

In a replacement, ideally the HTML structure would follow the same general format (to make it easier to confirm that the contents are the same, through a git diff), and hopefully would preserve any html anchors as well (to avoid breaking inbound links). I don't see a reason that the new version should necessarily be implemented as a doclet though, as long as the functionality is otherwise consistent.

To run this today, invoke
```bash
ant doc
```
The output can then be found at `build/out/doc/emul-ezt/fragment.html`.

vegegoku:
Fixes #9821


niloc132 (commented):
What do you think about changing the name of the ezt class to something that reflects its purpose more closely - as discussed in chat, it doesnt appear to use the ezt format, but is just an html snippet.

Also, do you think we can get rid of JreDocTool itself and just invoke this from ant as javadoc with a doclet?

vegegoku:
> What do you think about changing the name of the ezt class to something that reflects its purpose more closely - as discussed in chat, it doesnt appear to use the ezt format, but is just an html snippet.
> 
> Also, do you think we can get rid of JreDocTool itself and just invoke this from ant as javadoc with a doclet?

I guess that would be better, all that JreDocTool doing is reading the args and provide some help messages then just delegating to JavaDocTool, and we can point and reference the JavaDocTool documentation instead of those help messages.

doc/build.xml
```diff
@@ -92,8 +92,7 @@
              linkoffline="http://download.oracle.com/javaee/6/api/ validation-package-list">
       <!-- suppress errors that don't break the output, we should refine this to just suppress things we haven't fixed yet -->
       <arg value="-Xdoclint:none" />
-      <!-- re-enable frames for now, until we get to >=13 and have search -->
-      <arg value="--frames" />
+
```
niloc132, line 95: We should consider leaving this in for the next release so that we can get similar looking output? Or maybe not, and jump all the way to 17 for this next build?

vegegoku: It fails to build with flag on 17.

vegegoku: flag restored

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -0,0 +1,273 @@
...
+        List<String> methods = cls.getEnclosedElements()
+                .stream()
+                .filter(element -> ElementKind.METHOD == element.getKind())
+                .filter(member -> member.getModifiers().contains(Modifier.PUBLIC))
+                .map(member -> (ExecutableElement) member)
+                .map(executableElement -> flatSignature(env, cls, executableElement))
+                .collect(Collectors.toList());
+
+        List<String> members = new ArrayList<>(constructors);
+        members.addAll(methods);
```
niloc132, line 147: It _seems_ better to just put all three of these kinds of members into one big stream... I see that to keep the diff exactly the same, the fields have to be set up on their own.

Fields and constructors though are almost identical, it looks like the difference is basically to just sort constructors before methods?

I think it is probably best leaving alone, since we want to keep the diff the same and just replace the code - use this as the way to ensure it is stable (absent any other tests, etc). Maybe if we have a reason to change this in the future, we could consolidate, or maybe even style things differently.

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -0,0 +1,273 @@
...
+
+        // Print out all constructors and methods
+        if (!members.isEmpty()) {
+            pw.format("  <dd>%s</dd>\n", String.join(", ", members));
```
niloc132, line 151: If this was higher throughput, I'd strongly encourage writing each member rather than join and then write.

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -0,0 +1,273 @@
...
+    private String outputFile;
+
+    @Override
+    public boolean run(DocletEnvironment env) {
+        try {
+
+            File outFile = new File(outputFile);
+            outFile.getParentFile().mkdirs();
+            FileWriter fw = new FileWriter(outFile);
+            PrintWriter pw = new PrintWriter(fw, true);
```
niloc132, line 67: try-with-resources?

(like these other notes, nearly no value here since this a one-off that won't keep running)

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -0,0 +1,273 @@
...
+import java.util.HashSet;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Locale;
+import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * A doclet for using producing EZT output listing the specified classes and
```
niloc132, line 49: Remove ezt here and in the constant name, and other docs? We may still want it in the constant value so that we have consistent output.

niloc132 (approved):
I think there is a lot that could be improved, but keeping the output stable and dropping the Java 9-11 requirement seems more important. Only remark here that I think might actually matter is replacing most of the `ezt` details with something about what this does rather than ... what tool was once being used _near_ this output. 

I think we should still leave a few things the same, at least for the moment - the ant target itself, and the output file, but the rest seem like fair game.

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
vegegoku: My goal here was to make sure that the output will remain exactly identical to the original one, there was some implicit sorting and ordering in the original API, this way I was able to compare and make sure I didn't change something I shouldn't or I didn't produce a different output.

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
vegegoku: I would remove the ezt from places except the output file location, because I am not sure if there other references to the file from that specific location.

niloc132 (approved):

PR #9825, opened 2023-04-17
Proposal to update Javadoc to 17

niloc132:
This draft PR is to discuss updating the version of Java we use to produce GWT's Javadoc output, and to capture some of the changes that would make sense.

For a start, I've updated the two links to the JRE's own docs, so that we able to reference newer classes and methods as necessary. Note that this only affects the JRE Emulation at this time, as the source level for gwt-dev and gwt-user are still set to Java 8.

Next, for Java 13+'s search capability, I've re-enabled the `index` feature of Javadoc. It isn't clear to me why we had this disabled to begin with - perhaps we should also re-enable `tree`?

Finally, at least for Java >11 builds, I've removed the --frames flag. If we decide to go with a newer build of Java like this (to support search, etc), we will not be able to use this flag.

The advantage of an older Javadoc build is to keep the "familiar" frames layout, while the newer Javadoc builds give us client-side search.

| Java version | Frames? | Search? | Link |
|--|--|--|--|
|8|yes|no|https://www.gwtproject.org/javadoc/latest/|
|11|yes|no|https://www.colinalworth.com/gwt-javadoc/java11/javadoc/ or https://www.colinalworth.com/gwt-javadoc/java11/javadoc/overview-summary.html to have frames|
|17|no|yes|https://www.colinalworth.com/gwt-javadoc/java17/javadoc/|
|19|no|yes|https://www.colinalworth.com/gwt-javadoc/java19/javadoc/|

JRE Emulation is the same for all builds, but the links now point at Java 11. Example: https://www.colinalworth.com/gwt-javadoc/java17/emul-ezt/fragment.html

---

The only reason I would see to not update to Java 17+ would be to retain the frames view. Is there any reason we might want to keep the frames view, or should we just jump ahead here and assume the latest version going forward?

### DIFF
--- a/build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
+++ b/build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
@@ -53,7 +53,7 @@ import java.util.stream.Stream;
 public class JavaEmulSummaryDoclet implements Doclet {
 
     public static final String OPT_OUTFILE = "-outfile";
-    private static final String JAVADOC_URL = "https://docs.oracle.com/javase/8/docs/api/";
+    private static final String JAVADOC_URL = "https://docs.oracle.com/en/java/javase/11/docs/api/";
 
     private Reporter reporter;
     private String outputFile;
--- a/doc/build.xml
+++ b/doc/build.xml
@@ -83,17 +83,14 @@
              access="package"
              packagenames="${USER_PKGS}"
              sourcefiles="${USER_CLASSES}"
-             noindex="true"
              notree="true"
              use="true"
              windowtitle="GWT Javadoc"
              doctitle="GWT API Reference"
              header="GWT ${gwt.version}"
-             linkoffline="http://download.oracle.com/javaee/6/api/ validation-package-list">
+             linkoffline="https://docs.oracle.com/en/java/javase/11/docs/api/ validation-package-list">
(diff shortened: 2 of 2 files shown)


tbroyer:
> I've re-enabled the `index` feature of Javadoc. It isn't clear to me why we had this disabled to begin with - perhaps we should also re-enable `tree`?

This was only because the generated files were too big for the storage backend for the AppEngine web site.

Fwiw, there is a search in JDK 11 javadoc (e.g. https://docs.oracle.com/en/java/javase/11/docs/api/index.html), that I also see in your java11 links.

Also, I'd very much prefer to ditch the frames, it unnecessarily complicates sharing links to javadocs, having to right-click on a link rather than just copy/paste from the browser's URL bar.

niloc132:
Hmm, not sure how I missed that (though likely from toggling each flag so many times...). My understanding was that this was introduced in Java 13, but I'm happy to be mistaken here.

This doesn't change my vote, I'd rather update to latest lts (currently Java 17). I'd also lean towards re-enabling the tree option (esp as both that and the index feature seem very usable from my modern 25mbps internet connection hosted from the highly advanced "hey nginx, there's a directory, please serve it" provider...).

niloc132:
Having not heard any other complaints, I've turned off notree as well, and left frames and noindex removed. This branch still builds on Java 11 (we can't pass tests yet on Java 17), I'll have a followup for that once javadoc works on 17.

PR #9826, opened 2023-04-18
Deprecate and remove references to RPCServletUtils.CHARSET_UTF8

niloc132:
This was necessary pre-Java7 to avoid blocking on java.nio.charset.Charset operations, though this is unnecessary for our use case through the use of java.nio.charset.StandardCharsets for the default UTF8 charset.

Additionally, this removes a dependency on javax.servlet specific code from generic code that could be shared with a jakarta.servlet implementation.

See #6397
Partial #9727

### DIFF
--- a/user/src/com/google/gwt/user/server/rpc/RPCServletUtils.java
+++ b/user/src/com/google/gwt/user/server/rpc/RPCServletUtils.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.Method;
 import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
 import java.util.Locale;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.zip.GZIPOutputStream;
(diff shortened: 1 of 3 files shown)


niloc132:
Full tests for both Java 8 and 11 are running at https://github.com/niloc132/gwt/actions/runs/4733394641/jobs/8400865873.

tbroyer (commented):
Should the CHARSET_CACHE be removed entirely if the underlying issue has been fixed in the JDK?

I mean, either we admit Charset.forName is now _safe_ and we can remove the CHARSET_CACHE.
Or we only want to cut the dependency from SerializabilityUtil on javax.servlet (transitively through RPCServletUtils) and we can then only change the reference in SerializabilityUtil (and in a separate commit, now or later, refactor RPCServletUtils to account for the fixed behavior of Charset.forName)

niloc132:
It appears to not be fixed in Java 8, so I'd hesitate to remove it entirely while still supporting Java 8 at all. For that reason, I only used the provided static+final field (which will never incur a lock once the class is initialized) in each of the old references.

I didn't confirm by reading Java 8 sources/bytecode, but found a bug report referencing Java 8 build 231 or so which exhibited the behavior, showed a lock. I did scan Java 11 bytecode and confirmed that instead of synchronized, there appear to be transient fields holding arrays to the names/instances of charsets.

tbroyer (approved):
Approved, assuming the PR message will be used as commit message, and the actual commit's message will be discarded when merging (squash and merge? should this be configured as the default strategy then?)

niloc132:
Ah nice, that wasn't an option the last time I looked. Does it auto wrap to 72 char lines now, or do you still have to manually insert those line breaks to be readable?

niloc132:
Note to watchers, it does re-format to wrap at 72 chars automatically! If possible we should more strongly emphasize good PR descriptions. I've been rewriting them to not just format nicely, but as long as I'm at it rewording here and there, but if it'll do this automatically, a good PR message is all the more valuable...

PR #9830, opened 2023-05-18
Run build, tests in Java 17

niloc132:
Full build with Java 8, 11, 17 was run https://github.com/niloc132/gwt/actions/runs/5007780467/jobs/8974836774.

With Javadoc now able to run on Java 17, we can make Java 17 the default going forward for builds. This patch enables running tests in 17, and fixes a few gotchas discovered - an old asm jar was being used for tests, and when we updated to the latest asm for GWT 2.10, we didn't update ClassVisitors to use the newer version.


niloc132:
https://github.com/gwtproject/gwt/pull/9830/commits/ac2df65073bcf1da16866e43909d21c727827cb4 is probably optional - all tests pass without it (there are a few that did need to be changed, but those are in a different commit), but historically when we update asm to a new version, we always update all api versions to latest.

MCMicS (approved):

Issue #9823, opened 2023-04-09
Update copyright entries, style check to reference GWT Project Authors


niloc132:
See https://github.com/gwtproject/gwt/pull/9819#discussion_r1161300098 for discussion, and https://opensource.google/documentation/reference/releasing/authors for specific guidelines from Google in removing their name from all source files, replacing with a reference in a top-level AUTHORS file.

PR #9833, opened 2023-05-24
Update Collections to support changes made in Java 9

### LINKED ISSUE #9547: Java 9, 10, 11 JRE Emulation
Just for the future java 9 support in GWT.

https://blog.codefx.org/java/java-9-stream/

niloc132:
See earlier review at https://gwt-review.googlesource.com/c/gwt/+/21501.

Build changes get us ready for missing Java 10 and 11 emulation as well.

Fixes #9547

### DIFF
--- a/user/super/com/google/gwt/emul/java/util/List.java
+++ b/user/super/com/google/gwt/emul/java/util/List.java
@@ -18,6 +18,7 @@ package java.util;
 import static javaemul.internal.InternalPreconditions.checkNotNull;
 
 import java.util.function.UnaryOperator;
+import javaemul.internal.ArrayHelper;
 
 import jsinterop.annotations.JsIgnore;
 import jsinterop.annotations.JsMethod;
@@ -33,6 +34,69 @@ import jsinterop.annotations.JsType;
  */
 @JsType
 public interface List<E> extends Collection<E> {
+
+  @JsIgnore
+  static <E> List<E> of() {
+    return Collections.emptyList();
+  }
+
+  @JsIgnore
+  static <E> List<E> of(E e1) {
+    return of((E[]) new Object[] {e1});
+  }
+
+  @JsIgnore
+  static <E> List<E> of(E e1, E e2) {
+    return of((E[]) new Object[] {e1, e2});
+  }
+
+  @JsIgnore
+  static <E> List<E> of(E e1, E e2, E e3) {
+    return of((E[]) new Object[] {e1, e2, e3});
+  }
+
+  @JsIgnore
+  static <E> List<E> of(E e1, E e2, E e3, E e4) {
+    return of((E[]) new Object[] {e1, e2, e3, e4});
+  }
+
+  @JsIgnore
+  static <E> List<E> of(E e1, E e2, E e3, E e4, E e5) {
+    return of((E[]) new Object[] {e1, e2, e3, e4, e5});
+  }
+
+  @JsIgnore
+  static <E> List<E> of(E e1, E e2, E e3, E e4, E e5, E e6) {
+    return of((E[]) new Object[] {e1, e2, e3, e4, e5, e6});
+  }
+
+  @JsIgnore
+  static <E> List<E> of(E e1, E e2, E e3, E e4, E e5, E e6, E e7) {
+    return of((E[]) new Object[] {e1, e2, e3, e4, e5, e6, e7});
+  }
+
+  @JsIgnore
+  static <E> List<E> of(E e1, E e2, E e3, E e4, E e5, E e6, E e7, E e8) {
+    return of((E[]) new Object[] {e1, e2, e3, e4, e5, e6, e7, e8});
+  }
+
+  @JsIgnore
+  static <E> List<E> of(E e1, E e2, E e3, E e4, E e5, E e6, E e7, E e8, E e9) {
+    return of((E[]) new Object[] {e1, e2, e3, e4, e5, e6, e7, e8, e9});
+  }
+
+  @JsIgnore
+  static <E> List<E> of(E e1, E e2, E e3, E e4, E e5, E e6, E e7, E e8, E e9, E e10) {
+    return of((E[]) new Object[] {e1, e2, e3, e4, e5, e6, e7, e8, e9, e10});
+  }
+
+  @JsIgnore
+  static <E> List<E> of(E... elements) {
+    for (int i = 0; i < elements.length; i++) {
+      checkNotNull(elements[i]);
+    }
+    return Collections.unmodifiableList(Arrays.asList((E[]) ArrayHelper.unsafeClone(elements, 0, elements.length)));
+  }
 
   @JsMethod(name = "addAtIndex")
   void add(int index, E element);
--- a/user/super/com/google/gwt/emul/java/util/List.java
+++ b/user/super/com/google/gwt/emul/java/util/List.java
@@ -37,7 +37,7 @@ public interface List<E> extends Collection<E> {
 
   @JsIgnore
   static <E> List<E> of() {
-    return Collections.emptyList();
+    return Collections.unmodifiableList(Collections.emptyList());
   }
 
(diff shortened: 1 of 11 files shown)


jhickman (approved):
LGTM

user/super/com/google/gwt/emul/java/util/Map.java
```diff
@@ -35,6 +36,220 @@
...
+  }
+
+  @JsIgnore
+  static <K, V> Map<K, V> ofEntries(Entry<? extends K,? extends V>... entries) {
+    Map<K, V> map = new HashMap<>();
+
+    for (int i = 0; i < entries.length; i++) {
+      Entry<? extends K, ? extends V> entry = checkNotNull(entries[i]);
+      K key = entry.getKey();
+      if (entry instanceof AbstractMap.NonNullableImmutableEntry) {
```
niloc132, line 236: This is debatable - it might save a bit of time at runtime, but increases output size a bit - the instanceof is never optimized out (the type of the array is never tightened to be NonNullableImmutableEntry[] nor does the compiler recognize that all instances are that specific entry type), so we still have the if check.

MCMicS (approved):

user/super/com/google/gwt/emul/java/util/List.java
```diff
@@ -34,6 +35,71 @@
 @JsType
 public interface List<E> extends Collection<E> {
 
+  @JsIgnore
+  static <E> List<E> of() {
+    return Collections.unmodifiableList(Collections.emptyList());
+  }
+
+  @JsIgnore
+  static <E> List<E> of(E e1) {
```
jnehlmeier, line 44: Are these types of methods really needed, given that the emulation does not optimize to avoid creating an array? Do we get these methods for free if we only emulate `of(E... elements)` because JDT will match any other call to that single emulated method then?

niloc132: You're almost certainly right that this would work. My hesitancy is around allowing the compiler to _not_ inline the varargs version of this method. Consider 10 calls to `of(E)` - we would rather a single `function of(e1)` implementation that creates the array literal, the array, and then calls `of(arr)` so that each callsite is simply `a(b)`, rather than 10 callsites that are roughly `a(c(d(e, 1), e, 123, 0, b))` (the required boilerplate to create an array with a class literal, and the element `b`, wrapped in a call to `a(..)`).

jnehlmeier: Hmm I see, interesting point. Should we then communicate this intent in code by adding `@DoNotInline`? Possibly with a comment that this could be removed on some methods if someone provides a specialized implementation for the first couple methods that avoids array creation?

Also I note that all the `of()` methods create an array which is then immediately cloned again in `of(E... )` because that API method can not trust its input and needs to make a defensive, unmodifiable copy. But the array created in all the `of()` factory methods can be used directly as it is known to be not referenced by user code. So maybe we can change the code a bit to

```java
static <E> List<E> of(E e1, E e2, E e3) {
  return ofInternal(new Object[] {e1, e2, e3});
}

@JsIgnore
static <E> List<E> of(E... elements) {
  return ofInternal(ArrayHelper.unsafeClone(elements, 0, elements.length));
}

@SuppressWarnings("unchecked")
private static <E> List<E> ofInternal(Object[] elements) {
  for (int i = 0; i < elements.length; i++) {
    checkNotNull(elements[i]);
  }
  return Collections.unmodifiableList(Arrays.asList((E[]) elements));
}
```

Now the `null` check is later, but I assume that it will only trigger in rare cases anyways as JavaDoc states that `null` elements are forbidden. So I would be fine if in these rare cases an array has been unnecessarily cloned beforehand.

user/test/com/google/gwt/emultest/EmulJava9Suite.java
```diff
@@ -0,0 +1,33 @@
+/*
+ * Copyright 2023 Google Inc.
```
jnehlmeier, line 2: gwtproject authors?

niloc132: #9823 needs to be fixed first, else checkstyle will fail these files, in the meantime we've been just marking these as Google.

user/super/com/google/gwt/emul/java/util/List.java
niloc132: I would assume (but do not know) that the compiler won't just inline without weighing tradeoffs (and for gwt, closure-compiler, almost all tradeoffs are based on output size).

Good catch on the copy (or rather, our absence of a defensive copy), that is definitely unsafe. Unfortunately, we can't have private methods in interfaces while we still support sourceLevel=8, so make our own static method, with `__` prefix or something to make it unlikely to collide? Make a new package-protected class in java.util for this? Other ideas I am not noticing?

jnehlmeier: > Good catch on the copy (or rather, our absence of a defensive copy), that is definitely unsafe.

Hmm maybe you misread: I said that the code does too many defensive copies right now. The object array created internally by all the `of(e1, e2, e3, ..)` methods should not be cloned again in `of(E... elements)`, because we already know that these arrays are effectively immutable. But because `of(E... elements)` is public API and needs to make a defensive copy, so we need an internal method to delegate to.

> Make a new package-protected class in java.util for this?

I would say so. Something like CollectionsHelper. I would put it in https://github.com/gwtproject/gwt/tree/main/user/super/com/google/gwt/emul/javaemul/internal 

If someone wants to optimize the 1-3 parameter factory methods, that helper class could also hold the specialized list/set/map implementations needed.

niloc132: I understood what you said, but mistakenly believed that the current patch had avoided a copy (though of course unsafeClone does clone the array, just in a way that loses the ctm, etc).

I don't love javaemul as a place to put this, since those types are _almost_ only about how to handle mapping js<->java, but it does seem as good of a place as any to handle this. 

ICYMI, the initial review rejected even the 1 param case as having its own implementation.

jnehlmeier: > ICYMI, the initial review rejected even the 1 param case as having its own implementation.

Puh 2019, hehe. But sure he also has a point of not introducing too many implementations  if you just use List.of variants in a small application.

niloc132: > 2019

Yeah, this patch has been backed up behind a long run of other patches to get us building properly on Java >8.

As far as size goes, I think it is very reasonable to keep out other implementations - each class costs something extra, and each List will need its own corresponding Iterator type (or you squander any performance gains you happen to make). The benefits of a specialized type are likely very slim indeed - especially since we're focusing on the use case of needing to iterate a collection that by definition can't have more than 10 items - are you going to be iterating collections of collections in a tight loop? If that's the case... its probably time for some specialized code in your own app to address that performance issue.

Thanks for taking a look, nice to async chat with you again. I'll get this updated in the next day or so.

akbertram:
@niloc132 This is really great. It would also be very helpful to have `Optional.stream()`. Should I open a seperate issue/PR for those emulations?

niloc132:
Thanks Alex, those are actually already done and tested (poke around in my GitHub branches if you want it right away), but just waiting for this one to land so we can make more PRs on the build changes.

user/super/com/google/gwt/emul/java/util/Map.java
niloc132: I can't get this to optimize out cleanly at this time, so I'm going to remove it and replace it with a TODO (as it was in an earlier revision) - if at such a time we can strengthen the compiler's type tightening to handle this case, we can always put it back to generally decrease code size here while maintaining correctness/safety.

niloc132: Reverted.

jnehlmeier (approved):

PR #9837, opened 2023-06-19
Add Java 9 Stream APIs

niloc132:
Partial #9547

### DIFF
--- a/user/super/com/google/gwt/emul/java/util/stream/Collectors.java
+++ b/user/super/com/google/gwt/emul/java/util/stream/Collectors.java
@@ -337,11 +337,6 @@ public final class Collectors {
     return toCollection(ArrayList::new);
   }
 
-  public static <T> Collector<T, ?, List<T>> toUnmodifiableList() {
-    Collector<T, ?, List<T>> mapping = mapping(Objects::requireNonNull, toList());
-    return collectingAndThen(mapping, Collections::unmodifiableList);
-  }
-
   public static <T, K, U> Collector<T, ?, Map<K, U>> toMap(
       final Function<? super T, ? extends K> keyMapper,
       final Function<? super T, ? extends U> valueMapper) {
@@ -358,27 +353,6 @@ public final class Collectors {
       Function<? super T, ? extends U> valueMapper,
       BinaryOperator<U> mergeFunction) {
     return toMap(keyMapper, valueMapper, mergeFunction, HashMap::new);
-  }
-
-  public static <T, K, U> Collector<T, ?, Map<K, U>> toUnmodifiableMap(Function<? super T, ?
-      extends K> keyMapper, Function<? super T, ? extends U> valueMapper) {
-    return collectingAndThen(
-            toMap(disallowNulls(keyMapper), disallowNulls(valueMapper)),
-            Collections::unmodifiableMap
-    );
-  }
-
-  public static <T, K, U> Collector<T, ?, Map<K, U>> toUnmodifiableMap(Function<? super T, ?
-      extends K> keyMapper, Function<? super T, ? extends U> valueMapper, BinaryOperator<U>
-                                                                            mergeFunction) {
-    return collectingAndThen(
-            toMap(disallowNulls(keyMapper), disallowNulls(valueMapper), mergeFunction),
-            Collections::unmodifiableMap
-    );
-  }
-
-  private static  <T, R> Function<T, R> disallowNulls(Function<T, R> func) {
-    return func.andThen(Objects::requireNonNull);
   }
 
   public static <T, K, U, M extends Map<K, U>> Collector<T, ?, M> toMap(
@@ -411,11 +385,6 @@ public final class Collectors {
         s -> s,
         Collector.Characteristics.UNORDERED, Collector.Characteristics.IDENTITY_FINISH
     );
-  }
-
-  public static <T> Collector<T, ?, Set<T>> toUnmodifiableSet() {
-    Collector<T, ?, Set<T>> mapping = mapping(Objects::requireNonNull, toSet());
-    return collectingAndThen(mapping, Collections::unmodifiableSet);
   }
 
   private static <T, D, A> D streamAndCollect(Collector<? super T, A, D> downstream, List<T> list) {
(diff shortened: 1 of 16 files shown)


niloc132:
https://github.com/niloc132/gwt/actions/runs/5315644262/jobs/9624263435 shows successful build in all three tested Java runtime versions.

user/super/com/google/gwt/emul/java/util/stream/Collectors.java
```diff
@@ -307,6 +337,11 @@ public static <T> Collector<T,?,List<T>> toList() {
     return toCollection(ArrayList::new);
   }
 
+  public static <T> Collector<T, ?, List<T>> toUnmodifiableList() {
```
niloc132, line 340: The unmodifiable methods belong in the java 10 patch, with the tests for these methods.

user/super/com/google/gwt/emul/java/util/stream/Collectors.java
```diff
@@ -170,6 +170,34 @@ public static Collector<CharSequence,?,String> joining(CharSequence delimiter) {
         downstream.finisher());
   }
 
+  public static <T, U, A, R> Collector<T, ?, R> flatMapping(Function<? super T,? extends Stream<?
+      extends U>> mapper, Collector<? super U, A, R> downstream) {
+    return new CollectorImpl<>(
+        downstream.supplier(),
+        (A a, T t) -> {
+            mapper.apply(t).forEach(u -> {
+              downstream.accumulator().accept(a, u);
```
zbynek, line 179: from Javadoc

> (If a mapped stream is {@code null} an empty stream is used, instead.)

user/super/com/google/gwt/emul/java/util/stream/DoubleStream.java
```diff
@@ -239,5 +277,19 @@ static DoubleStream of(double t) {
 
   DoubleSummaryStatistics summaryStatistics();
 
+  default DoubleStream takeWhile(DoublePredicate predicate) {
```
zbynek, line 280:
```
DoubleStream.iterate(0.0, d -> d + 2.0)
        .takeWhile(d -> d < 50)
        .boxed()
        .collect(Collectors.toList());
```
should terminate, but with `filter` it won't

niloc132: Ah thank you, good catch. Ill add the example code you provided to ensure that we stop with infinite streams.

user/super/com/google/gwt/emul/java/util/stream/Collectors.java
niloc132: While, you're right, that's what the Javadoc says... whats the point of treating null as empty here but forcing null to throw elsewhere. I don't care for the inconsistency.

Thanks for pointing this out, I'll make the fix and test for it.

user/super/com/google/gwt/emul/java/util/stream/DoubleStream.java
```diff
@@ -162,6 +162,30 @@ public boolean tryAdvance(DoubleConsumer action) {
...
+          public boolean tryAdvance(DoubleConsumer action) {
+            if (terminated) {
+              return false;
+            }
+            if (!hasNext.test(next)) {
+              terminated = true;
+              return false;
+            }
+            action.accept(next);
+            next = f.applyAsDouble(next);
```
zbynek, line 182: when used together with `limit()` or `takeWhile()`, this calls `f` one more time than needed

user/super/com/google/gwt/emul/java/util/stream/DoubleStream.java
niloc132: Been unable to work on this for some time, and this next iteration of the work took longer than I had intended, but I think the added tests confirm that I'm on the right track now, PTAL.

I'm going to look into your other note the next chance I get.

user/super/com/google/gwt/emul/java/util/stream/DoubleStream.java
```diff
@@ -185,6 +209,39 @@ static DoubleStream of(double t) {
 
   DoubleStream distinct();
 
+  default DoubleStream dropWhile(DoublePredicate predicate) {
+    Spliterator.OfDouble prev = spliterator();
+    Spliterator.OfDouble spliterator =
+        new Spliterators.AbstractDoubleSpliterator(
+                prev.estimateSize(), prev.characteristics() & ~(Spliterator.SIZED | Spliterator.SUBSIZED)) {
```
niloc132, line 216: Looks like I've should wrap this (and 7 other lines too).

user/super/com/google/gwt/emul/java/util/stream/DoubleStream.java
niloc132: Looks like `*Stream.iterate(seed, f)` is also wrong, has been since initial commit. I'll add tests and get this fixed, thanks for calling it out.

niloc132: Thanks again for calling this out - I updated tests of both old and new `iterate()`, and made the old version delegate to the new one (which will reduce the size impact of using both methods).

niloc132:
@zbynek so it seems that Java8 has the same bug that GWT had here previously, so my new test above is failing in legacy dev mode.

```java
import java.util.stream.DoubleStream;
import java.util.Arrays;

class Test {
                int[] callCount = { 0 };
                double[] values = DoubleStream.iterate(0, val -> {
                        callCount[0]++;
                        return val + 1;
                }).limit(5).toArray();
                
                System.out.println(callCount[0]);
                System.out.println(Arrays.toString(values));
        }
}
```

In Java 11+, this results in
```
4
[0.0, 1.0, 2.0, 3.0, 4.0]
```
But in Java 8 (both my local copy and the build that the GHA runner is using), this yields:
```
5
[0.0, 1.0, 2.0, 3.0, 4.0]
```
Excitingly, Java 8 only has this bug for IntStream, DoubleStream, and LongStream, just plain Stream has the same behavior in 8 and 11/17:
```java
import java.util.stream.Stream;
import java.util.Arrays;

class Test {
                int[] callCount = { 0 };
                Integer[] values = Stream.iterate(0, val -> {
                        callCount[0]++;
                        return val + 1;
                }).limit(5).toArray(Integer[]::new);
                
                System.out.println(callCount[0]);
                System.out.println(Arrays.toString(values));
        }
}
```
```
4
[0, 1, 2, 3, 4]
```

niloc132:
https://github.com/niloc132/gwt/actions/runs/6792333097 All current tests pass.

zbynek (approved):

PR #9841, opened 2023-06-22
Remove redundant readme instructions

zbynek:
Since elemental was removed from this repo, thementions of python and c++ seem redundant.


niloc132 (approved):
Thanks, agreed it doesn't make sense - the two commands are the same.

PR #9849, opened 2023-08-23
Additional JRE emulation types marked as Serializable

### LINKED ISSUE #9848: Collections.singleton(...) is not Serializable
**GWT version: all** 
**Browser (with version): all** 
**Operating System: all** 

---

##### Description

Collections.singleton(...) is not Serializable 
In Java it is Serializable, so this is just a GWT issue.

On here:
https://github.com/gwtproject/gwt/blob/9b9e888bfc5c8991425da65b0e2f12577d0ac638/user/super/com/google/gwt/emul/java/util/Collections.java#L354

it shows: TODO: make the unmodifiable collections serializable.

##### Steps to reproduce

Try to serialize a collection created using Collections.singleton(...).

niloc132:
Nested types in java.util.Collections and java.util.AbstractMap should implement Serializable.

Fixes #9848

### DIFF
--- a/user/super/com/google/gwt/emul/java/util/AbstractMap.java
+++ b/user/super/com/google/gwt/emul/java/util/AbstractMap.java
@@ -19,6 +19,8 @@ import static javaemul.internal.InternalPreconditions.checkNotNull;
 
 import jsinterop.annotations.JsNonNull;
 
+import java.io.Serializable;
+
 /**
  * Skeletal implementation of the Map interface. <a
  * href="http://java.sun.com/j2se/1.5.0/docs/api/java/util/AbstractMap.html">[Sun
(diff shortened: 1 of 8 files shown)


niloc132:
This is probably insufficient, need to test gwt-rpc and make sure it can handle these, add tests as appropriate.

niloc132:
Note that these tests fail, serializers need to be written. It seems to not be quite sufficient to copy the existing "singleton" impls and just match the JRE's own internal classes - we need to provide special handling for the fact there are two underlying concrete types that can be created:

```
➜  gwt git:(9848-serializable-emul) ✗ jshell
|  Welcome to JShell -- Version 11.0.20
|  For an introduction type: /help intro

jshell> Collections.unmodifiableList(new LinkedList<>()).getClass()
$1 ==> class java.util.Collections$UnmodifiableList

jshell> Collections.unmodifiableList(new ArrayList<>()).getClass()
$2 ==> class java.util.Collections$UnmodifiableRandomAccessList
```

At a glance, nine such implementations will need to be added to `com.google.gwt.user.client.rpc.core.java.util.Collections` to finish this patch.

niloc132:
Closing for now, see https://github.com/gwtproject/gwt/issues/9848#issuecomment-1700294854 for more notes on what remains to be done to get this working. I'll make a smaller PR in a week or so that covers the bare bones of "at least `instanceof` does what it should", but it won't actually make these types serializable using gwt-rpc without more effort/ideas.

Issue #9734, opened 2021-07-22
Uncaught exception handler loses exception object, stack trace, message


niloc132:
**GWT version: 2.8.2, 2.9.0, HEAD
**Browser (with version):** latest Chrome
**Operating System:** Tested on linux, mac, windows

---

##### Description

At some point, Chrome stopped relaying the details of an `onerror` event when the error happens in a script file loaded cross-origin. This does not apply to old-style `$entry`-wrapped stack traces, but only to errors caught by the uncaught exception handler from the gwt iframe onerror handler.

When this occurs, the onerror handler in Impl.java should get the parameters "msg, url, line, column, error", but instead receives "Script error.", "", 0, 0, null for the parameters, so the actual exception cannot be extracted.

Note that this also applies to the "unhandledrejection" event, which should contain the rejected details, but instead doesn't go off at all.

According to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-crossorigin, an attribute can be set on the script tag when it is created to allow the onerror handler to receive the details of the exception when it happens. The server must opt-in as well, as documented in that link, by setting the appropriate CORS headers. Note that the CodeServer class `com.google.gwt.dev.codeserver.WebServer` already correctly sets this header for served JS content (I think at the time this was required for Firefox, so the header was added, but since then Chrome additionally required the attribute on the script tag?):
```
    // Wrap the response to send the extra headers.
    return new Response() {
      @Override
      public void send(HttpServletRequest request, HttpServletResponse response, TreeLogger logger)
          throws IOException {
        // TODO: why do we need this? Looks like Ray added it a long time ago.
        response.setHeader("Access-Control-Allow-Origin", "*");
```

##### Steps to reproduce
The simple case:
* Attach a simple error handler, just GWT.setUncaughtExceptionHandler is sufficient, and log the exception type or message
* In a non-JSNI callback, throw an exception of some kind. If adding an onerror handler manually, be sure to add it to the GWT iframe, or it won't go off at all.
* Load the app from one ~domain~ origin with SuperDevMode, so that the compiled JS is served from a different ~domain~ origin, observe the logged message

Rejected promise:
* Add a "unhandledrejection" event handler to the gwt iframe
* Reject any promise, probably in an event handler of some kind to make it easily reproducible
* Load the app as above - do not set breakpoints in the rejection, sometimes Chrome will end up actually calling the event handler if this happens.

[...]

PR #9854, opened 2023-09-21
Do not set crossOrigin for file protocol

zbynek:
This allows using GWT compiled code with `installCode=false` also with `file://` protocol.

See https://github.com/gwtproject/gwt/issues/9734#issuecomment-1729695795 for the relevant Chrome error message.

### DIFF
--- a/dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptDirect.js
+++ b/dev/core/src/com/google/gwt/core/ext/linker/impl/installScriptDirect.js
@@ -9,7 +9,9 @@ function installScript(filename) {
     var docbody = doc.body;
     var script = doc.createElement('script');
     script.language='javascript';
-    script.crossOrigin='';
+    if (location.host) {
+      script.crossOrigin='';
+    }
     script.src = code;
     if (__MODULE_FUNC__.__errFn) {
       script.onerror = function() {


niloc132:
To confirm, this change makes it possible to open the page, but the stack traces will still not work (since a `file:` url can't opt-in to supporting cross-origin requests)?

zbynek:
After this change I get a Java stacktrace in Chrome and source map seems to work even with 
* file:// protocol used both for JS and document
* localhost used for JS, file:// protocol used for document
but the stacktrace is different compared to using localhost for everything

image
vs
image
Note that the code doesn't use GWT's builtin UncaughtExceptionHandler but a [custom onerror handler](https://github.com/geogebra/geogebra/blob/08a2e4be8b04ea779bd7c5241fff1441802d7258/web-common/src/main/java/org/geogebra/web/html5/util/SuperDevUncaughtExceptionHandler.java#L18) together with [exception unwrapping utility](https://github.com/geogebra/geogebra/blob/08a2e4be8b04ea779bd7c5241fff1441802d7258/gwtutil/src/main/java/org/geogebra/gwtutil/ExceptionUnwrapper.java#L6).

niloc132:
Nice - that's pretty much exactly what was failing that caused us to add this to begin with, glad it isn't necessary here.

niloc132 (approved):

tbroyer:
Correct me if I'm wrong but it looks like there's no stack trace then. What does it look like without the custom error handling code? (I suppose the same without the white text?)

zbynek:
> no stack trace then. 
yep, but still the dev tools show you all the filenames and line numbers.

> What does it look like without the custom error handling
I guess very similar. The custom code is there just to 
1) make sure the native JS exception is printed to the console, not the wrapping Java Exception (just saves you some clicks expanding objects in console)
2) work around https://github.com/gwtproject/gwt-core/pull/18

Bottom line: for file:// protocol the script didn't start before, now it starts. For http:// protocol nothing changed in this PR.

PR #9860, opened 2023-11-10
Add Java 10 Collectors APIs

niloc132:
Partial #9547

### DIFF
--- a/user/test/com/google/gwt/emultest/java10/util/stream/CollectorsTest.java
+++ b/user/test/com/google/gwt/emultest/java10/util/stream/CollectorsTest.java
@@ -1,0 +1,172 @@
+/*
+ * Copyright 2023 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.gwt.emultest.java10.util.stream;
+
+import static com.google.gwt.emultest.java8.util.stream.CollectorsTest.applyItems;
+import static java.util.stream.Collectors.toUnmodifiableList;
+import static java.util.stream.Collectors.toUnmodifiableMap;
+import static java.util.stream.Collectors.toUnmodifiableSet;
+
+import com.google.gwt.emultest.java.util.EmulTestBase;
+
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+/**
+ * Tests for java.util.stream.Collectors Java 10 API emulation.
+ */
+public class CollectorsTest extends EmulTestBase {
+  private static <T> boolean unmodifiableCollection(Collection<T> c, T existingSample, T newSample) {
+    try {
+      c.remove(existingSample);
+      return false;
+    } catch (Exception ignore) {
+      // expected
+    }
+    try {
+      c.add(newSample);
+      return false;
+    } catch (Exception ignore) {
+      // expected
+    }
+    Iterator<T> itr = c.iterator();
+    itr.next();
+    try {
+      itr.remove();
+      return false;
+    } catch (Exception e) {
+      // expected
+    }
+    return true;
+  }
+
+  public void testToUnmodifiableList() {
+    applyItems(List.of("a", "b"), toUnmodifiableList(), "a", "b", (a, b) -> {
+      if (!a.equals(b)) {
+        return false;
+      }
+
+      // check both, so it is obvious we got the right one
+      if (!unmodifiableCollection(a, "a", "z")) {
+        return false;
+      }
+      if (!unmodifiableCollection(b, "a", "z")) {
+        return false;
+      }
+
+      return true;
+    });
+
+    // verify nulls fail
+    try {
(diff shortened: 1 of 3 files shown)


niloc132:
https://github.com/niloc132/gwt/actions/runs/6820138965

FrankHossfeld (approved):
LGTM

user/test/com/google/gwt/emultest/java10/util/stream/CollectorsTest.java
```diff
@@ -0,0 +1,172 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+/**
+ * Tests for java.util.stream.Collectors Java 10 API emulation.
+ */
+public class CollectorsTest extends EmulTestBase {
+  private static <T> boolean unmodifiableCollection(Collection<T> c, T existingSample, T newSample) {
```
jnehlmeier, line 37: It misses a lot of cases. Maybe we should refactor a bit in a follow up commit that introduces a helper class to check immutability of collections/sets, lists, maps. We now have many tests that need it: `CollectionsTest`, `ListTest`, `SetTest`, `MapTest`, `CollectorsTest`. Each doing it a bit differently and not covering everything.

user/test/com/google/gwt/emultest/java10/util/stream/CollectorsTest.java
```diff
@@ -0,0 +1,172 @@
...
+      itr.remove();
+      return false;
+    } catch (Exception e) {
+      // expected
+    }
+    return true;
+  }
+
+  public void testToUnmodifiableList() {
+    applyItems(List.of("a", "b"), toUnmodifiableList(), "a", "b", (a, b) -> {
```
jnehlmeier, line 62: All these tests are fine but a bit difficult to review/read. `applyItems` isn't a great name as it also executes tests using the provided equals function. I had to lookup what `applyItems` actually does. Also using `(a, b)` was a bit distracting as `a` and `b` are used as values as well. Using `expected` and `actual` would probably make more sense. Then we also don't need to test immutability of `expected`.

user/test/com/google/gwt/emultest/java10/util/stream/CollectorsTest.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+    try {
+      Stream.of("a").collect(toUnmodifiableMap(Function.identity(), obj -> null));
+      fail("Expected NPE");
+    } catch (Exception ignore) {
+      // expected
+    }
+  }
+
+  private <K, V> boolean unmodifiableMap(Map<K, V> a, K existingKey, V existingValue, K newKey,
```
jnehlmeier, line 125: Parameter formatting. We don't indent second line that much.

FrankHossfeld: OT question: why aren't we using google formatter for Java?

niloc132: We don't use the Google formatter because Google set this up with a specific format rather than their own format. We have a few other changes (like altering the 'owner' of the code) that would be nice to have, patches definitely welcome here. For my own part, I had intended to leave this alone until we could drop ant, and then use some specific Maven (or Gradle, if someone else was volunteering to help there...) plugin to validate and format.

user/test/com/google/gwt/emultest/java10/util/stream/CollectorsTest.java
niloc132: Happy to have a proposed name change for `applyItems()`, but this has gone through 4ish patches now, and I think the Javadoc is fairly clear - this method applies the collector to the items specified in a way that validates that the Collector instance functions according to its contract. A more accurate name would probably be too unwieldy, and still require that the test author/reader check for themselves what it does (`applyCollectorToItemsWithAndWithoutSplittingAndValidateResultsMatchExpected`?), or we could make it only marginally more clear - but this would still require the reader to actually check the code to understand the implications (`applyCollectorToItemsAndValidate`, `applyCollectorAndValidate`, `validateCollectorBehavior`?)

I'll rename the lambda args.

user/test/com/google/gwt/emultest/java10/util/stream/CollectorsTest.java
niloc132: Good call, I'll file a followup. At least for the `List`-verification method, the simplest answer might be to make `CollectionsTest.doTestModificationsToList` public and static, and lean on it from here.

https://github.com/gwtproject/gwt/issues/9882

niloc132:
Java10 suite passed locally, full build running at https://github.com/niloc132/gwt/actions/runs/7309660201.

user/test/com/google/gwt/emultest/java10/util/stream/CollectorsTest.java
jnehlmeier: >  at some point we have to assume that the levels below us are roughly sane

Yes, I had the thought that it might be good to just make the emulated immutable collection classes public and then simply do an instanceof check. If Collections.unmodifiableList/List.of and friends all use the same implementations, that would be fine I guess. Then we could make one test for the immutable implementations and all other tests check if that implementation has been used.

user/test/com/google/gwt/emultest/java10/util/stream/CollectorsTest.java
jnehlmeier: > applyCollectorToItemsWithAndWithoutSplittingAndValidateResultsMatchExpected
Hehe..

I thought about unwrapping it, e.g.
```
var list1 = collectItemsWithSplitting(collector, items...)
var list2 = collectItemsWithoutSplitting(collector, items...)
assertAllImmutable(list1, list2);
assertAllEqual(expected, list1, list2)
```
or
```
var result = collectItems(collector, items...)
assertAllImmutable(result.withSplitting, result.withoutSplitting);
assertAllEqual(expected, result.withSplitting, result.withoutSplitting)
```

jnehlmeier (approved):

FrankHossfeld (approved):

PR #9861, opened 2023-11-10
Updated Enumeration emul for Java 9

niloc132:
Partial #9547


niloc132:
This seems to be the obvious implementation for https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Enumeration.html#asIterator() based on the implementation notes. As far as I can tell we don't have any implementations of Enumeration in GWT, just adapters that transform other Iterators into Enumerations.

jnehlmeier (approved):

zbynek (approved):

PR #9867, opened 2023-11-16
Fix for namespace="<window>" names that clash with variable names

### LINKED ISSUE #9573: Name conflict between JS method name and local variable
**GWT version:** 2.8.2 (Maven repository)

**Browser (with version):** Chrome 62.0.3202.94

**Operating System:** Windows

---

##### Description
Call to Double.isNaN fails if local variable isNaN is present. Tested with following code
```java
@Override
public void onModuleLoad() {
	boolean isNaN = Double.isNaN(42);
	GWT.log("" + isNaN);
}
```

##### Steps to reproduce
Compile a project with the entry point above, it will fail to initiate (both normal and super dev mode).  When compiled with DETAILED style you can see that isNaN on right hand side of the assignment is also replaced with the local variable. 

##### Known workarounds
Rename the local variable.

niloc132:
For example:

  @JsMethod(namespace = "<window>")
  static native boolean isNaN();

  private void test() {
    boolean isNaN = isNaN();
  }

will incorrectly infer that the method isNaN and the variable isNaN
refer to the same program element and use the name assigned to the
variable for both.

Fixes #9573


niloc132:
Passing nightly builds https://github.com/niloc132/gwt/actions/runs/6885094990

PR #9875, opened 2023-12-12
Relax naming rules to allow valid java bean properties as jsproperties

### LINKED ISSUE #9554: JsType properties starting with dollar are not accepted
**GWT version:** 2.8.1
**Browser (with version):** N/A
**Operating System:** N/A

---

##### Description

GWT does not allow an @JsProperty accessor method if the property starts with '$'.

##### Steps to reproduce

Just try to compile:
```
@JsType
public class Test {
  @JsProperty
  public native String get$1();
}
```

##### Known workarounds

Must set @JsProperty name attribute to be accepted. E.g.:
```
  @JsProperty(name="$1")
  public static native String get$1();
```

##### Discussion

  I understand this is not a priority issue, but I've just tested in the JVM using JavaBeans Introspector:

```
package jbeans.test;

public class TestBean {
	private String name;
	private String $dollar;
	private String _underscore;
	private String $0;
	private String $1;
	
	public String getName() {
		return name;
	}
	
	public void setName(String name) {
		this.name = name;
	}
	
	public void set$dollar(String $dollar) {
		this.$dollar = $dollar;
	}
	
	public String get$dollar() {
		return $dollar;
	}
	
	public String get_underscore() {
		return _underscore;
	}
	
	public void set_underscore(String _underscore) {
		this._underscore = _underscore;
	}
	
	public void set$0(String $0) {
		this.$0 = $0;
	}
	
	public String get$0() {
		return $0;
	}
	
	public String get$1() {
		return $1;
	}
	
	public void set$1(String $1) {
		this.$1 = $1;
	}
}

package jbeans.test;

import java.beans.BeanInfo;
import java.beans.IntrospectionException;
import java.beans.Introspector;
import java.beans.PropertyDescriptor;

public class Main {
		try {
			BeanInfo beanInfo = Introspector.getBeanInfo(TestBean.class);

			for (PropertyDescriptor props : beanInfo.getPropertyDescriptors()) {
				System.out.println("Property: " + props.getName());
			}

		} catch (IntrospectionException e) {
			// TODO Auto-generated catch block
			e.printStackTrace();
		}
	}
}

```
outputs:

```
Property: $0
Property: $1
Property: $dollar
Property: _underscore
Property: class
Property: name
```

niloc132:
The old implementation only allowed a capital letter as the first character of a bean-like property, which isn't sufficient. Instead, we need to support cases where the first character is *not* lower case, so that non-letters that are valid for the first character are permitted.

Fixes #9554

### DIFF
--- a/dev/core/src/com/google/gwt/dev/jjs/ast/HasJsInfo.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/ast/HasJsInfo.java
@@ -117,7 +117,7 @@ public interface HasJsInfo extends HasJsName, CanBeJsNative {
 
     private static boolean startsWithCamelCase(String string, String prefix) {
       return string.length() > prefix.length() && string.startsWith(prefix)
-          && Character.isUpperCase(string.charAt(prefix.length()));
+          && !Character.isLowerCase(string.charAt(prefix.length()));
     }
   }
 
--- a/dev/core/src/com/google/gwt/dev/js/JsUtils.java
+++ b/dev/core/src/com/google/gwt/dev/js/JsUtils.java
@@ -480,7 +480,7 @@ public class JsUtils {
   /**
    * A JavaScript identifier contains only letters, numbers, _, $ and does not begin with a number.
    * There are actually other valid identifiers, such as ones that contain escaped Unicode
-   * characters but we disallow those for the time being.
+   * characters, but we disallow those for the time being.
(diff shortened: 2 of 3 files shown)


niloc132:
Full nightly build to confirm no other tests break: https://github.com/niloc132/gwt/actions/runs/7188011833

Issue #9709, opened 2020-09-17
Java Deserialization vulnerability in GWT-RPC


Medo42:
I just stumbled upon these discussions while looking for information on how the "enhanced classes" mechanism works:

https://groups.google.com/forum/#!msg/google-web-toolkit/j36D9-11JF4/OZwNQgvSAgAJ
https://groups.google.com/forum/#!topic/google-web-toolkit-contributors/Jmo5qFoaw2M

In short, enhanced fields are serialized using Java Object serialization, and when the client sends back that serialized data, the server will deserialize it. Since there does not appear to be any safeguard (like a signature) to ensure that it has not been tampered with on the client-side, the client could inject any data it wants and the server will attempt to deserialize it.

Deserializing untrusted input is considered a security vulnerability, see e.g. https://owasp.org/www-community/vulnerabilities/Deserialization_of_untrusted_data

Unfortunately, I was not able to find any clue that these discussions actually led to improvements in the GWT RPC implementation, and a brief search in the current master branch shows that the relevant code (https://github.com/gwtproject/gwt/blob/master/user/src/com/google/gwt/user/server/rpc/impl/ServerSerializationStreamReader.java#L804) seems to be unchanged since 2013.

Further, I'd argue that the implementation in GWT-RPC suffers from an additional vulnerability, since it allows an attacker to override the value of any field declared on the enhanced class, even transient and static fields. I think it is safe to say that application authors will not expect the possiblity that static or transient fields might be affected by deserialization, so this provides additional attack surface even in the absence of a gadget that would allow exploiting the Java Object Deserialization.

### Edit to make a response prominent: 
Mitigation for this has [been released in GWT 2.11.0](https://github.com/gwtproject/gwt/releases/tag/2.11.0) and is [backported to a GWT 2.10.1 release](https://github.com/gwtproject/gwt/releases/tag/2.10.1) - this will disable this feature by default, showing warnings when compiling, and refusing to start the servlet unless the author explicitly opts-in to re-enabling it. This allows projects running those versions to be confident that they are not using this feature and thus unaffected by this bug.

The issue https://github.com/gwtproject/gwt/issues/9880 is filed to fix this by signing payloads when sent from the server. Additionally, https://github.com/gwtproject/gwt/issues/9881 is filed to disable attempting to serialize such payloads even if a class is annotated with JPA/JDO annotations. At present, both issues are waiting for someone to take them up, either by developing them, or to sponsor someone else to complete them.

[...]

PR #9879, opened 2023-12-19
Disable rpc.enhancedClasses by default at runtime

niloc132:
Logs warnings at compile time, indicating which classes need to be cleaned up to remove this feature.

Mitigation for #9709

### DIFF
--- a/user/src/com/google/gwt/user/rebind/rpc/SerializableTypeOracleBuilder.java
+++ b/user/src/com/google/gwt/user/rebind/rpc/SerializableTypeOracleBuilder.java
@@ -869,8 +869,8 @@ public class SerializableTypeOracleBuilder {
 
       if (tic.maybeEnhanced()
           || (enhancedClasses != null && enhancedClasses.contains(type.getQualifiedSourceName()))) {
-        logger.log(TreeLogger.WARN, "The class " + type.getQualifiedSourceName() + " has JPA " +
-                "annotations or is explicitly configured as an enhanced class using the " +
+        logger.log(TreeLogger.WARN, "The class " + type.getQualifiedSourceName() + " has " +
+                "JPA/JDO annotations or is explicitly configured as an enhanced class using the " +
                 "configuration property rpc.enhancedClasses. This makes the server vulnerable " +
                 "to an issue with deserialization of unsafe data. See " +
                 "https://github.com/gwtproject/gwt/issues/9709 for more information.");
--- a/user/src/com/google/gwt/user/rebind/rpc/SerializableTypeOracleBuilder.java
+++ b/user/src/com/google/gwt/user/rebind/rpc/SerializableTypeOracleBuilder.java
@@ -869,6 +869,10 @@ public class SerializableTypeOracleBuilder {
 
       if (tic.maybeEnhanced()
           || (enhancedClasses != null && enhancedClasses.contains(type.getQualifiedSourceName()))) {
+        logger.log(TreeLogger.WARN, "The class " + type.getQualifiedSourceName() + " is both " +
+                "referenced from configuration as rpc.enhancedClasses and has JPA annotations. " +
+                "This makes the server vulnerable to an issue with deserialization of unsafe " +
+                "data. See https://github.com/gwtproject/gwt/issues/9709 for more information.");
         type.setEnhanced();
       }
     }
--- a/user/src/com/google/gwt/user/rebind/rpc/SerializableTypeOracleBuilder.java
+++ b/user/src/com/google/gwt/user/rebind/rpc/SerializableTypeOracleBuilder.java
@@ -320,7 +320,7 @@ public class SerializableTypeOracleBuilder {
       JPA_ENTITY_ANNOTATION =
           Class.forName("javax.persistence.Entity").asSubclass(Annotation.class);
     } catch (ClassNotFoundException e) {
-      // Ignore, JPA_ENTITY_CAPABLE_ANNOTATION will be null
+      // Ignore, JPA_ENTITY_CAPABLE_ANNOTATION will be null
     }
   }
 
@@ -869,10 +869,11 @@ public class SerializableTypeOracleBuilder {
 
       if (tic.maybeEnhanced()
           || (enhancedClasses != null && enhancedClasses.contains(type.getQualifiedSourceName()))) {
-        logger.log(TreeLogger.WARN, "The class " + type.getQualifiedSourceName() + " is both " +
-                "referenced from configuration as rpc.enhancedClasses and has JPA annotations. " +
-                "This makes the server vulnerable to an issue with deserialization of unsafe " +
-                "data. See https://github.com/gwtproject/gwt/issues/9709 for more information.");
+        logger.log(TreeLogger.WARN, "The class " + type.getQualifiedSourceName() + " has JPA " +
+                "annotations or is explicitly configured as an enhanced class using the " +
+                "configuration property rpc.enhancedClasses. This makes the server vulnerable " +
+                "to an issue with deserialization of unsafe data. See " +
+                "https://github.com/gwtproject/gwt/issues/9709 for more information.");
         type.setEnhanced();
       }
     }
--- a/user/src/com/google/gwt/user/server/rpc/RemoteServiceServlet.java
+++ b/user/src/com/google/gwt/user/server/rpc/RemoteServiceServlet.java
@@ -16,6 +16,8 @@
 package com.google.gwt.user.server.rpc;
 
 import static com.google.gwt.user.client.rpc.RpcRequestBuilder.MODULE_BASE_HEADER;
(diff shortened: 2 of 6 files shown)


niloc132:
The good news is that all tests passed with this change https://github.com/niloc132/gwt/actions/runs/7267798597

The bad news is that this means that the `rpc.enhancedClasses` feature apparently has no tests - probably only bad news in retrospect, given that we are trying to turn it off.

niloc132:
In light of https://github.com/gwtproject/gwt/issues/9709#issuecomment-1863677096, it might be wise to generalize these error messages slightly, and if the developer doesn't believe that JPA/JDO is in use, to just remove the annotations and see if that still works.

user/src/com/google/gwt/user/rebind/rpc/SerializableTypeOracleBuilder.java
```diff
@@ -869,6 +869,10 @@ public SerializableTypeOracle build(TreeLogger logger) throws UnableToCompleteEx
 
       if (tic.maybeEnhanced()
           || (enhancedClasses != null && enhancedClasses.contains(type.getQualifiedSourceName()))) {
+        logger.log(TreeLogger.WARN, "The class " + type.getQualifiedSourceName() + " is both " +
+                "referenced from configuration as rpc.enhancedClasses and has JPA annotations. " +
```
jnehlmeier, line 873: The message should be `...has JPA/JDO annotations or is explicitly configured as enhanced class using configuration property ...`

user/src/com/google/gwt/user/server/rpc/RemoteServiceServlet.java
```diff
@@ -94,6 +96,20 @@ static SerializationPolicy loadSerializationPolicy(HttpServlet servlet,
...
             serializationPolicy = SerializationPolicyLoader.loadFromStream(is,
                 null);
+            if (serializationPolicy.hasClientFields()) {
+              if (ENABLE_ENHANCED_CLASSES) {
+                servlet.log("WARNING: Enhanced JPA client fields are in use for this " +
+                        "application. See https://github.com/gwtproject/gwt/issues/9709 for " +
+                        "more detail. on the vulnerability that this presents.");
+              } else {
+                servlet.log("ERROR: Service uses enhanced classes, which are unsafe. Review " +
+                        "build logs to see where this can be fixed, or set " +
```
jnehlmeier, line 106: Little more explicit:

`Service de-/serializes enhanced JPA/JDO classes which is unsafe. Review build logs to see which classes are affected, ..`

user/src/com/google/gwt/user/server/rpc/RemoteServiceServlet.java
```diff
@@ -94,6 +96,20 @@ static SerializationPolicy loadSerializationPolicy(HttpServlet servlet,
           try {
             serializationPolicy = SerializationPolicyLoader.loadFromStream(is,
                 null);
+            if (serializationPolicy.hasClientFields()) {
+              if (ENABLE_ENHANCED_CLASSES) {
+                servlet.log("WARNING: Enhanced JPA client fields are in use for this " +
+                        "application. See https://github.com/gwtproject/gwt/issues/9709 for " +
```
jnehlmeier, line 102: `Service de-/serializes enhanced JPA/JDO classes which is unsafe. See ...`

user/src/com/google/gwt/user/server/rpc/RemoteServiceServlet.java
```diff
@@ -94,6 +96,20 @@ static SerializationPolicy loadSerializationPolicy(HttpServlet servlet,
...
+                        "more detail. on the vulnerability that this presents.");
```
jnehlmeier, line 103: `...for more detail on ...`

user/src/com/google/gwt/user/rebind/rpc/SerializableTypeOracleBuilder.java
```diff
@@ -869,6 +869,11 @@ public SerializableTypeOracle build(TreeLogger logger) throws UnableToCompleteEx
 
       if (tic.maybeEnhanced()
           || (enhancedClasses != null && enhancedClasses.contains(type.getQualifiedSourceName()))) {
+        logger.log(TreeLogger.WARN, "The class " + type.getQualifiedSourceName() + " has JPA " +
```
jnehlmeier, line 872: To be consistent JPA/JDO should be used here as well.

jnehlmeier (approved):

user/src/com/google/gwt/user/server/rpc/RemoteServiceServlet.java
```diff
@@ -94,6 +96,20 @@ static SerializationPolicy loadSerializationPolicy(HttpServlet servlet,
           try {
             serializationPolicy = SerializationPolicyLoader.loadFromStream(is,
                 null);
+            if (serializationPolicy.hasClientFields()) {
+              if (ENABLE_ENHANCED_CLASSES) {
+                servlet.log("WARNING: Service deserializes enhanced JPA/JDO classes, which is " +
+                        "unsafe. See https://github.com/gwtproject/gwt/issues/9709 for more " +
+                        "detail on the vulnerability that this presents.");
```
FrankHossfeld, line 103: it should be: "details on the vulnerability"

jnehlmeier: As German I tend to use `details` because that is how we speak natively but English also has "more detail on" if you want to emphasize that some content exists which provides a more granular version of what has already been said. So I think `for more detail on` is correct here as the previous sentence has given a very condensed summary. I think "more details on" would be more appropriate if the text has already given some details but not all.
But as I am not a native speaker I am fine with whatever @niloc132 choses :)

FrankHossfeld: yeah, as German I would say 'details'. Checked also on Google both is ok. It is up to @niloc132 :-)

FrankHossfeld (approved):

user/src/com/google/gwt/user/server/rpc/RemoteServiceServlet.java
niloc132: Both "sound" correct to me - it appears that "for more detail" would be incorrect, but "for more details" or "for more detail on" are both acceptable? I'm going to leave this as is.

niloc132:
Testing to confirm the new behavior:

1. Started with the DynaTable example, confirmed it started and worked correctly
2. Added `<extend-configuration-property name="rpc.enhancedClasses" value="com.google.gwt.sample.dynatable.client.Person" />` to the .gwt.xml file
3. Confirmed that the compiler logged a warning:
    ```
    [java]             Rebinding com.google.gwt.sample.dynatable.client.SchoolCalendarService
    [java]                Invoking generator com.google.gwt.user.rebind.rpc.ServiceInterfaceProxyGenerator
    [java]                   Generating client proxy for remote service interface 'com.google.gwt.sample.dynatable.client.SchoolCalendarService'
    [java]                      [WARN] The class com.google.gwt.sample.dynatable.client.Person has JPA/JDO annotations or is explicitly configured as an enhanced class using the configuration property rpc.enhancedClasses. This makes the server vulnerable to an issue with deserialization of unsafe data. See https://github.com/gwtproject/gwt/issues/9709 for more information.
    ```
4. Confirmed that the server would not load the policy and logged a server error:
    ```
    [java] 2023-12-23 12:24:51.073:INFO:oejshC.ROOT:qtp559361838-84: calendar: ERROR: Service deserializes enhanced JPA/JDO classes, which is unsafe. Review build logs to see which classes are affected, or set gwt.enhancedClasses.enabled to true to allow using this service. See https://github.com/gwtproject/gwt/issues/9709 for more detail.
    [java] 2023-12-23 12:24:51.074:INFO:oejshC.ROOT:qtp559361838-84: calendar: WARNING: Failed to get the SerializationPolicy 'D08326A5FC5842E352DF103522C617ED' for module 'http://127.0.0.1:8888/dynatable/'; a legacy, 1.3.3 compatible, serialization policy will be used.  You may experience SerializationExceptions as a result.
    ```
5. Note also that the client could not load data correctly with the legacy policy, and had its own error.
6. Added `-Dgwt.enhancedClasses.enabled=true` to the JVM args, confirmed that the above error was replaced by a warning and the application worked:
    ```
    [java] 2023-12-23 12:38:39.032:INFO:oejshC.ROOT:qtp448000133-82: calendar: WARNING: Service deserializes enhanced JPA/JDO classes, which is unsafe. See https://github.com/gwtproject/gwt/issues/9709 for more detail on the vulnerability that this presents.
    ```
7. Next, removed the configuration property, and instead added `@javax.persistence.Entity` to the Person bean, `org.datanucleus:javax.persistence:2.2.3` added to `war/WEB-INF/lib/`, and repeated the test from step 3.

Issue #9863, opened 2023-11-15
Proposal: Deprecate DevMode usage without -noserver


niloc132:
For many years, it has been suggested that non-trivial applications should bring their own server, and not rely on the Jetty instance built-in to GWT's DevMode to host their app, even for development purposes. There are many reasons to say this, and with a new one arising, it may be time to formally deprecate this.

The `javax.servlet` -> `jakarta.servlet` transition makes it hard for a single container to serve both old and new applications. Inevitably, there will still be `javax.servlet` applications running for many years, even though most app servers have deprecated or end-of-life'd their implementations by now.

Jetty in particular is interesting for us, as GWT DevMode and CodeServer are based on Jetty. Their versioning policy has Jetty 9, the version that GWT currently uses, at end-of-life. Jetty 10 will probably be maintained for at least a few more years, and while it supports `javax.servlet`, it does require Java 11. Jetty 11 and 12 are both stable releases as well, but require `jakarta.servlet`. This deprecation would remove the need to offer a migration timeline/path for DevMode to transition to `jakarta.servlet`.

---

This issue proposes to deprecate usage of DevMode as an application server, potentially moving to a default of `-noserver`, and potentially removing `-server` entirely in the future. Unlike some other deprecations, this should cause no loss in functionality for end users, as (I dearly, dearly hope) no application in the world is actually deployed on DevMode, but already has an appropriate application server that it can use.

My proposed implementation for GWT 2.11 is to log once on startup if an application is being served. The simplest check here could be a filter on `/*` - any traffic at all triggers a single warning message - this would not log until such a call was made. Alternatively if we wanted to allow only trivial "static files only" apps, we could query the JettyLauncher's ServletHolder to see if any servlets or filters were configured, and if so, log at the end of startup that this usage was deprecated.

A future implementation where `-noserver` was the default could instead warn if `-server` was specified at all.

tbroyer:
Fwiw, I'd rather deprecate the use of servlets only, and at one point move to remove servlet support altogether, only supporting static files (possibly with a "catch-all" mode that returns the `index.html` for all URLs; if anyone finds that useful).

IMO, deprecating `-server` altogether would be equivalent to deprecating DevMode and asking everyone to use CodeServer, and much simpler to implement 😉

eliasbalasis:
> IMO, deprecating `-server` altogether would be equivalent to deprecating DevMode and asking everyone to use CodeServer, and much simpler to implement 😉

[...]

PR #9883, opened 2023-12-29
Deprecate webAppCreator, encourage using the getting started guide

niloc132:
This encourages a few bad practices that we'd like to avoid, most notably mixing server and client classpaths, and encouraging using DevMode's built-in server as an application server.

Partial #9863


FrankHossfeld (approved):
Great idea pointing to the website. So we need to make sure, that the gettingstarted.html page on gwtproject.org is updated.

tbroyer (approved):

Issue #9869, opened 2023-12-06
Epic: Add Java 17 language support


niloc132:
This will be one of the leading features for GWT 2.12 - we've experimentally built out more than half of the new features already. The limiting factor is that in order to support a newer JDT, we will need to raising support for running Compiler/DevMode/CodeServer/JUnitShell to Java 11.

Subtasks:
- [x] #9870
- [ ] Optional: Explore CI testing for running server classes on Java 8. Will need additional resourcing/testing - this doesn't appear to be a priority for the vast majority of users. An alternative could be to backport fixes to release/2.11 as requested.
- [x] #9871
- [x] #9859
- [x] Add tests for Java 15 text blocks https://openjdk.org/jeps/378
- [x] Add tests for Java 17 sealed classes https://openjdk.org/jeps/409
- [x] #9894
- [x] #9893
- [x] #9878

JRE emulation changes, if any, should be filed in a different task. While reviewing JEPs, I only noticed https://openjdk.org/jeps/356, which may not apply anyway.

jnehlmeier:
Shouldn't we try to aim 21 directly? Eclipse 2023-12 seems to support it so I guess some JDT should exist?

niloc132:
@jnehlmeier 
For the JDT, if already stable, with support for all features, and still supports running on Java 11 I'd agree. Otherwise, I'd personally vote for incrementally building these features and using this issue as a milestone for Java 17, with a follow up milestone (likely for a later release?) for Java 21.

Besides the reasons above, we don't have a lot of PR reviewers anyway, I wouldn't want to say that we can't ship GWT 2.12 until all of 17 and 21 are finished.

niloc132:
Looks like there are just two non-preview language JEPs in 18/19/20/21: https://openjdk.org/jeps/440 and https://openjdk.org/jeps/441. I haven't checked an updated JDT yet, but at least 441 looks like it will need codegen changes.

niloc132:
Closing as completed.

PR #9884, opened 2023-12-31
Point to the latest JDT and asm

### LINKED ISSUE #9871: Update JDT to support Java 17 language features
In earlier testing, JDT 3.27.50 was the first version that had all Java 17 features - we used a beta release for this work. We should update to the latest stable JDT, in an attempt to also resolve bugs found in testing since that version was released.

Likely will also require updating asm.

This ticket will also include testing for the Java 12-17 language features that do not require changes to GWT. This probably includes:
 * Java 15 text blocks https://openjdk.org/jeps/378
 * Java 17 sealed classes https://openjdk.org/jeps/409

If any of those need GWT changes, we'll pull it out into its own ticket.

vegegoku:
Update build scripts to point to the latest JDT [3.36.0.v20231115-1055] and the latest asm 9.6

fix gwtproject/gwt#9871
related to gwtproject/gwt#9869

### DIFF
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -29,7 +29,8 @@ jobs:
       - name: Checkout GWT tools into a sibling directory
         uses: actions/checkout@v4
         with:
-          repository: 'gwtproject/tools'
+          repository: 'Vertispan/tools'
+          ref: 'errorprone-2.23'
           path: 'tools'
       - name: Set up JDK ${{ matrix.java-version }}
         # GWT requires Java 11+ to build
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -1,4 +1,4 @@
-# Run all tests and builds all aspects of GWT using Java 11 and 17. Runs
+# Run all tests and builds all aspects of GWT using Java 8, 11, and 17. Runs
 # nightly (plus or minus timzeones) on the main branch, and will also run right
 # away on a push to a release branch. Release zips are uploaded as part of the
 # build, though maven snapshots are not yet deployed.
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java-version: [ '11', '17' ]
+        java-version: [ '11', '17', '21' ]
     steps:
       - name: Checkout GWT itself into one directory
         uses: actions/checkout@v4
@@ -83,7 +83,7 @@ jobs:
       - name: Set up sonatype credentials
         # Using the same java version as above, set up a settings.xml file
         uses: actions/setup-java@v4
-        if: ${{ github.event_name == 'schedule' && github.repository_owner == 'gwtproject' && matrix.java-version == '17' }}
+        if: ${{ github.event_name == 'schedule' && github.repository_owner == 'gwtproject' && matrix.java-version == '21' }}
         with:
           java-version: ${{ matrix.java-version }}
           distribution: 'temurin'
@@ -93,7 +93,7 @@ jobs:
           server-password: SONATYPE_PASSWORD
 
       - name: Nightly builds should be deployed as snapshots to sonatype
-        if: ${{ github.event_name == 'schedule' && github.repository_owner == 'gwtproject' && matrix.java-version == '17' }}
+        if: ${{ github.event_name == 'schedule' && github.repository_owner == 'gwtproject' && matrix.java-version == '21' }}
         env:
           SONATYPE_USERNAME: ${{ secrets.SONATYPE_USERNAME }}
           SONATYPE_PASSWORD: ${{ secrets.SONATYPE_PASSWORD }}
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -29,8 +29,8 @@ jobs:
       - name: Checkout GWT tools into a sibling directory
         uses: actions/checkout@v4
         with:
-          repository: 'Vertispan/tools'
-          ref: 'errorprone-2.23'
+          repository: 'vegegoku/tools'
+          ref: 'java-17-support'
           path: 'tools'
       - name: Set up JDK ${{ matrix.java-version }}
         # GWT requires Java 11+ to build
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -52,7 +52,7 @@ jobs:
             GWT_VERSION=HEAD-SNAPSHOT
           # Run the ant tasks, disabling watchFileChanges to work around a github actions limitation
           ant clean test dist doc \
-            -Dtest.jvmargs='-ea -Dgwt.watchFileChanges=false' \
+            -Dtest.jvmargs='-ea -noverify -Dgwt.watchFileChanges=false' \
             -Dtest.web.htmlunit.disable=true \
             -Dtest.nometa.htmlunit.disable=true \
             -Dtest.emma.htmlunit.disable=true
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java-version: [ '11', '17', '21' ]
+        java-version: [ '11', '17' ]
     steps:
       - name: Checkout GWT itself into one directory
         uses: actions/checkout@v4
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -1,4 +1,4 @@
-# Run all tests and builds all aspects of GWT using Java 8, 11, and 17. Runs
+# Run all tests and builds all aspects of GWT using Java 8, 11, 17, and 21. Runs
 # nightly (plus or minus timzeones) on the main branch, and will also run right
 # away on a push to a release branch. Release zips are uploaded as part of the
 # build, though maven snapshots are not yet deployed.
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java-version: [ '11', '17' ]
+        java-version: [ '11', '17', '21' ]
     steps:
       - name: Checkout GWT itself into one directory
         uses: actions/checkout@v4
@@ -29,8 +29,7 @@ jobs:
       - name: Checkout GWT tools into a sibling directory
         uses: actions/checkout@v4
         with:
-          repository: 'vegegoku/tools'
-          ref: 'java-17-support'
+          repository: 'gwtproject/tools'
           path: 'tools'
       - name: Set up JDK ${{ matrix.java-version }}
         # GWT requires Java 11+ to build
@@ -52,7 +51,7 @@ jobs:
             GWT_VERSION=HEAD-SNAPSHOT
           # Run the ant tasks, disabling watchFileChanges to work around a github actions limitation
           ant clean test dist doc \
-            -Dtest.jvmargs='-ea -noverify -Dgwt.watchFileChanges=false' \
+            -Dtest.jvmargs='-ea -Dgwt.watchFileChanges=false' \
             -Dtest.web.htmlunit.disable=true \
             -Dtest.nometa.htmlunit.disable=true \
             -Dtest.emma.htmlunit.disable=true
--- a/dev/build.xml
+++ b/dev/build.xml
@@ -63,11 +63,11 @@
           <include name="apache/tapestry-util-text-4.0.2.jar"/>
           <include name="apache/ant-zipscanner/ant-zipscanner-1.6.5-1-rebased.jar"/>
           <include name="colt/colt-1.2.jar"/>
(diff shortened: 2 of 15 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@
 /build/
 # Generated poms
 /maven/poms/**/pom.xml
+/.idea/
```
niloc132, line 9: See the comment at the top of the file and the link on gwtproject.org - maybe the readme should point that out too?

niloc132:
Do you think it would make sense to also update SourceLevel to accept 17 as a value, and pass that to JDT? That should at least get us text blocks and sealed classes "for free" (though of course we still need tests).

I think that makes sense to include in this PR, and give that a quick test so that we know that the new jdt actually does something.

jnehlmeier:
This version should also already support Java 21 (except some preview features).

https://github.com/eclipse-jdt/eclipse.jdt.core/issues/890

Most interesting would be record patterns and pattern matching for switch.

niloc132:
@jnehlmeier unfortunately Ahmad has found that builds of JDT past 4.27 require a minimum of Java 17 to compile and run, so if we're dropping just Java 8 for now, we can't update to 4.29. Additionally, there is something funny happening with a new required dependency of the jdt core, and shading it into the gwt-dev.jar seems to prevent classes from being loaded in that jar correctly. 

If you want to push for going all the way to Java 21 now, you could take that up, discuss dropping both Java 8 and 11 on the mailing list, and work up a fix for the new batch dependency? Otherwise I continue to be in favor of a more incremental approach (especially as record/pattern matching is likely going to require more custom codegen to make sense in js).

dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
```diff
@@ -537,13 +537,20 @@ public void endVisit(BreakStatement x, BlockScope scope) {
 
     @Override
     public void endVisit(CaseStatement x, BlockScope scope) {
+      assert x.isExpr == false : "Switch expressions not yet supported";
```
niloc132, line 540: Let's make this a Precondition instead, so it doesn't get skipped if asserts are disabled? 

(Note that this is feedback on my own original patch, I wrote this in the first draft, intending to do all testing with asserts enabled, but that isn't what a real user would be likely to do.)

dev/core/test/com/google/gwt/dev/jjs/impl/Java17AstTest.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.gwt.dev.javac.testing.impl.JavaResourceBase;
+import com.google.gwt.dev.jjs.ast.JInterfaceType;
+import com.google.gwt.dev.jjs.ast.JProgram;
+import com.google.gwt.dev.jjs.ast.JStringLiteral;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Tests that {@link GwtAstBuilder} correctly builds the AST for
+ * features introduced in Java 11.
```
niloc132, line 25:
```suggestion
 * features introduced in Java 17.
```

.github/workflows/full-check.yml
```diff
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java-version: [ '8', '11', '17' ]
+        java-version: [ '11', '17' ]
```
niloc132, line 23: Can you re-add 21 in your branch? with the asm update, this should work correctly.

.github/workflows/full-check.yml
```diff
@@ -83,7 +84,7 @@ jobs:
       - name: Set up sonatype credentials
         # Using the same java version as above, set up a settings.xml file
         uses: actions/setup-java@v4
-        if: ${{ github.event_name == 'schedule' && github.repository_owner == 'gwtproject' && matrix.java-version == '17' }}
+        if: ${{ github.event_name == 'schedule' && github.repository_owner == 'gwtproject' && matrix.java-version == '21' }}
```
niloc132, line 87: I should drop these back down to 17 in my branch, but yours should keep 21

dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
```diff
@@ -537,13 +537,25 @@ public void endVisit(BreakStatement x, BlockScope scope) {
 
     @Override
     public void endVisit(CaseStatement x, BlockScope scope) {
+      if (x.isExpr) {
+        InternalCompilerException excption =
+            new InternalCompilerException("Switch expressions not yet supported");
+        excption.addNode(new JCaseStatement(makeSourceInfo(x), null));
+        throw excption;
```
niloc132, line 544:
```suggestion
        InternalCompilerException exception =
            new InternalCompilerException("Switch expressions not yet supported");
        exception.addNode(new JCaseStatement(makeSourceInfo(x), null));
        throw exception;
```

niloc132: Should we also have a temporary exception in the visit/endVisit for classes to make sure they aren't records, and a test to confirm?

dev/core/test/com/google/gwt/dev/jjs/impl/Java17AstTest.java
```diff
@@ -0,0 +1,126 @@
+/*
+ * Copyright 2019 Google Inc.
```
niloc132, line 2: 2023/2024

dev/core/test/com/google/gwt/dev/jjs/impl/Java17AstTest.java
```diff
@@ -0,0 +1,126 @@
...
+import com.google.gwt.dev.jjs.ast.JProgram;
+import com.google.gwt.dev.jjs.ast.JStringLiteral;
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * Tests that {@link GwtAstBuilder} correctly builds the AST for
+ * features introduced in Java 17.
+ */
+public class Java17AstTest extends FullCompileTestBase {
```
niloc132, line 29: Looks like we need one more test class to be complete - the Java17AstTest runs parts of the compiler and makes sure that java can be translated to "gwt java", but we also need a class like Java11Test (and supersource...) to verify that it actually gets compiled to js and works as expected.

dev/core/test/com/google/gwt/dev/jjs/impl/Java17AstTest.java
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  public void testSwitchExpressionsNotSupported() {
+    try {
+      compileSnippet("void", "var month = Months.JUNE;" +
+          "var result = switch(month) {\n" +
+          "    case JANUARY, JUNE, JULY -> 3;\n" +
+          "    case FEBRUARY, SEPTEMBER, OCTOBER, NOVEMBER, DECEMBER -> 1;\n" +
+          "    case MARCH, MAY, APRIL, AUGUST -> 2;\n" +
+          "    default -> 0;" +
```
niloc132, line 110: aside: what's the pattern here? J means 3, M and A means 2, everything else is 1?

vegegoku: It is just a random pattern for testing, didnt need to mean anything.

dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
vegegoku: If we are throwing exception on visit, do we still need to do the same for endVisit?

niloc132 (commented):
I didn't think this through - but I think you're going to need to re-introduce the isJava8 flag as isJava11 so that the Java 17 test won't try to be compiled with Java 11. LMK if you want a hand, this Ant wiring is a pain.

niloc132:
Cherry-pick that last commit to its own PR? Looks like it will need to have at least one line wrapped too, to fix stylecheck.

dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
```diff
@@ -2752,6 +2764,12 @@ protected void pushBinaryOp(BinaryExpression x, JBinaryOperator op) {
     }
 
     protected boolean visit(TypeDeclaration x) {
+      if (x.isRecord()) {
+        InternalCompilerException excption =
+            new InternalCompilerException("Records not yet supported");
+        excption.addNode(new JClassType(makeSourceInfo(x), null, false, false));
```
niloc132, line 2770:
```suggestion
        InternalCompilerException exception =
            new InternalCompilerException("Records not yet supported");
        exception.addNode(new JClassType(makeSourceInfo(x), intern(x.name), false, false));
```
Prevents a NPE later in throwing this exception (and fixes a typo).

dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
niloc132: No, if a visitor throws, it will stop visiting.

I think we need more testing of switch exprs to fail, this doesn't cause an error (since it doesn't have a `case`):
```
    int i = switch(0) {
      default:
        yield 0;
    };
```

Actually I can't get it to fail even with a case:
```
    int i = switch(0) {
      case 1:
        yield 2;
      default:
        yield 7;
    };
```
Maybe we also need a check on the switchstatement? Yeah, confirmed, this only fails using `case ... -> ...` or `default -> ...`.

user/test-super/com/google/gwt/dev/jjs/super/com/google/gwt/dev/jjs/test/Java17Test.java
```diff
@@ -0,0 +1,80 @@
...
+ * Tests Java 17 features. It is super sourced so that gwt can be compiles under Java 11.
+ *
+ * IMPORTANT: For each test here there must exist the corresponding method in the non super sourced
+ * version.
+ *
+ * Eventually this test will graduate and not be super sourced.
+ */
+@GwtScriptOnly
+public class Java17Test extends GWTTestCase {
+
```
niloc132, line 34:
```suggestion
```

user/test/com/google/gwt/dev/jjs/Java17Test.java
```diff
@@ -0,0 +1,56 @@
...
+import com.google.gwt.junit.Platform;
+import com.google.gwt.junit.client.GWTTestCase;
+
+/**
+ * Dummy test case. Java17Test is super sourced so that GWT can be compiled by Java 8.
+ *
+ * NOTE: Make sure this class has the same test methods of its supersourced variant.
+ */
+@DoNotRunWith(Platform.Devel)
+public class Java17Test extends GWTTestCase {
```
niloc132, line 30: I think this is in the wrong package? I think it should be in com/google/gwt/dev/jjs/test instead, that would match the supersource vers

dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
niloc132: Still need a test, and failing code for something like this:
```java
    int i = switch(1) {
      case 1:
        yield 2;
      default:
        yield 7;
    };
```

I think you need a visit/endVisit for SwitchExpression to catch this - the error thrown above is only good for the new `case Foo ->` syntax.

.github/workflows/full-check.yml
```diff
@@ -29,7 +30,8 @@ jobs:
       - name: Checkout GWT tools into a sibling directory
         uses: actions/checkout@v4
         with:
-          repository: 'gwtproject/tools'
+          repository: 'vegegoku/tools'
+          ref: 'java-17-support'
```
niloc132, line 34: With the tools patch merged, revert this (in both yml files)

.github/workflows/full-check.yml
```diff
@@ -1,4 +1,4 @@
-# Run all tests and builds all aspects of GWT using Java 11 and 17. Runs
+# Run all tests and builds all aspects of GWT using Java 8, 11, and 17. Runs
```
niloc132, line 1:
```suggestion
# Run all tests and builds all aspects of GWT using Java 11, 17, and 21. Runs
```

dev/build.xml
```diff
@@ -2,7 +2,7 @@
   <property name="gwt.root" location=".."/>
   <property name="project.tail" value="dev"/>
   <property name="test.args" value="-ea"/>
-  <property name="test.jvmargs" value="-ea"/>
+  <property name="test.jvmargs" value="-ea -agentlib:jdwp=transport=dt_socket,server=y,suspend=y,address=8005"/>
```
niloc132, line 5: revert

niloc132 (commented):
Looks good, just some last minute cleanup. After https://github.com/gwtproject/gwt/pull/9905 is merged let's do one more CI run (to be sure github actions is happy to run on 11/17/21), then merge?

PR #9888, opened 2024-01-01
Remove unexpected sources from gwt-user by making valid servlet sources

### LINKED ISSUE #9886: RequestFactory client projects fail to compile using gwt-user.jar without jakarta
**GWT version:** HEAD-SNAPSHOT (discovered testing for 2.11.0 release)
**Browser (with version):** n/a
**Operating System:** any 

---

##### Description
Compiling a project using RequestFactory with gwt-user.jar and without requestfactory-server-jakarta.jar will cause a compile-time error. The source file (but not the .class file) for the jakarta variant of Logging and LoggingRequest are included in gwt-user, as the gwt-user sources are used for producing some other source jars.
https://github.com/gwtproject/gwt/blob/42c1e8a2057380d9cb67d86c2ca281b1e9f6b194/maven/lib-gwt.sh#L118-L126

Instead, those likely need to be filtered out of the actual gwt-user.jar, even if they end up in sources jars.

To recap the jakarta.servlet discussion, since RF requires that the bean/service proxies reference the server implementations (or, use `@ProxyFor` and `@ServiceFor`, but gwt's provided classes don't do that), and since the `com.google.web.bindery.requestfactory.server.Logging` service implementation uses `javax.servlet` types, there must exist a jakarta implementation of that service, and so there must exist a jakarta proxy for the service.  

##### Steps to reproduce
Grab a nightly sdk, unpack the DynaTableRf project, and attempt to compile it.

Expected, compiles without errors
Actual:
```
[INFO] --- gwt-maven-plugin:1.0.1:compile (default) @ DynaTableRf ---
[INFO] Compiling module com.google.gwt.sample.dynatablerf.DynaTableRf
[INFO]    Tracing compile failure path for type 'com.google.web.bindery.requestfactory.shared.jakarta.LoggingRequest'
[INFO]       [ERROR] Errors in 'jar:file:/home/colin/.m2/repository/org/gwtproject/gwt-user/HEAD-SNAPSHOT/gwt-user-HEAD-SNAPSHOT.jar!/com/google/web/bindery/requestfactory/shared/jakarta/LoggingRequest.java'
[INFO]          [ERROR] Line 27: Logging cannot be resolved to a type
[INFO]          [ERROR] Line 19: The import com.google.web.bindery.requestfactory.server.jakarta.Logging cannot be resolved
[INFO]    [ERROR] Aborting compile due to errors in some input files
```

##### Workaround
Disable `-strict`/`-failOnError`, or add the -jakarta rf jar, even if unused.

niloc132:
Removes an old hack which uses gwt-user's sources as if they were the correct sources for gwt-servlet by creating a real build artifact that contains the sources which correspond to the classes found in gwt-servlet and gwt-servlet-jakarta. This lets us avoid using gwt-user's own sources as if they were the sources for other jars, which in turn allows us to only package gwt-user sources in gwt-user.jar.

Fixes #9886

### DIFF
--- a/distro-source/build.xml
+++ b/distro-source/build.xml
@@ -31,10 +31,8 @@
       <!-- jars -->
       <zipfileset file="${gwt.build.lib}/gwt-dev.jar" prefix="${project.distname}" />
       <zipfileset file="${gwt.build.lib}/gwt-user.jar" prefix="${project.distname}" />
-      <zipfileset file="${gwt.build.lib}/gwt-servlet.jar" prefix="${project.distname}" />
-      <zipfileset file="${gwt.build.lib}/gwt-servlet-jakarta.jar" prefix="${project.distname}" />
+      <zipfileset file="${gwt.build.lib}/gwt-servlet*.jar" prefix="${project.distname}" />
       <zipfileset file="${gwt.build.lib}/gwt-codeserver.jar" prefix="${project.distname}" />
-      <zipfileset file="${gwt.build.lib}/gwt-servlet-deps.jar" prefix="${project.distname}" />
       <zipfileset file="${gwt.build.lib}/gwt-api-checker.jar" prefix="${project.distname}" />
       <zipfileset file="${gwt.build.lib}/requestfactory*.jar" prefix="${project.distname}" />
 
--- a/servlet/build.xml
+++ b/servlet/build.xml
@@ -13,7 +13,7 @@
       location="${gwt.build.lib}/gwt-${ant.project.name}-deps.jar" />
 
   <target name="build" description="Packages this project into a jar"
-      depends="-servlet, -servlet-jakarta, -deps" />
+      depends="-servlet, -servlet-jakarta, -servlet-sources, -servlet-jakarta-sources, -deps" />
 
   <target name="clean" description="Cleans this project's intermediate and output files">
     <delete file="${project.lib}" />
@@ -30,7 +30,6 @@
       <zipfileset src="${gwt.tools.lib}/streamhtmlparser/streamhtmlparser-jsilver-r10/streamhtmlparser-jsilver-r10-1.5-rebased.jar" />
       <zipfileset src="${gwt.tools.lib}/protobuf/protobuf-2.5.0/protobuf-java-rebased-2.5.0.jar" />
       <fileset dir="${gwt.dev.bin}">
-        <include name="com/google/gwt/dev/asm/**" />
         <include name="com/google/gwt/dev/util/Name*.class" />
         <include name="com/google/gwt/dev/util/StringKey.class" />
         <include name="com/google/gwt/util/tools/shared/**" />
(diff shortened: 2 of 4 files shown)


distro-source/build.xml
```diff
@@ -31,10 +31,8 @@
       <!-- jars -->
       <zipfileset file="${gwt.build.lib}/gwt-dev.jar" prefix="${project.distname}" />
       <zipfileset file="${gwt.build.lib}/gwt-user.jar" prefix="${project.distname}" />
-      <zipfileset file="${gwt.build.lib}/gwt-servlet.jar" prefix="${project.distname}" />
```
niloc132, line 34: Note to reviewers: added this wildcard to handle the existing 3 jars plus the two new -sources jars, in the same way that we wildcard requestfactory jars below.

servlet/build.xml
```diff
@@ -30,7 +30,6 @@
       <zipfileset src="${gwt.tools.lib}/streamhtmlparser/streamhtmlparser-jsilver-r10/streamhtmlparser-jsilver-r10-1.5-rebased.jar" />
       <zipfileset src="${gwt.tools.lib}/protobuf/protobuf-2.5.0/protobuf-java-rebased-2.5.0.jar" />
       <fileset dir="${gwt.dev.bin}">
-        <include name="com/google/gwt/dev/asm/**" />
```
niloc132, line 33: Note to reviewers: this package, along with those below (including the existing of gwtar files...) no longer exist in gwt.

servlet/build.xml
```diff
@@ -43,13 +42,28 @@
         <exclude name="com/google/gwt/json/**" />
         <exclude name="com/google/gwt/junit/*" />
         <exclude name="com/google/gwt/junit/client/GWTTestCase.*" />
-        <exclude name="com/google/gwt/junit/remote/**" />
         <exclude name="com/google/gwt/junit/server/**" />
-        <exclude name="com/google/gwt/benchmarks/*" />
-        <exclude name="**/*.gwtar" />
       </fileset>
-
-      <!-- additional dependencies (used by scripts). -->
```
niloc132, line 52: This was a placeholder added in https://github.com/gwtproject/gwt/commit/498861e7f00d4b89f23d7bc0900f295b46a160e9, but was never used (at least in open source).

niloc132:
Passing nightly builds: https://github.com/niloc132/gwt/actions/runs/7374819171

jnehlmeier:
Seems to look good. Just because it came into my mind while looking that this PR: Do you have any idea why most of gwt-user is also in gwt-servlet?

niloc132:
> Seems to look good. Just because it came into my mind while looking that this PR: Do you have any idea why most of gwt-user is also in gwt-servlet?

I don't know, no. It sure seems to me like it should only be the server-side and shared RPC classes, but this seems to be a historical decision that predates the import to SVN, the initial commit in the current git repo. Reviewing more recent commits, later changes took at least some care to avoid putting classes into gwt-servlet, but not uniformly. It seems to suggest that there was a purpose in client-only code being available _as bytecode_ to the server... My best guesses are for Class serialization (not supported now, if ever), for stack trace deobfuscation (doesn't require this), or for some kind of UI-builder? It doesn't seem to make sense that this is needed, but the bazel/blaze wiring seems to say that these are required
https://github.com/gwtproject/gwt/blob/d78f9e9c843af460218189d9839eea4a7a85a0e7/user/BUILD#L305-L327

I'm skeptical of this - although the last line in that block includes SuggestOracle, which does make sense to me to include. Maybe they have an internal rule of some kind to avoid `NoClassDefFoundError`s by statically ensuring that all transitive classes are present on a classpath, reachable or not? I'm not convinced that this makes sense. At least with this in mind, it seems deliberate to Google to have left it this way, at least as of 2014 when that file was added, and at least as late as 2018 when that list was still being edited.

gwt-servlet is also unusual (as this patch discusses) in that it doesn't contain its own sources - but both gwt-user and gwt-dev do. This patch doesn't change that, but just introduces a separate jar following the same steps, rather than rock the boat too much and change the server-side classpath.

gwt-servlet is also different from gwt-user and gwt-dev in that gwt-user and gwt-dev must match their version (there are certain string references in gwt-dev that will fail to compile correctly if the right gwt-user version isn't on the classpath), but there is no dependency that gwt-servlet matches the running client code. In fact, gwt-rpc explicitly handles a version mismatch (just requires that the serialized types and services remain consistent, and unlike the user/dev mismatch, that is actually validated at runtime).

The requestfactory jars handle this problem differently - from a set of roots, bytecode is examined to discover references to collect other required GWT types. Through a path that I've now forgotten, this ends up meaning that the server-only jars must include some apparently client-only types including the Scheduler, JsArray, JavaScriptObject, etc. 

Over the years I've had a hobby project to attempt to splinter the compiler into actual separate modules and break up dependencies, so that one day we could migrate cleanly to maven/gradle to build GWT, and distribute an actual DAG of jars rather than duplicate so many classes across so many jars. I honestly don't understand how it started off this way to begin with, and how none of the big refactors (including partially introducing bazel/blaze) managed to touch this at all. Even if successful, this project would likely want to continue to produce our legacy gwt-user/gwt-servlet/etc jars, duplicates and all, to ensure that existing use cases (like building from ant) continue to function, at least for a time.

niloc132:
I've tested this by updating DynaTableRf and rewriting it to work with Jetty 11, and it all works correctly, with the exception of needing to copy LoggingRequest into the project's own sources (as we knew would be necessary for RequestFactoryLogHandler), so it was a mistake to clone and modify that type. I don't think it is worth reverting, just documenting as part of release notes and jakarta.servlet instructions.

Issue #9844, opened 2023-07-15
Jakarta Validation Support


dodgex:
In addition to Jakarta Servlet Support discussed in https://github.com/gwtproject/gwt/issues/9727 and worked on in https://github.com/niloc132/gwt/pull/3 I think Jakarta Validation 3.0 should be supported too.

While the main part of validation is deprecated as stated on its [documentation](https://www.gwtproject.org/doc/latest/DevGuideValidation.html) there are also some validation related bits in requestfactory where we are especially interested in, as missing Jakarta Support prevents us from upgrading our GWT projects to Spring Boot 3

In the context of the above linked PR I already started some work on this. As of now the changes made so far can be found [here](https://github.com/dodgex/gwt/commits/9727-jakarta-servlet-support). These changes have not yet been tested.

@niloc132 [mentioned](https://github.com/niloc132/gwt/pull/3#issuecomment-1636885848) that there are already 3rd party projects that provide client side validation. He named https://gitlab.com/ManfredTremmel/gwt-bean-validators but with the note that it looks like it also only supports `javax.validation`, I found that they have a 2.0.x release with jakarta support ([related issue](https://gitlab.com/ManfredTremmel/gwt-bean-validators/-/issues/30) and [branch](https://gitlab.com/ManfredTremmel/gwt-bean-validators/-/tree/feature/validationapi3?ref_type=heads))

I have no idea how the deprecation and removal policy in GWT is. But as the code does not mark the API as `@Deprecated` I assume removing would not be an option. 

For now I will continue to test the changes I made so far and apply fixes where necessary. While my initial focuse here might be requestfactory, I am willing to also test the client side validation part too.

dodgex:
I have created a new branch https://github.com/dodgex/gwt/tree/9844-jakarta-validation-support for this. The branch is based on  the work done in PR https://github.com/gwtproject/gwt/pull/9845.

After updating the branch to the changes in that PR, i successfully tested the validation part used by requestfactory in a small test project.

For the test project, currently `de.knightsoft-net:gwt-time-dummy` in a 2.x version is required as it provides (dummy) emulation classes for `java.time.Clock` required by jakarta validation 3.0 stuff in `com.google.gwt.validation.client`.

niloc132:
Checking in for any update here - should we proceed with #9845, or do you think you have a proposal ready to discuss?

dodgex:
[...]

PR #9891, opened 2024-01-04
Add jaxb dependencies

FrankHossfeld:
This PR fixes an issue with the dynatablerf example. Starting with GWT 2.11.0 the jabx dependencies are no longer available and need to be added.

### DIFF
--- a/samples/dynatablerf/pom.xml
+++ b/samples/dynatablerf/pom.xml
@@ -94,6 +94,18 @@
       </exclusions>
     </dependency>
 
+    <!-- added to work with GWT 2.11.0 and newer --> 
+    <dependency>
+      <groupId>javax.xml.bind</groupId>
+      <artifactId>jaxb-api</artifactId>
+      <version>2.4.0-b180830.0359</version>
+    </dependency>
+    <dependency>
+      <groupId>com.sun.xml.bind</groupId>
+      <artifactId>jaxb-impl</artifactId>
+      <version>4.0.4</version>
+    </dependency>
+    
     <!-- Required by Hibernate validator because slf4j-log4j is
          optional in the hibernate-validator POM  -->
     <dependency>


samples/dynatablerf/pom.xml
```diff
@@ -94,6 +94,18 @@
       </exclusions>
     </dependency>
 
+    <!-- added to work with GWT 2.11.0 and newer -->
```
jnehlmeier, line 97: Doesn't this provide classes using jakarta package and some imports need to be updated as well somewhere?

FrankHossfeld: After adding the dependencies I were able to run the sample with Java 11 & 17. In case we switch to **jakarta.xml.bind**, we might have code changes.

jnehlmeier: Or to be more precise: jaxb-impl:4.x is Jakarta EE 10 (based on jakarta package) and I am wondering if it really matches jaxb-api:2.4.x which should be Jakarta EE 8 (using the javax package).

FrankHossfeld:
Tested again. The original implementation works well and without error with Java 8, but will fail in Java 11 & Java 17. I think in Java 9 the jaxb package have been removed from the JDK. Starting with Java 9 (I think) the dependencies must be added to the pom to avoid the error.

Checked the PR with Java 8. Works also.

Starting with GWT 2.12.0 - when Java 8 is dropped - we can switch to the `jakarta.xml.bind` package. This change will  have a bigger impact. I expect at least changes on the package names ...

niloc132:
I'm not sure this is the best solution, but it is _a_ solution - and changing to jakarta for xml doesn't make sense. The issue is that GWT's requestfactory implementation needs some javax.validation implementation (#9844 is not simple it appears, at least to maintain backwards compat, and we've already said that for validation implementations users should look outside of GWT), and the _example_ picks hibernate as the implementation. That particular hibernate version happens to require javax.xml.bind - see from the stack trace:
```
java.lang.ClassNotFoundException: javax.xml.bind.JAXBException
    at org.codehaus.plexus.classworlds.strategy.SelfFirstStrategy.loadClass (SelfFirstStrategy.java:50)
    at org.codehaus.plexus.classworlds.realm.ClassRealm.unsynchronizedLoadClass (ClassRealm.java:271)
    at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass (ClassRealm.java:247)
    at org.codehaus.plexus.classworlds.realm.ClassRealm.loadClass (ClassRealm.java:239)
    at org.eclipse.jetty.webapp.WebAppClassLoader.loadClass (WebAppClassLoader.java:511)
    at java.lang.ClassLoader.loadClass (ClassLoader.java:527)
    at org.hibernate.validator.engine.ConfigurationImpl.parseValidationXml (ConfigurationImpl.java:252)
    at org.hibernate.validator.engine.ConfigurationImpl.buildValidatorFactory (ConfigurationImpl.java:143)
    at javax.validation.Validation.buildDefaultValidatorFactory (Validation.java:111)
```

...It almost seems like the exclude lines, which keep jaxb-api _out_ of the build should be removed, but as far as I can tell, those excludes don't do anything. Looking at https://search.maven.org/artifact/org.hibernate/hibernate-validator/4.1.0.Final/jar, those dependencies are scope=provided anyway.

So I'm not sure there is a better fix here, aside from "rewrite Editors and RequestFactory to stop using javax.validation apis (just SimpleViolation I think?) to use a newer, external implementation of jakarta.validation instead"?

jnehlmeier:
GWTs documentation says that XML configuration is not supported and Hibernate Validator 4.1 documentation says that JAXB is only needed for XML configuration. If the code does not use XML configuration then you can disable it calling `Configuration.ignoreXmlConfiguration()` during `ValidationFactory` creation. If that method is called, it should stop trying to load JAXB.

Maybe we should change RequestFactory server code to probe for JAXB and if not available call `Configuration.ignoreXmlConfiguration()`?

niloc132:
> Maybe we should change RequestFactory server code to probe for JAXB and if not available call Configuration.ignoreXmlConfiguration()?

Finally looking at this (as part of #10232) - looks like the change is:
```
diff --git a/user/src/com/google/gwt/validation/rebind/BeanHelperCache.java b/user/src/com/google/gwt/validation/rebind/BeanHelperCache.java
index 1db6b9c4bb..269ae7d9f6 100644
--- a/user/src/com/google/gwt/validation/rebind/BeanHelperCache.java
+++ b/user/src/com/google/gwt/validation/rebind/BeanHelperCache.java
@@ -52,7 +52,11 @@ public class BeanHelperCache { // public for testing
    */
   public BeanHelperCache() {
     cache = new HashMap<JClassType, BeanHelper>();
-    serverSideValidator = Validation.buildDefaultValidatorFactory().getValidator();
+    serverSideValidator = Validation.byDefaultProvider()
+        .configure()
+        .ignoreXmlConfiguration()
+        .buildValidatorFactory()
+        .getValidator();
   }
```

This appears to work, I'll submit it separately.

niloc132:
Superseded by https://github.com/gwtproject/gwt/pull/10249

PR #9899, opened 2024-01-12
Update ErrorProne to run on Java 21, drop Java 8 support

### LINKED ISSUE #9859: Update ErrorProne to support building (and testing) GWT on Java 21
**GWT version:** 
**Browser (with version):** 
**Operating System:** 

---

##### Description

Do you have a tentative date for GWT support of the Java 21 LTS version?

The current supported version of GWT is https://www.gwtproject.org/release-notes.html#Release_Notes_2_10_0, it supports Java 17.

### LINKED ISSUE #9870: Increase minimum Java to build GWT or run devmode/compiler to 11
Where possible, we will leave Java 8 support in gwt-servlet and requestfactory-server, but any code that uses the JDT or is built (only) by the JDT can now use Java 11 syntax/features.

This enables updating the JDT to somewhere around 3.32 - past that, and we must also update to a minimum of 17.

niloc132:
This patch updates to the latest version of ErrorProne, which requires dropping support for Java 8 (as planned). This  allows us to remove several "is this Java 8" checks in the build.

As the new ErrorProne version required adding new suppressions, all existing project-wide suppressions have been removed and distributed to the methods throughout the project that actually need them.

We cannot yet add support for running CI on Java 21, that also requires updating asm - otherwise there are a handful of compiler internals tests that will fail when asm interacts with Java's own bytecode (e.g. java.lang.Object).

The javac arg `-XDstringConcat=inline` must be added to gwt-user tests when compiled to release=11, as https://bugs.openjdk.org/browse/JDK-8323657 breaks a regression test for JDT, and the class fails to load for the ExtraCompilerSuite.

Fixes #9870 
Fixes #9859

### DIFF
--- a/user/build.xml
+++ b/user/build.xml
@@ -2,7 +2,7 @@
   <property name="gwt.root" location=".."/>
   <property name="project.tail" value="user"/>
   <property name="test.args" value="-ea -sourceLevel auto"/>
-  <property name="test.jvmargs" value="-ea -noverify"/>
+  <property name="test.jvmargs" value="-ea"/>
 
   <!-- support old variables names -->
   <condition property="gwt.hosts.web.selenium" value="${gwt.selenium.hosts}">
--- a/user/build.xml
+++ b/user/build.xml
@@ -2,7 +2,11 @@
   <property name="gwt.root" location=".."/>
   <property name="project.tail" value="user"/>
   <property name="test.args" value="-ea -sourceLevel auto"/>
-  <property name="test.jvmargs" value="-ea"/>
+  <!--
+  Adding "-noverify" is a workaround for https://bugs.openjdk.org/browse/JDK-8323657
+  so that tests can be compiled under Java9+.
+  -->
+  <property name="test.jvmargs" value="-ea -noverify"/>
 
   <!-- support old variables names -->
   <condition property="gwt.hosts.web.selenium" value="${gwt.selenium.hosts}">
--- a/user/build.xml
+++ b/user/build.xml
@@ -14,27 +14,16 @@
   <property name="test.timeout" value="5"/>
   <property name="emma.merged.out" value="${junit.out}/emma-coverage"/>
 
-  <property name="gwt.junit.testcase.web.includes" value="${gwt.junit.testcase.includes}"/>
-  <condition property="gwt.junit.testcase.web.excludes"
-            value="**/*JreSuite.class,**/OptimizedOnly*"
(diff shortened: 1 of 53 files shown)


niloc132:
This PR requires https://github.com/gwtproject/tools/pull/29 to merge first (then a temporary commit to change the tools repo can be rolled back).

Nightly build: https://github.com/niloc132/gwt/actions/runs/7505713642

user/build.xml
```diff
@@ -295,33 +271,11 @@
...
-                 -Xep:LoopConditionChecker:OFF
-                 -Xep:JUnitAssertSameCheck:OFF
-                 -Xep:CollectionIncompatibleType:OFF
-                 -Xep:DeadThread:OFF
-                 -Xep:ComplexBooleanConstant:OFF
-                 -Xep:ComparableType:OFF
-                 -Xep:DoNotCall:OFF
-                 -Xep:BadAnnotationImplementation:OFF
-               ">
+               release="11">
```
niloc132, line 278: Consider adding `-XDstringConcat=inline` here instead of `-noverify` in a few different places, since that generates correct bytecode until either the fix is backported to 11, or we set our minimum java level to 23+. See https://github.com/openjdk/jdk/pull/17483

niloc132:
Made a few more changes to restrict the scope of the JDK bug workaround, and to fix CI details suggesting that we were able to build on Java 21 (not yet, but soon).

Nightlies running at https://github.com/niloc132/gwt/actions/runs/7642560568

niloc132:
If there are no objections in the next 24ish hours, I'll go ahead and merge the tools change, then roll back the GHA change to use my tools fork, and then land this branch, so that we can continue with work in https://github.com/gwtproject/gwt/issues/9869.

FrankHossfeld (approved):
There are a lot of @SuppressWarnings. Are they just there to avoid warnings during the build or are there any other reasons?
Thanks for make it clear ..

niloc132:
Each  `@SuppressWarnings("$warning")` is to replace the removed `-Xep:$warning:OFF` , so that rather than allowing all instances of some specific warning in the entire source directory, we highlight "this method is ignoring a specific warning". That way, we can be more careful as new warnings are added to the codebase.

Most of the existing warnings are suppressed for a good reason, like wanting the compiler to test apparently silly things (since the code might not compile correctly). Some outside of test code probably could be cleaned up (BoxedPrimitiveEquality in Checkbox cell perhaps, DoubleBraceInitialization in BeanParser?). I deliberately only transformed the existing compiler flags to SuppressWarnings annotations at this time though.

FrankHossfeld:
Thanks for clearing this ... sounds logical ..

tbroyer (approved):

PR #9903, opened 2024-01-19
Add .editorconfig

hjhgitw:
Configuration file format for defining coding styles in shared projects

addressed in this pull request, e.g. "Fixes #42" to close the linked
issue, or "Partial #42" to indicate that this merge will address it,
but that it isn't ready to be closed yet.


hjhgitw:
i know :) 
i did think same thing 1 year ago, but it did correct on many projects the style and even wrong file encodings that we didn't notice.. (without it)

see : https://editorconfig.org/ its an old mature thing.
or : https://www.freecodecamp.org/news/how-to-use-editorconfig-to-standardize-code-styles/

you did switch to JAVA 21 and i guess we should even Manage Style Rules

or only the mandatory part

root = true

[*]
charset = utf-8
end_of_line = lf
indent_size = 2
indent_style = space
insert_final_newline = false
max_line_length = 140
tab_width = 2


files like https://github.com/hjhgitw/gwt-material/blob/master/CONTRIBUTING.md
note : Code Style & Indentation, would than be senseless and clear forced

my 2 cents but decide (it's a small file with big impact)


Conclusion
If you're a maintainer, either working in a collaborative environment or in open-source, I strongly recommend adding an .editorconfig file defining the project's conventions to the root of your repository.

Maintainers can then spend more time reviewing clean pull requests that adhere to the style guide, as the editor will automatically start enforcing or applying the conventions.

Committers get a better experience, as the project will override their workspace settings. This reduces the need to reformat code or work against preconfigured editor settings.

And projects will be cleaner, as the conventions will be in a single vendor-agnostic file, rather than editor-specific files that only certain contributors can use anyway.

.editorconfig
```diff
@@ -0,0 +1,23 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = false
+max_line_length = 140
```
niloc132, line 9: Our stylecheck sets this to 100 (though only as a warning, not an error).
```suggestion
max_line_length = 100
```

.editorconfig
```diff
@@ -0,0 +1,23 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = false
```
niloc132, line 8: I believe this should generally be true, that all files should end on a newline? Inherited from c, also a convention to get reduced noise in git diffs, and I know better for interacting with files from shell too...
```suggestion
insert_final_newline = true
```

.editorconfig
```diff
@@ -0,0 +1,23 @@
...
+max_line_length = 140
+tab_width = 2
+ij_continuation_indent_size = 2
+
+[*.java]
+ij_java_class_count_to_use_import_on_demand = 1000 # Do not group imports with an asterisk
+ij_java_names_count_to_use_import_on_demand = 1000 # Do not group imports with an asterisk
+trim_trailing_whitespace = true
```
niloc132, line 16: I can't think of a reason this should be specific to java.

niloc132 (commented):
I'm not opposed to it if we're going to get general value out of it - can other contributors weigh in? I would generally prefer not to put IDE-specific configuration in the repo (though a proper formatter would be lovely, but something I wasn't intended to put time into until we drop ant itself). This config seems to mostly be `ij_` prefixed, which appears to mean IntelliJ rather than something that eclipse/vscode/github/vim will recognize - do we really want those?

FrankHossfeld:
Having an automated code style, that formats the code at commit time is something many repros are looking for a solution.

In my opinion, using a code formatter of an IDE is not a great idea. First, we need to support at least the most used IDEs - IntelliJ, Eclipse and maybe VsCode. I made the experience that with every release of IntelliJ the code formatter changes and my code style settings create different results. 

So we tried the google-java-format plugin (available for Eclipse and IntelliJ). But formatting must be manual triggered. Next we created a hook, that runs the formatter at commit time. But then someone has to commit using the terminal. 

At the end, none of the solutions were really satisfying.

In my opinion IDE settings or runtime configurations are no files I would push in a repro.

.editorconfig
```diff
@@ -3,21 +3,8 @@ root = true
 [*]
 charset = utf-8
 end_of_line = lf
-indent_size = 2
+indent_size = 4
```
niloc132, line 6: This should definitely be 2, so that String doesn't get reformatted. 

Does tab_width need to be set if tabs arent used?

Probably try editing a bunch of files and make sure your configuration in this file doesn't try to rewrite other files in the repo.

hjhgitw: Ok THX

PR #9905, opened 2024-01-21
ProxyAutoBean should be able to handle List<Integer> in Java 21

### LINKED ISSUE #9902: RequestFactory tests fails under java 21 due to changes in the List API
**GWT version: 2.11** 
**Browser (with version):** 
**Operating System:Ubuntu** 

---

##### Description

While building GWT using java 21 `RequestFactorySuite` tests failed with 58 errors

```
Testcase: testFetchEntityWithRelation took 0.007 sec
	Caused an ERROR
Exception caught: Cannot invoke "com.google.web.bindery.autobean.vm.impl.ProxyAutoBean.traverse(com.google.web.bindery.autobean.shared.AutoBeanVisitor, com.google.web.bindery.autobean.shared.impl.AbstractAutoBean$OneShotContext)" because "bean" is null
com.google.web.bindery.event.shared.UmbrellaException: Exception caught: Cannot invoke "com.google.web.bindery.autobean.vm.impl.ProxyAutoBean.traverse(com.google.web.bindery.autobean.shared.AutoBeanVisitor, com.google.web.bindery.autobean.shared.impl.AbstractAutoBean$OneShotContext)" because "bean" is null
	at com.google.web.bindery.requestfactory.shared.impl.AbstractRequestContext$StandardPayloadDialect.processPayload(AbstractRequestContext.java:413)
	at com.google.web.bindery.requestfactory.shared.impl.AbstractRequestContext$5.onTransportSuccess(AbstractRequestContext.java:1160)
	at com.google.web.bindery.requestfactory.server.testing.InProcessRequestTransport.send(InProcessRequestTransport.java:64)
	at com.google.web.bindery.requestfactory.shared.impl.AbstractRequestContext.doFire(AbstractRequestContext.java:1154)
	at com.google.web.bindery.requestfactory.shared.impl.AbstractRequestContext.fire(AbstractRequestContext.java:595)
	at com.google.web.bindery.requestfactory.shared.impl.AbstractRequest.fire(AbstractRequest.java:56)
	at com.google.web.bindery.requestfactory.shared.impl.AbstractRequest.fire(AbstractRequest.java:61)
	at com.google.web.bindery.requestfactory.gwt.client.RequestFactoryTest.testFetchEntityWithRelation(RequestFactoryTest.java:856)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at com.google.gwt.junit.client.GWTTestCase.runTest(GWTTestCase.java:419)
	at com.google.gwt.junit.client.GWTTestCase.run(GWTTestCase.java:247)
Caused by: java.lang.NullPointerException: Cannot invoke "com.google.web.bindery.autobean.vm.impl.ProxyAutoBean.traverse(com.google.web.bindery.autobean.shared.AutoBeanVisitor, com.google.web.bindery.autobean.shared.impl.AbstractAutoBean$OneShotContext)" because "bean" is null
	at com.google.web.bindery.autobean.vm.impl.ProxyAutoBean.traverseProperties(ProxyAutoBean.java:382)
	at com.google.web.bindery.autobean.shared.impl.AbstractAutoBean.traverse(AbstractAutoBean.java:166)
	at com.google.web.bindery.autobean.vm.impl.ProxyAutoBean.traverseProperties(ProxyAutoBean.java:358)
	at com.google.web.bindery.autobean.shared.impl.AbstractAutoBean.traverse(AbstractAutoBean.java:166)
	at com.google.web.bindery.autobean.shared.impl.AbstractAutoBean.accept(AbstractAutoBean.java:101)
	at com.google.web.bindery.requestfactory.shared.impl.ProxySerializerImpl.serialize(ProxySerializerImpl.java:118)
	at com.google.web.bindery.requestfactory.gwt.client.RequestFactoryTestBase.checkSerialization(RequestFactoryTestBase.java:181)
	at com.google.web.bindery.requestfactory.gwt.client.RequestFactoryTest$23.onSuccess(RequestFactoryTest.java:860)
	at com.google.web.bindery.requestfactory.gwt.client.RequestFactoryTest$23.onSuccess(RequestFactoryTest.java:857)
	at com.google.web.bindery.requestfactory.shared.impl.AbstractRequest.onSuccess(AbstractRequest.java:131)
	at com.google.web.bindery.requestfactory.shared.impl.AbstractRequestContext$StandardPayloadDialect.processPayload(AbstractRequestContext.java:380)
	
```
tracing the issue alongside @niloc132 we found the reason of the failure is that in java 21 a new methods are added to the list class `getFirst`, `getLast`, `isEmpty`, and in the ProxyAutoBean when tries to build the propertData for a list field is collect those methods as getters, this result in the field to be recognized as a nested entity that need to be traversed for its internal properties and is being processed as a reference instead of a value causing the above exception. 

full logs are attached 
[logs.txt](https://github.com/gwtproject/gwt/files/13981387/logs.txt)

##### Steps to reproduce

cd into gwt user folder and run the command 

`ant test.nongwt -Dgwt.nongwt.testcase.includes='**/RequestFactoryJreSuite.class'`

##### Known workarounds

adding a simple check to skip Collection classes from the propertyData will make this pass.

```
private static Map<String, Data> calculateData(Class<?> beanType) {
    Map<String, Data> toReturn;
    synchronized (cache) {
      toReturn = cache.get(beanType);
      if (toReturn == null) {
        Map<String, Data> getters = new HashMap<String, Data>();
        List<Method> setters = new ArrayList<Method>();
        if (!Collection.class.isAssignableFrom(beanType)) { // <------ here
          for (Method method : beanType.getMethods()) {
            if (BeanMethod.GET.matches(method)) {
              // match methods on their name for now, to find the most specific
              // override
              String name = method.getName();

            Type genericReturnType = TypeUtils.resolveGenerics(beanType, method.getGenericReturnType());
              Class<?> returnType = TypeUtils.ensureBaseType(genericReturnType);

              Data data = getters.get(name);
              
...
```

vegegoku:
Fix request factory tests fails with errors under java 21 due to added apis to the List class `getLast`, `getFirst`, `isEmpty` which are causing ProxyAutoBean treats the list as a traversable object.

fix gwtproject/gwt#9902

### DIFF
--- a/user/src/com/google/web/bindery/autobean/vm/impl/ProxyAutoBean.java
+++ b/user/src/com/google/web/bindery/autobean/vm/impl/ProxyAutoBean.java
@@ -112,34 +112,38 @@ public class ProxyAutoBean<T> extends AbstractAutoBean<T> {
       if (toReturn == null) {
         Map<String, Data> getters = new HashMap<String, Data>();
         List<Method> setters = new ArrayList<Method>();
-        for (Method method : beanType.getMethods()) {
-          if (BeanMethod.GET.matches(method)) {
-            // match methods on their name for now, to find the most specific
-            // override
-            String name = method.getName();
+        if (!(Collection.class.isAssignableFrom(beanType)
+            || Map.class.isAssignableFrom(beanType))) {
+          for (Method method : beanType.getMethods()) {
+            if (BeanMethod.GET.matches(method)) {
+              // match methods on their name for now, to find the most specific
+              // override
+              String name = method.getName();
 
-            Type genericReturnType = TypeUtils.resolveGenerics(beanType, method.getGenericReturnType());
-            Class<?> returnType = TypeUtils.ensureBaseType(genericReturnType);
+              Type genericReturnType = TypeUtils.resolveGenerics(beanType,
+                  method.getGenericReturnType());
+              Class<?> returnType = TypeUtils.ensureBaseType(genericReturnType);
 
-            Data data = getters.get(name);
-            if (data == null || data.type.isAssignableFrom(returnType)) {
-              // no getter seen yet for the property, or a less specific one
-              PropertyType propertyType;
-              if (TypeUtils.isValueType(returnType)) {
-                propertyType = PropertyType.VALUE;
-              } else if (Collection.class.isAssignableFrom(returnType)) {
-                propertyType = PropertyType.COLLECTION;
-              } else if (Map.class.isAssignableFrom(returnType)) {
-                propertyType = PropertyType.MAP;
-              } else {
-                propertyType = PropertyType.REFERENCE;
+              Data data = getters.get(name);
+              if (data == null || data.type.isAssignableFrom(returnType)) {
+                // no getter seen yet for the property, or a less specific one
+                PropertyType propertyType;
+                if (TypeUtils.isValueType(returnType)) {
+                  propertyType = PropertyType.VALUE;
+                } else if (Collection.class.isAssignableFrom(returnType)) {
+                  propertyType = PropertyType.COLLECTION;
+                } else if (Map.class.isAssignableFrom(returnType)) {
+                  propertyType = PropertyType.MAP;
+                } else {
+                  propertyType = PropertyType.REFERENCE;
+                }
+                data = new Data(method, genericReturnType, returnType, propertyType);
+
+                getters.put(name, data);
               }
-              data = new Data(method, genericReturnType, returnType, propertyType);
-
-              getters.put(name, data);
+            } else if (BeanMethod.SET.matches(method) || BeanMethod.SET_BUILDER.matches(method)) {
+              setters.add(method);
             }
-          } else if (BeanMethod.SET.matches(method) || BeanMethod.SET_BUILDER.matches(method)) {
-            setters.add(method);
           }
         }
 


user/src/com/google/web/bindery/autobean/vm/impl/ProxyAutoBean.java
```diff
@@ -112,34 +112,36 @@ private static Map<String, Data> calculateData(Class<?> beanType) {
       if (toReturn == null) {
         Map<String, Data> getters = new HashMap<String, Data>();
         List<Method> setters = new ArrayList<Method>();
-        for (Method method : beanType.getMethods()) {
-          if (BeanMethod.GET.matches(method)) {
-            // match methods on their name for now, to find the most specific
-            // override
-            String name = method.getName();
+        if (!Collection.class.isAssignableFrom(beanType)) {
```
niloc132, line 115: To future-proof this, should we also test for Map? 

Perhaps also TypeUtils.isValueType() to decide "its a value, never need to check it for properties", but I can't really tell to begin with why this method is being called on non-beans to begin with... It seems like the checks done below should be done at a much higher level to prevent this from even happening to begin with...
...that said, it might also make sense to be as conservative as possible, and avoid rewriting and breaking things not explicitly tested already. But it does seem like `List<Date>` should break in the same way `List<Integer>` broke here, since Dates are values and not beans (like Integers).

niloc132: After our call, I did some testing/debugging, and determined a few things:
 * While it is true that this `calculateData` method has historically collected `Collection.isEmpty`/`Map.isEmpty`, it never did anything with those - once assigned to the `ProxyAutoBean.propertyData` field, nothing ever consumed it.
 * This method also apparently was called on `Iterator`, and collected `Iterator.hasNext` as a boolean field - also incorrect/nonsense, since Iterator can't be a collection/bean/value.

From my perspective, this calculateData methods is written "backwards" - it probably should do the `VALUE`/`REFERENCE`/`COLLECTION`/`MAP` check up front before treating the object like a bean to begin with. The only reason I can think of right now for why this isn't done this way is that a bean could have a generic property T, and a subtype might make T into a value/reference/collection/map, so the specific getter/setter must be visited to even decide what T will mean - but even there, the return value of the getter could be read (MyBean/List/etc) if that was the problem? I'm not confident enough in this to propose a full refactor here (and this only applies to the "Proxy" implementation anyway... IMHO this should all be annotation based as it is, though RF might be a bigger hurdle).

vegegoku: Maybe the safest and quickest solution to start a new exclusion registry and include only the classes/methods that causes the issue? If you think this is an acceptable approach for now I can make a PR for that. WDYT?

vegegoku: Also in regards to the generic part, I guess this is exactly what we have here, both `getFirst` and `getLast` return generic types in the class `java.util.List` and while collecting the propertyData it cant determine the actual type for those methods and falls-back to treat them as references, so a field like `private List<Integer> numbers` will process the methods as getters for a reference instead of a value and that is the main reason for the issue.

Now also if we can find how and why generics are processed like this without and since you said that even though they are processed nothing happens later later maybe that is also an area to check if it can be fixed but I think that is not as easy as it looks and might end up being a bigger change than what we think.

niloc132: No, Collections are made into json lists, and Maps are made into either a map (if the key is a string), or a list of pairs - it will never make sense for a List/Set to have a json list plus a keyed json value.

Additionally, all of these methods that we're discussing (hasNext/isEmpty/getFirst/getLast) are derrived from the real value of the underlying collection information - so we don't need to visit/serialize these properties ever - the remote end receiving the data can re-derive the values.

To validate this for yourself, check out AutoBeanCodexText - this is run both as a gwt test and a jre test when you run the full AutoBeanSuite. The `testSplittable` method doesn't quite go to JSON as far as the test is concerned (see SplittableTest etc for that, which validates that Splittables turn into JSON and back), but only ensures that AutoBean->Splittable->AutoBean work. Adding a few assertions or a breakpoint should maybe help you convince yourself that it never makes sense for these properties to do anything:
```
diff --git a/user/test/com/google/web/bindery/autobean/shared/AutoBeanCodexTest.java b/user/test/com/google/web/bindery/autobean/shared/AutoBeanCodexTest.java
index 8a0881f65b..dde11263a1 100644
--- a/user/test/com/google/web/bindery/autobean/shared/AutoBeanCodexTest.java
+++ b/user/test/com/google/web/bindery/autobean/shared/AutoBeanCodexTest.java
@@ -440,6 +442,8 @@ public class AutoBeanCodexTest extends GWTTestCase {
 
     List<Splittable> list = decoded.as().getSimpleList();
     assertEquals(3, list.size());
+    assertFalse(list.isEmpty());
+    assertTrue(list.iterator().hasNext());
     assertNull(list.get(1));
     assertEquals("Simple", AutoBeanCodex.decode(f, Simple.class, list.get(2)).as().getString());
   }
```
Better still, replace the checkEncode method (and _only_ run the one `testSplittable` method, so the output isn't to spammy):
```
diff --git a/user/test/com/google/web/bindery/autobean/shared/AutoBeanCodexTest.java b/user/test/com/google/web/bindery/autobean/shared/AutoBeanCodexTest.java
index 8a0881f65b..dde11263a1 100644
--- a/user/test/com/google/web/bindery/autobean/shared/AutoBeanCodexTest.java
+++ b/user/test/com/google/web/bindery/autobean/shared/AutoBeanCodexTest.java
@@ -450,8 +454,11 @@ public class AutoBeanCodexTest extends GWTTestCase {
   }
 
   private <T> AutoBean<T> checkEncode(AutoBean<T> bean) {
+    System.out.println(bean.getClass());
     Splittable split = AutoBeanCodex.encode(bean);
+    System.out.println(split.getPayload());
     AutoBean<T> decoded = AutoBeanCodex.decode(f, bean.getType(), split);
+    System.out.println(decoded.getClass());
     assertTrue(AutoBeanUtils.deepEquals(bean, decoded));
     return decoded;
   }
```
Java 17:
```
------------- Standard Output ---------------
class com.google.web.bindery.autobean.vm.impl.ProxyAutoBean
{"string":"Hello ['\"] world","simpleList":[{"string":"Simple"},null,{"string":"Simple"}],"splittableMap":[["12345"],["5678"]],"simple":{"string":"Simple"}}
class com.google.web.bindery.autobean.vm.impl.ProxyAutoBean
```
Java 21:
```
------------- Standard Output ---------------
class com.google.web.bindery.autobean.vm.impl.ProxyAutoBean
{"string":"Hello ['\"] world","simpleList":[{"string":"Simple"},null,{"string":"Simple"}],"splittableMap":[["12345"],["5678"]],"simple":{"string":"Simple"}}
class com.google.web.bindery.autobean.vm.impl.ProxyAutoBean
```
No new properties added to the generated JSON, and yet definitely reading from and writing to a ProxyAutoBean.

user/src/com/google/web/bindery/autobean/vm/impl/ProxyAutoBean.java
```diff
@@ -112,34 +112,37 @@ private static Map<String, Data> calculateData(Class<?> beanType) {
...
-            String name = method.getName();
+        if (!(Collection.class.isAssignableFrom(beanType)
+            || Map.class.isAssignableFrom(beanType))) {
+          for (Method method : beanType.getMethods()) {
+            if (BeanMethod.GET.matches(method)) {
+              // match methods on their name for now, to find the most specific
+              // override
+              String name = method.getName();
 
             Type genericReturnType = TypeUtils.resolveGenerics(beanType, method.getGenericReturnType());
```
niloc132, line 123: Needs an indent too - and should wrap to another line

user/src/com/google/web/bindery/autobean/vm/impl/ProxyAutoBean.java
```diff
@@ -112,34 +112,37 @@ private static Map<String, Data> calculateData(Class<?> beanType) {
       if (toReturn == null) {
         Map<String, Data> getters = new HashMap<String, Data>();
         List<Method> setters = new ArrayList<Method>();
-        for (Method method : beanType.getMethods()) {
-          if (BeanMethod.GET.matches(method)) {
-            // match methods on their name for now, to find the most specific
-            // override
-            String name = method.getName();
+        if (!(Collection.class.isAssignableFrom(beanType)
+            || Map.class.isAssignableFrom(beanType))) {
```
niloc132, line 116: I think this should be
```suggestion
        if (!(Collection.class.isAssignableFrom(beanType)
            && !Map.class.isAssignableFrom(beanType))) {
```
so that we don't run this code with either maps or collections (since this is explicitly for beans)?

niloc132: Ignore, my comment is wrong.

Issue #9831, opened 2023-05-19
Java 11 JRE emulation


rjeeb:
Methods to support:

[String](https://github.com/gwtproject/gwt/pull/9975):
- `lines()`
- `isBlank()`
- `format()` (will not implement, this is missing deliberately from earlier versions of String)
- `repeat()`
- `stripTrailing()`
- `stripLeading()`
- `strip()`

[Predicate](https://github.com/gwtproject/gwt/pull/9862) (done): 
- `not()`

[OptionalDouble](https://github.com/gwtproject/gwt/pull/9853) (done):
- `isEmpty()`

[OptionalDouble](https://github.com/gwtproject/gwt/pull/9853) (done):
- `isEmpty()`

[OptionalLong](https://github.com/gwtproject/gwt/pull/9853) (done):
- `isEmpty()`

CharSequence was split into https://github.com/gwtproject/gwt/issues/9991 so this could be closed as complete.

niloc132:
`Predicate` and `Optional*` are all merged. 

`String.format()` is likely not feasible - it requires parsing the format string at runtime, and handling the various possible patterns. That sort of code can't be optimized by the compiler without knowing ahead of time what all format strings will be. 
https://github.com/gwtproject/gwt/blob/e981e97325e0bbf8b16b407143ac297d7f07be01/user/super/com/google/gwt/emul/java/lang/String.java#L71-L73

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Formatter.html#syntax

In theory we could require that the format param will be a string literal and pre-compile the formats so that unused formatting code can be pruned. In practice, that sort of change likely wouldn't be accepted to j2cl, so we should consider carefully if we want to introduce that incompatibility.

zbynek:
Also CharSequence.chars is already implemented: https://github.com/gwtproject/gwt/blob/22ff07e00d2d0d9b224397cbfc2cb7e2e248bf7d/user/super/com/google/gwt/emul/java/lang/CharSequence.java#L41

niloc132:
Closing as fixed, all emulation mentioned here is fixed (CharSequence methods split into 
https://github.com/gwtproject/gwt/issues/9991 for later followup).

PR #9906, opened 2024-01-21
Missing string methods (1. try)

hjhgitw:

addressed in this pull request, e.g. 

Partial #9831

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,23 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = false
+max_line_length = 140
+tab_width = 2
+ij_continuation_indent_size = 2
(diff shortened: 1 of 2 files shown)


zbynek:
@hjhgitw please don't add the editorconfig here, let's keep that in the [separate PR](https://github.com/gwtproject/gwt/pull/9903).

Please add a test that shows that your addition works.

Most importantly: I don't think native string has either of the methods you're trying to call, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String

For `lines` you want something like `Arrays.stream(this.split("\\r\\n?|\\n"))` (see [Javadoc](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#lines()))

For `isBlank()` you can probably iterate through all the characters and check `Character.isWhitespace`.

PR #9914, opened 2024-01-31
GWT-RPC's type signature for java.lang.Enum should be consistent

### LINKED ISSUE #9912: Enum and Enum[] can't be received on Java 21
Running `com.google.gwt.user.client.rpc.CollectionsTest#testEnumArray` on Java 21 results in an error on the server:
```
com.google.gwt.user.client.rpc.SerializationException: Invalid type signature for [Ljava.lang.Enum;
	at com.google.gwt.user.server.rpc.impl.ServerSerializationStreamReader.validateTypeVersions(ServerSerializationStreamReader.java:1154)
	at com.google.gwt.user.server.rpc.impl.ServerSerializationStreamReader.deserialize(ServerSerializationStreamReader.java:634)
	at com.google.gwt.user.server.rpc.impl.ServerSerializationStreamReader.readObject(ServerSerializationStreamReader.java:591)
	at com.google.gwt.user.server.rpc.impl.ServerSerializationStreamReader$ValueReader$8.readValue(ServerSerializationStreamReader.java:148)
	at com.google.gwt.user.server.rpc.impl.ServerSerializationStreamReader.deserializeValue(ServerSerializationStreamReader.java:433)
	at com.google.gwt.user.server.rpc.RPC.decodeRequest(RPC.java:312)
```

At that frame, clientTypeSignature is `3128639609`, and serverTypeSignature is `2029229986`. The former is computed at compile time via `com.google.gwt.user.rebind.rpc.SerializationUtils#getSerializationSignature`, while the later is from `com.google.gwt.user.server.rpc.impl.SerializabilityUtil#encodeSerializedInstanceReference`.

It appears that `java.lang.Enum` added a new field between Java 17 and 21. The special casing in the server code would ensure that enum types will not have this problem, but only code that tries to serialize `java.lang.Enum` and arrays thereof will see this difference (not a very common case). 

Prior to Java 21, Enum had two fields, name and ordinal (both final, so not serialized by default). Then [in Java 21, Enum had a `hash` field added](https://github.com/openjdk/jdk/commit/3da987adacc3acc120a2781ee9a2111e26356723). This is somewhat similar to the Java 9 change in JEP-254 to alter the internal representation of Strings to use `byte[]` rather than `char[]` - except GWT was unaffected by this, as GWT has a custom field serializer for String.

The `java.lang.Enum` type could have a custom field serializer like `java.lang.String` does - but we should also keep in mind that this might break compatibility for apps built with an older version of GWT. Instead, I propose that Enum be special cased, and always return a constant CRC that matches what older Java versions and GWT's emulation would do - this is a one-off that is unlikely to repeat itself with other changes to base Java classes (Record, Object... others?).

niloc132:
In Java 21 a new field was added to Enum, making it incompatible with gwt-rpc clients (and older servers). This patch hardcodes the expected fields for java.lang.Enum so that changes of this nature cannot change the type signature.

Fixes #9912

### DIFF
--- a/user/src/com/google/gwt/user/server/rpc/impl/SerializabilityUtil.java
+++ b/user/src/com/google/gwt/user/server/rpc/impl/SerializabilityUtil.java
@@ -915,6 +915,19 @@ public class SerializabilityUtil {
       for (Enum<?> constant : constants) {
         crc.update(constant.name().getBytes(StandardCharsets.UTF_8));
       }
+    } else if (Enum.class.equals(instanceType)) {
+      // Edge case around Java 21 where a new non-final field was added to Enum. While the Enum
+      // class is rarely serialized itself (see the block above to deal with enum types), it can
+      // happen - this code ensures that the server always treats the Enum type the same way,
+      // regardless of JRE changes.
+      // See https://github.com/gwtproject/gwt/issues/9912
+      if (policy.shouldSerializeFinalFields()) {
+        crc.update("name".getBytes(StandardCharsets.UTF_8));
+        crc.update(getSerializedTypeName(String.class).getBytes(StandardCharsets.UTF_8));
+        crc.update("ordinal".getBytes(StandardCharsets.UTF_8));
+        crc.update(getSerializedTypeName(int.class).getBytes(StandardCharsets.UTF_8));
+      }
+      generateSerializationSignature(Object.class, crc, policy);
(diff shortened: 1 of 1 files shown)


niloc132:
Passing all nightly tests on java 11/17/21 https://github.com/vegegoku/gwt/actions/runs/7776962215/job/21204721087.

PR #9917, opened 2024-02-06
Add support for instanceof pattern matching feature

### LINKED ISSUE #9893: Add compiler implementation, tests for Java 16 pattern matching instanceof
See https://openjdk.org/jeps/394.

vegegoku:
Add compiler implementation, tests for Java 16 pattern matching instanceof

JEP : https://openjdk.org/jeps/394
fix #9893
related to #9869

addressed in this pull request, e.g. "Fixes #42" to close the linked
issue, or "Partial #42" to indicate that this merge will address it,
but that it isn't ready to be closed yet.

### DIFF
--- a/dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
@@ -94,6 +94,7 @@ import com.google.gwt.dev.jjs.ast.JType;
 import com.google.gwt.dev.jjs.ast.JUnaryOperator;
 import com.google.gwt.dev.jjs.ast.JUnsafeTypeCoercion;
 import com.google.gwt.dev.jjs.ast.JVariable;
+import com.google.gwt.dev.jjs.ast.JVariableRef;
 import com.google.gwt.dev.jjs.ast.JWhileStatement;
 import com.google.gwt.dev.jjs.ast.js.JMultiExpression;
 import com.google.gwt.dev.jjs.ast.js.JsniClassLiteral;
@@ -1095,9 +1096,70 @@ public class GwtAstBuilder {
     public void endVisit(InstanceOfExpression x, BlockScope scope) {
       try {
         SourceInfo info = makeSourceInfo(x);
-        JExpression expr = pop(x.expression);
+        JExpression expr;
+        JDeclarationStatement jDeclarationStatement = null;
+        if (x.pattern != null) {
+          jDeclarationStatement = (JDeclarationStatement) pop();
+        }
+        expr = pop(x.expression);
         JReferenceType testType = (JReferenceType) typeMap.get(x.type.resolvedType);
-        push(new JInstanceOf(info, testType, expr));
+
+        if (jDeclarationStatement == null) {
+          push(new JInstanceOf(info, testType, expr));
+        } else {
+          // If <expr> is of type X, then
+          //
+          // rewrite (<expr> instanceof Foo foo)
+          // to
+          // Foo foo;
+          // X $instanceof_1;
+          // (($instanceof_1 = <expr>) instanceof Foo && null != (foo = (Foo) $instanceof_1))
+          //
+          // to avoid wrong behavior if o is a method call with side effects
+
+          // Foo foo;
+          String patternDeclarationName = jDeclarationStatement.getVariableRef().getTarget().getName();
+          if (!curMethod.instanceOfDeclarations.containsKey(patternDeclarationName)) {
+            curMethod.body.getBlock().addStmt(0, jDeclarationStatement);
+            curMethod.instanceOfDeclarations.put(patternDeclarationName, jDeclarationStatement);
+          }
+
+          // X $instanceof_1;
+          JType expressionType = typeMap.get(x.expression.resolvedType);
+          JLocal local =
+              createLocal(info, "$instanceOfExpr", expressionType);
+          JDeclarationStatement expressionDeclaration =
+              makeDeclaration(info, local, null);
+          curMethod.body.getBlock().addStmt(0, expressionDeclaration);
+          curMethod.instanceOfDeclarations.put(local.getName(), expressionDeclaration);
+
+          // (Foo) $instanceof_1
+          JVariableRef variableRef = jDeclarationStatement.getVariableRef();
+          JCastOperation jCastOperation =
+              new JCastOperation(info, variableRef.getType(), local.createRef(info));
+
+          // foo = (Foo) $instanceof_1
+          JBinaryOperation assignOperation =
+              new JBinaryOperation(info, variableRef.getType(), JBinaryOperator.ASG, variableRef,
+                  jCastOperation);
+
+          // null != (foo = (Foo) $instanceof_1)
+          JBinaryOperation nullCheckOperation =
+              new JBinaryOperation(info, JPrimitiveType.BOOLEAN, JBinaryOperator.NEQ,
+                  JNullLiteral.INSTANCE, assignOperation);
+
+          // $instanceof_1 = o
+          JBinaryOperation assignLocalOperation =
+              new JBinaryOperation(info, expressionType, JBinaryOperator.ASG, local.createRef(info),
+                  expr);
+
+          // (($instanceof_1 = o) instanceof Foo && null != (foo = (Foo) $instanceof_1))
+          JBinaryOperation rewrittenSwitch =
+              new JBinaryOperation(info, JPrimitiveType.BOOLEAN, JBinaryOperator.AND,
+                  new JInstanceOf(info, testType, assignLocalOperation), nullCheckOperation);
+
+          push(rewrittenSwitch);
+        }
       } catch (Throwable e) {
         throw translateException(x, e);
       }
@@ -3847,6 +3909,7 @@ public class GwtAstBuilder {
     public final Map<LocalVariableBinding, JVariable> locals = Maps.newIdentityHashMap();
     public final JMethod method;
     public final MethodScope scope;
+    public final Map<String, JStatement> instanceOfDeclarations = Maps.newHashMap();
 
     public MethodInfo(JMethod method, JMethodBody methodBody, MethodScope methodScope) {
       this.method = method;
--- a/dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
@@ -1118,7 +1118,8 @@ public class GwtAstBuilder {
           // to avoid side effects from evaluating the original expression twice
 
           // Foo foo;
-          String patternDeclarationName = jDeclarationStatement.getVariableRef().getTarget().getName();
+          String patternDeclarationName = jDeclarationStatement.getVariableRef().getTarget()
+              .getName();
           if (!curMethod.instanceOfDeclarations.containsKey(patternDeclarationName)) {
             curMethod.body.getBlock().addStmt(0, jDeclarationStatement);
             curMethod.instanceOfDeclarations.put(patternDeclarationName, jDeclarationStatement);
@@ -3911,7 +3912,7 @@ public class GwtAstBuilder {
     public final MethodScope scope;
     public final Map<String, JStatement> instanceOfDeclarations = Maps.newHashMap();
 
-    public MethodInfo(JMethod method, JMethodBody methodBody, MethodScope methodScope) {
+    MethodInfo(JMethod method, JMethodBody methodBody, MethodScope methodScope) {
       this.method = method;
       this.body = methodBody;
       this.scope = methodScope;
--- a/dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
@@ -1115,7 +1115,7 @@ public class GwtAstBuilder {
           // X $instanceof_1;
           // (($instanceof_1 = <expr>) instanceof Foo && null != (foo = (Foo) $instanceof_1))
           //
-          // to avoid wrong behavior if o is a method call with side effects
+          // to avoid side effects from evaluating the original expression twice
 
           // Foo foo;
           String patternDeclarationName = jDeclarationStatement.getVariableRef().getTarget().getName();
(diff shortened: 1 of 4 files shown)


FrankHossfeld:
LGTM

FrankHossfeld (approved):

dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
```diff
@@ -1095,9 +1096,44 @@ public void endVisit(Initializer x, MethodScope scope) {
...
         JReferenceType testType = (JReferenceType) typeMap.get(x.type.resolvedType);
-        push(new JInstanceOf(info, testType, expr));
+
+        if (jDeclarationStatement == null) {
+          push(new JInstanceOf(info, testType, expr));
+        } else {
+          String localVariableName = jDeclarationStatement.getVariableRef().getTarget().getName();
+          if (!curMethod.instanceOfDeclarations.containsKey(localVariableName)) {
+            curMethod.body.getBlock().addStmt(0, jDeclarationStatement);
+            curMethod.instanceOfDeclarations.put(localVariableName, jDeclarationStatement);
```
niloc132, line 1113: Admittedly I haven't taken the time to dig in and try this, but I see that `x.pattern` is a jdt `Pattern`, which has a (public?) `secretPatternVariable` field of type `LocalVariableBinding` - does it work to use that in the existing `MethodInfo.locals` map instead of making a new `instanceOfDeclarations` map?

vegegoku: Just to add a clarification on the case we are trying to solve here, in this sample

```java
    Shape shape1 = new Square();
    Shape shape2 = new Square();

    if (shape1 instanceof Square square && square.getLength() > 0) {
    }
    if (shape2 instanceof Square square && square.getLength() > 0) {
    }
```

According to our solution we will be rewriting this to 

```java
    Shape shape1 = new Square();
    Shape shape2 = new Square();
    
    Square square;
    Square square;

    if (shape1 instanceof Square square && square.getLength() > 0) {
    }
    if (shape2 instanceof Square square && square.getLength() > 0) {
    }
```

and we want to avoid that duplicate declaration, I had thoughts of using the pattern secretPatternvariable, but for each instanceof that is a new instance of that variable which means we end up with entries in the method locals.

What I could do is loop over those and try a different way of checking but that sounds like it defeats the purpose of the map and might be confusing, so I decided to introduce my own to avoid this kind of mixed logic. WDYT.

niloc132: I'm convinced (after you showing me live on a call), this probably needs to be this way - the other ways of making implicit locals don't really look like this use case, so we need something closer to what you've written than what already exists, thanks for explaining.

niloc132:
After some discussion, we've discovered that this replacement isn't sufficient, consider the case
```
return hasSideEffects() instanceof Foo foo && foo.isEmpty();
```

The current impl would rewrite to something like this:
```
Foo foo;
return hasSideEffects() instanceof Foo && (foo = (Foo) hasSideEffects()) && foo.isEmpty();
```
which would result in the same method `hasSideEffects()` being invoked twice, when it is only expected to go off once.

We should have a specific test for this case, and instead consider another pattern, maybe closer to:
```
Foo foo;
Object old;
return (old = hasSideEffects()) instanceof Foo && (foo = (Foo) old) && foo.isEmpty();
```

dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
```diff
@@ -1095,9 +1096,45 @@ public void endVisit(Initializer x, MethodScope scope) {
...
+          }
+
+          // rewrite (o instanceof Foo foo)
+          // to
+          // Foo foo;
+          // (o instanceof Foo && null != (foo = (Foo) o))
+          CloneExpressionVisitor cloner = new CloneExpressionVisitor();
+
+          JVariableRef variableRef = jDeclarationStatement.getVariableRef();
+          JCastOperation jCastOperation = new JCastOperation(info, variableRef.getType(), cloner.cloneExpression(expr));
```
niloc132, line 1123: Remember that cloning this will produce unwanted side effects - see my last comment on the PR. Instead we want another local variable.

If `!expr.hasSideEffects()`, then what you have is shorter and better I think. (not sure if the compiler will optimize that automatically).

vegegoku:
> After some discussion, we've discovered that this replacement isn't sufficient, consider the case
> 
> ```
> return hasSideEffects() instanceof Foo foo && foo.isEmpty();
> ```
> 
> The current impl would rewrite to something like this:
> 
> ```
> Foo foo;
> return hasSideEffects() instanceof Foo && (foo = (Foo) hasSideEffects()) && foo.isEmpty();
> ```
> 
> which would result in the same method `hasSideEffects()` being invoked twice, when it is only expected to go off once.
> 
> We should have a specific test for this case, and instead consider another pattern, maybe closer to:
> 
> ```
> Foo foo;
> Object old;
> return (old = hasSideEffects()) instanceof Foo && (foo = (Foo) old) && foo.isEmpty();
> ```

Ok, your are correct, but I still think that is still also wrong in case that `hasSideEffects()` is supposed to change something if called twice, and now we just call it once. WDYT?

user/test-super/com/google/gwt/dev/jjs/super/com/google/gwt/dev/jjs/test/Java17Test.java
```diff
@@ -41,17 +41,54 @@ public interface TextBlock {
   }
 
   public sealed class Shape permits Square, Circle {
+    public static int count = 0;
+
+    public static Shape returnAndIncrement(Shape shape){
```
niloc132, line 46:
```suggestion
    public static Shape returnAndIncrement(Shape shape) {
```

merging up to date will probably catch this with the checkstyle running on java8+ tests now...

dev/core/test/com/google/gwt/dev/jjs/impl/Java17AstTest.java
```diff
@@ -16,12 +16,15 @@
...
+import com.google.gwt.dev.jjs.ast.JAbstractMethodBody;
+import com.google.gwt.dev.jjs.ast.JDeclaredType;
 import com.google.gwt.dev.jjs.ast.JInterfaceType;
 import com.google.gwt.dev.jjs.ast.JProgram;
 import com.google.gwt.dev.jjs.ast.JStringLiteral;
-
+import com.google.gwt.dev.jjs.ast.JType;
 import java.util.Arrays;
 import java.util.List;
+import org.apache.commons.lang3.StringUtils;
```
niloc132, line 27: checkstyle says this file needs cleanup:

```
 /home/runner/work/gwt/gwt/gwt/dev/core/test/com/google/gwt/dev/jjs/impl/Java17AstTest.java:25:1: warning: 'java.util.Arrays' should be separated from previous imports. (com.puppycrawl.tools.checkstyle.checks.imports.ImportOrderCheck)
/home/runner/work/gwt/gwt/gwt/dev/core/test/com/google/gwt/dev/jjs/impl/Java17AstTest.java:27:1: warning: Wrong order for 'org.apache.commons.lang3.StringUtils' import. (com.puppycrawl.tools.checkstyle.checks.imports.ImportOrderCheck)
/home/runner/work/gwt/gwt/gwt/dev/core/test/com/google/gwt/dev/jjs/impl/Java17AstTest.java:234:0: info: Line matches the illegal pattern 'more than one blank line'. (com.puppycrawl.tools.checkstyle.checks.regexp.RegexpCheck)
```

dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
```diff
@@ -1095,9 +1096,70 @@ public void endVisit(Initializer x, MethodScope scope) {
...
+          // rewrite (o instanceof Foo foo)
+          // to
+          // Foo foo;
+          // X o$instanceof_1;
+          // (($instanceof_1 = o) instanceof Foo && null != (foo = (Foo) $instanceof_1))
+          //
+          // to avoid wrong behavior if o is a method call with side effects
+
+          // Foo foo;
+          String patternDeclarationName = jDeclarationStatement.getVariableRef().getTarget().getName();
```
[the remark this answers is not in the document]
vegegoku: Weird, I just ran checkstyle and it passed.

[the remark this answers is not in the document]
niloc132: Yes it passes, but with warnings - have to manually check the file (or the CI logs from reviewdog, which shows only the lines that apply to the diff)

dev/core/test/com/google/gwt/dev/jjs/impl/Java17AstTest.java
```diff
@@ -135,23 +162,244 @@ public void testSwitchExpressionsNotSupported() {
...
-      fail("Compile should have failed but succeeded, switch expressions as initializer should fail.");
-    } catch (Exception e) {
-      if (!(e.getCause() instanceof InternalCompilerException)
-          && !(e instanceof InternalCompilerException)) {
-        e.printStackTrace();
-        fail();
-      }
-      assertEquals("Switch expressions not yet supported", e.getMessage());
-    }
+  public void testInstanceOfPatternMatching() throws UnableToCompleteException {
```
niloc132, line 165: One more set of tests I can think of that might break with the `X $instanceof_1` decl: a `var` type, for something that can't be expressed easily, like an anonymous inner class, or a type intersection (like was fixed by https://github.com/gwtproject/gwt/commit/ffe61834bf8b83fb3d276b5e4c1f9bdf0af4e9c2) since that was also using `typeMap.get(...)`.

```java
var obj = new SomeClass() {
  public String addAnotherField = "hello";
};
if (obj instanceof SomeClass a) {
  //...
}
```
This could really only matter if there was some other logic going on that would let the type be something else entirely, I'm not thinking about "how this makes sense", but "how this might break the new declaration".

```java
var value = true ? "a" : 'c';
if (obj instanceof CharSequence c) {
  //...
}
```
This one sort of makes more sense, but still is mostly about "hmm, that code looks like something else that has been able to break before".

I'm not sure if this test belongs here or in the Java17Test, up to you.

vegegoku: for the first one isnt that a compilation error from the java compiler itself, should we also guard against those, I thought we get those for free?

for the 2nd one I will add the test.

user/test-super/com/google/gwt/dev/jjs/super/com/google/gwt/dev/jjs/test/Java17Test.java
```diff
@@ -76,4 +113,81 @@ private void checkIfCompiled(Shape square, Shape circle) {
...
+
+  public void testMultipleInstanceOfPatternMatchingWithSameVariableName() {
+    Shape shape1 = new Square();
+    Shape shape2 = new Square();
+    boolean a = false;
+    boolean b = false;
+    if (shape1 instanceof Square square && square.getLength() > 0) {
+      a = true;
+    }
+    if (shape2 instanceof Square square && square.getLength() > 0) {
```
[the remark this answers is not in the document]
vegegoku: Didnt get it very well, we define variables from 2 different types and check for the same type? or we define with different types and check for the same type?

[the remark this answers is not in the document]
vegegoku: I can write tests for both anyway.

dev/core/test/com/google/gwt/dev/jjs/impl/Java17AstTest.java
vegegoku:
```
public class Test {

    var testPolygon = new Polygon(){
      public int sideWidth = 10;
    };

    if(testPolygon instanceof Polygon p){
      System.out.println(p.sides);
    }
  }
}
```
```
java: expression type <anonymous Polygon> is a subtype of pattern type Polygon
```

niloc132: Ah - if I've made a mistake here and compiling the code catches (with JDT) it, then it wouldn't be a helpful test - I was trying to craft a bug that would only apply to how we're transforming this new instanceof by introducing new locals. We should not need negative test cases for JDT failures (esp as presumably most users are running their code through javac first anyway).

That said, apparently Java 21 can handle the Polygon example, and refers to it as an "unconditional pattern in instanceof" if you try to compile it with release=17, but if you use Java 17 on this, I do see the error you got. Changing the `var` to Object` lets it work in plain Java 17, but is no longer testing what I was concerned about.

The ternary seems to compile in Java 17 though? I can't say that I understand yet why var is okay for one but not the other...

niloc132:
Style check warnings:
dev:
```
/home/runner/work/gwt/gwt/gwt/dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java:1121:0: warning: Line is longer than 100 characters (found 103). (com.puppycrawl.tools.checkstyle.checks.sizes.LineLengthCheck)
/home/runner/work/gwt/gwt/gwt/dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java:3914:5: warning: Redundant 'public' modifier. (com.puppycrawl.tools.checkstyle.checks.modifier.RedundantModifierCheck)
```
dev test:
```
 /home/runner/work/gwt/gwt/gwt/dev/core/test/com/google/gwt/dev/jjs/impl/Java17AstTest.java:28:1: warning: Wrong order for 'org.apache.commons.lang3.StringUtils' import. (com.puppycrawl.tools.checkstyle.checks.imports.ImportOrderCheck)
/home/runner/work/gwt/gwt/gwt/dev/core/test/com/google/gwt/dev/jjs/impl/Java17AstTest.java:242:0: info: Line matches the illegal pattern 'more than one blank line'. (com.puppycrawl.tools.checkstyle.checks.regexp.RegexpCheck)
```

niloc132 (approved):

jnehlmeier:
Because it just came to my mind:

I think we should add another test that checks if the pattern matching variable is available outside the `if`-block if the condition is `!instanceof`, e.g.

```java
    Object object = "test";
    if (!(object instanceof String string)) {
      throw new RuntimeException();
    }
    System.out.println(string.length());
  }
```

PR #9918, opened 2024-02-08
Un-supersource Java 8-11 tests

niloc132:
Simplify test setup, as these tests no longer need to use supersource.


niloc132:
Looks like the checkstyle changes didn't account for some new warnings that now fail reviewdog - this same code used to be passing, but it seems there was an implicit extra group for standard java packages. Will solve (updating rules, updating java as needed) in a separate PR first.

niloc132:
Can merge after https://github.com/gwtproject/gwt/pull/9920, nightly build showing both of these passing at https://github.com/niloc132/gwt/actions/runs/7848953792.

PR #9920, opened 2024-02-09
Checkstyle import rules should be consistent with the codebase

niloc132:
The old checkstyle was similar but not identical to the Google Java Style Guide, and it seems that checkstyle updates have changed import behaviors in a way to make them incompatible with those conventions (but more compatible with the Google style). These new rules are a close approximation to the old convention, with a few exceptions, mostly that imports are all listed at the top. There were also some inconsistencies, such as grouping jsinterop and javaemul with java packages - those have been resolved as simply as possible, with checkstyle rules to enforce this.


niloc132:
Added a commit which relied on line numbers in Throwable not changing, so that this test still passes.

PR #9934, opened 2024-03-05
Add support for Java 14 Records

### LINKED ISSUE #9894: Add compiler implementation, JRE emulation, and tests for Java 16 record types
See https://openjdk.org/jeps/395.

niloc132:
Implements records as Java objects, so they can be transpiled to JS. Unlike in Java bytecode, hashCode/toString/equals have a complete implementation, rather than using method references and constant bootstrap to generate code at runtime.

Includes proposed jsinterop changes to support records without generating invalid code or compile/runtime errors.

Fixes #9894


niloc132:
Proposed jsinterop spec changes are implemented - building more tests now, then should be ready for review.

vegegoku:
The branch has conflicts with changes from https://github.com/gwtproject/gwt/pull/9917 specifically with the Java17Test classes

user/test-super/com/google/gwt/dev/jjs/super/com/google/gwt/dev/jjs/test/Java17Test.java
```diff
@@ -76,4 +78,143 @@ private void checkIfCompiled(Shape square, Shape circle) {
...
+    assertEquals(new InnerRecord(), new InnerRecord());
+
+    assertEquals("Example", new LocalRecord().toString());
+
+    TopLevelRecord withValues = new TopLevelRecord("Banana", 7);
+    assertTrue(withValues.toString().contains("7"));
+    assertTrue(withValues.toString().contains("Banana"));
+    assertEquals("Banana", withValues.name());
+    assertEquals(7, withValues.count());
+  }
```
vegegoku, line 124: For to string shouldnt we test for the full default format that is expected to match the JVM records to string and make sure the name and the value of each field present?

vegegoku:
I did a local build of GWT and tried to several record types and it compiles and works in both production and SDM as well, breakpoints inside of record constructor and methods also works as expected.

user/test-super/com/google/gwt/dev/jjs/super/com/google/gwt/dev/jjs/test/Java17Test.java
niloc132: My interpretation of the spec is that as long as it contains the name of the class, the name of each component, and the value of each component, we're within the spec. I did try to match pretty exactly to what OpenJDK will return.

https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.10
> A method public final String toString() that returns a string derived from the name of the record class and the names and string representations of every record component of R.
> The string representation of a record component c of an instance a of a record class is as follows:
> -   If the type of the record component c is a reference type, then the string representation is determined as if by invoking the toString method on the value of the component field c of a.
> -   If the type of the record component c is a primitive type T, then the string representation is determined as if by subjecting the value of the component field c of a to boxing conversion ([§5.1.7](https://docs.oracle.com/javase/specs/jls/se17/html/jls-5.html#jls-5.1.7)) and then invoking the method toString method of the wrapper class corresponding to T on the resulting object.


For the sake of respective the obfuscated type settings of a given application, I interpreted "the name of the record class" as "the value returned from `this.getClass().getSimpleName()`, so that the name will be consistently obfuscated, see ImplementRecordComponents.java:
```java
  private void implementToString(JRecordType type, JMethod method, SourceInfo info) {
    List<JExpression> args = new ArrayList<>();

    // Concatenate type with []s and component values, assigning to toStrExpr as we append
    // more concat operations. Using getClass().getSimpleName() rather than a string literal
    // allows the toString implementation to emit the obfuscated class name.
    JMethodCall getClass = new JMethodCall(info, new JThisRef(info, type), getClassMethod);
    JExpression toStrExpr = new JMethodCall(info, getClass, getSimpleNameMethod);
```

vegegoku (approved):

dev/core/src/com/google/gwt/dev/jjs/impl/ImplementRecordComponents.java
```diff
@@ -0,0 +1,231 @@
...
+    JLocal typedOther = JProgram.createLocal(info, "other", type, true, body);
+    // We can use an unsafe cast since we know the check will succeed
+    JUnsafeTypeCoercion uncheckedCast =
+            new JUnsafeTypeCoercion(info, type, otherParam.createRef(info));
+    JBinaryOperation uncheckedAssign = new JBinaryOperation(info, type, JBinaryOperator.ASG,
+            typedOther.createRef(info), uncheckedCast);
+    body.getBlock().addStmt(uncheckedAssign.makeStatement());
+
+    JExpression componentCheck = JBooleanLiteral.TRUE;
+    for (JField field : type.getFields()) {
```
jnehlmeier, line 180: This creates an `equals` method that uses `==` for all fields. It will fail for reference types that are equal. You should generate `Objects.equals(this.a, other.a)` for reference types and `PrimitiveWrapper.compare(this.a, other.a) == 0` for primitive types. This is mentioned in the implementation requirements section in the [javadoc of Record.equals()](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/Record.html#equals(java.lang.Object)).

Current tests don't catch it because all records being used in tests do not use reference types. Below I added a comment to introduce `RecordWithReferenceType(TopLevelRecord){}`. With the current `equals` implementation a test using this record should fail.

dev/core/test/com/google/gwt/dev/jjs/impl/JChangeTrackingVisitorTest.java
```diff
@@ -46,7 +48,10 @@ public AddParamWhenEnterMethodVisitor(OptimizerContext optimizerCtx) {
 
     @Override
     public boolean enter(JMethod x, Context ctx) {
-      x.createParameter(SourceOrigin.UNKNOWN, "_newParam_enter", JPrimitiveType.INT);
```
jnehlmeier, line 49: How do these changes in this class relate to records?

jnehlmeier (changes requested):

dev/core/src/com/google/gwt/dev/jjs/impl/ImplementRecordComponents.java
niloc132: Eesh, good catch.

dev/core/test/com/google/gwt/dev/jjs/impl/JChangeTrackingVisitorTest.java
niloc132: The answer is "Objects.hash(Object...)". Thanks for asking, I had forgotten this headache.

So some parts of gwt-dev depend on gwt-user (while of course gwt-user depends on gwt-dev...) especially for tests. Other tests do not use gwt-user, and mock out their own classes.

The `JavaResourceBase` type is a container of a bunch of partially-implemented classes - just enough to get past the bare bones requirements of "this is essentially Java, and so your code will work. In this patch, we needed to add Record both as Record.java and as a mock. We likewise needed ObjectMethods in both places despite the fact that it has no members - this satisfies JDT, which thinks we're going to need it for runtime MethodHandle creation for each of our Record types. Finally, Objects. 

Objects.hash() is the first method in this super-barebones JRE that uses varargs. 

And JChangeTrackingVisitorTest.AddParamWhenEnterMethodVisitor was implemented assuming there was no such thing as varargs. It is _possible_ (but not very likely) too that JChangeTrackingVisitor itself could have bugs around varargs since it was clearly not tested. Adding Objects.hash() thus requires that this test be updated to allow that varargs are a real thing.

If we'd made a new "mini-JRE" for each test, this wouldn't have happened, since this test wouldn't have failed with the addition of varargs for a different test's JRE. Likewise, if these were all built as integration tests and so used the "real" GWT jre emul, this would have been fixed long ago. But neither of those was done, so I got to find this surprise while implementing the feature.

user/test-super/com/google/gwt/dev/jjs/super/com/google/gwt/dev/jjs/test/Java17Test.java
```diff
@@ -114,6 +116,145 @@ private void checkIfCompiled(Shape square, Shape circle) {
...
```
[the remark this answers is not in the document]
niloc132: Thanks, exposed a vararg bug in how we compute hashcode too... oddly not a compiler issue, but a runtime class cast exception.

jnehlmeier (approved):

niloc132:
One more push to add Java 22 CI and pick up any changes from main.

Full CI run: https://github.com/niloc132/gwt/actions/runs/10319416307

Issue #9790, opened 2022-12-22
protobuf-java CVE-2022-3171 on gwt-servlet


archanakeshireddy:
**GWT version:2.8.2 and 2.10.0

##### Description
Our dependancy check report is reporting CVE for our project using gwt, because gwt-servlet depends on protobuf-java 2.5.0.

I found similar issue https://github.com/gwtproject/gwt/issues/9749
Is it safe to assume that gwt-servlet does not use google protobuf java.

I saw an another issue https://github.com/gwtproject/gwt/issues/9752I where plan is remove the proto-buf pom.xml from META-INF folder.
Is it safe for us if we remove this pom.xml and recompile gwt so our dependancy check report does not report this issue.

niloc132:
Please test the fix posted at https://github.com/gwtproject/gwt/pull/9785 - it is deployed to a maven server, so you can verify the fixed jar and scan one. Lots of people have reported this issue, but no one has tested that the fix stops scanners from alerting about this, so we haven't merged the fix yet.

ajgbarnes:
Thank you for the fix @niloc132 - will test and see if it passes our SonarQube scan.

Apologies for the naive question - is the vulnerability still there with this fix just hidden or was it a false positive?  If so, why?  Will check the other Issues posted against this repository but would appreciate your advice/feedback.

niloc132:
Its a reasonable question, but we have to get a bit philosophical, so forgive my answer being a little too self-indulgent. For a shorter answer, see https://github.com/gwtproject/gwt/issues/9659#issuecomment-481602218.

What is a security vulnerability? Is it enough to have a line of code, that if someone calls it wrong, could permit a denial of service like this one? Or does someone have to call that line for it to matter?

I'd suggest that there are many such lines of code that could be called in Java itself, like ByteBuffer.allocate(bignumber) or something else that might take a long time or consume a lot of resources. But while protobuf-java is considered to have a vulnerability here, why isn't Java itself? 

[...]

Issue #9931, opened 2024-02-26
StackTraceDeobfuscator sometimes produces total nonsense


jnehlmeier:
**GWT version:** head

---

##### Description

Once the application is large enough I think a hidden bug in GWTs symbol maps / source maps appears more and more often. It results in deobfuscated stack traces with incorrect file and line numbers.

Consider the following example:
```java
at java.lang.Throwable.Throwable(com/google/common/base/MoreObjects.java:437)
	at java.lang.RuntimeException.RuntimeException(com/google/common/base/Preconditions.java:1422)
	at com.google.web.bindery.event.shared.UmbrellaException.UmbrellaException(UmbrellaException.java:142)
	at com.google.gwt.event.shared.UmbrellaException.UmbrellaException(com/google/common/base/Splitter.java:424)
	at com.google.gwt.user.client.ui.AttachDetachException.AttachDetachException(com/google/common/collect/RegularImmutableAsList.java:44)
	at com.google.gwt.user.client.ui.AttachDetachException.tryCommand(com/google/common/collect/RegularImmutableAsList.java:34)
	at com.google.gwt.user.client.ui.Panel.doDetachChildren(Panel.java:108)
	at com.google.gwt.user.client.ui.Widget.$onDetach(com/google/common/base/Joiner.java:276)
	at com.google.gwt.user.client.ui.Widget.onDetach(Widget.java:108)
	at com.google.gwt.user.client.ui.Composite.$onDetach(com/google/common/collect/Iterators.java:1433)
	at com.google.gwt.user.client.ui.Composite.onDetach(Composite.java:225)
	at com.google.gwt.user.client.ui.AttachDetachException$2.execute(AttachDetachException.java:249)
	at com.google.gwt.user.client.ui.AttachDetachException.tryCommand(com/google/common/collect/RegularImmutableAsList.java:48)
```

For some reason it often contains Guava classes as files for GWT classes but it can also contain other wrong files from the application itself or GWT SDK.

##### Steps to reproduce

Difficult to reproduce in an example project

##### Known workarounds

None

niloc132:
[...]

PR #9936, opened 2024-03-10
Enable embedding sources inside the sourcemaps json as sourcesContent

### LINKED ISSUE #9752: Please remove protobuf pom from gwt-servlet

vegegoku:
This PR updates the rebased version of sourcemaps jar and other dependencies to support embedding the source files content into the source maps sourcesContent field.

requires : https://github.com/gwtproject/tools/pull/31

Fixes #9752 #9778 #9790

### DIFF
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -29,8 +29,7 @@ jobs:
       - name: Checkout GWT tools into a sibling directory
         uses: actions/checkout@v4
         with:
-          repository: 'vegegoku/tools'
-          ref: 'source-maps'
+          repository: 'gwtproject/tools'
           path: 'tools'
       - name: Set up JDK ${{ matrix.java-version }}
         # GWT requires Java 11+ to build
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -29,7 +29,8 @@ jobs:
       - name: Checkout GWT tools into a sibling directory
         uses: actions/checkout@v4
         with:
-          repository: 'gwtproject/tools'
+          repository: 'vegegoku/tools'
+          ref: 'source-maps'
           path: 'tools'
       - name: Set up JDK ${{ matrix.java-version }}
         # GWT requires Java 11+ to build
--- a/dev/build.xml
+++ b/dev/build.xml
@@ -68,11 +68,12 @@
           <include name="objectweb/asm-9.6/asm-9.6.jar"/>
           <include name="objectweb/asm-9.6/asm-commons-9.6.jar"/>
           <include name="objectweb/asm-9.6/asm-util-9.6.jar"/>
-          <include name="guava/guava-19.0/guava-19.0-rebased.jar"/>
+          <include name="guava/guava-33.0/guava-33.0.0-jre-rebased.jar"/>
+          <include name="guava/guava-33.0/failureaccess-1.0.2-rebased.jar"/>
           <include name="icu4j/63.1/icu4j.jar"/>
           <include name="jetty/jetty-9.4.44.v20210927/jetty-all-9.4.44.v20210927.jar"/>
           <include name="gson/gson-2.6.2.jar"/>
-          <include name="jscomp/20160315/sourcemap-rebased.jar"/>
+          <include name="jscomp/20231112/sourcemap-rebased.jar"/>
           <include name="jsr305/jsr305.jar"/>
           <include name="protobuf/protobuf-2.5.0/protobuf-java-rebased-2.5.0.jar"/>
           <!-- dependencies needed for JSP support in DevMode: BEGIN -->
@@ -129,7 +130,8 @@
               src="${gwt.tools.lib}/eclipse/org.eclipse.jdt.core_3.32.0.v20221108-1853.jar"/>
           <zipfileset
               src="${gwt.tools.lib}/eclipse/jdtCompilerAdapter_3.32.0.v20221108-1853.jar"/>
-          <zipfileset src="${gwt.tools.lib}/guava/guava-19.0/guava-19.0-rebased.jar"/>
+          <zipfileset src="${gwt.tools.lib}/guava/guava-33.0/guava-33.0.0-jre-rebased.jar"/>
+          <zipfileset src="${gwt.tools.lib}/guava/guava-33.0/failureaccess-1.0.2-rebased.jar" />
           <zipfileset src="${gwt.tools.lib}/icu4j/63.1/icu4j.jar"/>
           <zipfileset
               src="${gwt.tools.lib}/jetty/jetty-9.4.44.v20210927/jetty-all-9.4.44.v20210927.jar">
@@ -140,7 +142,7 @@
             <exclude name="META-INF/services/org.eclipse.jetty.security.Authenticator$Factory"/>
           </zipfileset>
           <zipfileset src="${gwt.tools.lib}/gson/gson-2.6.2.jar"/>
-          <zipfileset src="${gwt.tools.lib}/jscomp/20160315/sourcemap-rebased.jar"/>
+          <zipfileset src="${gwt.tools.lib}/jscomp/20231112/sourcemap-rebased.jar"/>
           <zipfileset src="${gwt.tools.lib}/jsr305/jsr305.jar"/>
           <zipfileset
               src="${gwt.tools.lib}/protobuf/protobuf-2.5.0/protobuf-java-rebased-2.5.0.jar"/>
@@ -227,9 +229,11 @@
         <pathelement
             location="${gwt.tools.lib}/eclipse/jdtCompilerAdapter_3.32.0.v20221108-1853.jar"/>
         <pathelement
-            location="${gwt.tools.lib}/guava/guava-19.0/guava-19.0-rebased.jar"/>
+            location="${gwt.tools.lib}/guava/guava-33.0/guava-33.0.0-jre-rebased.jar"/>
+        <pathelement
+            location="${gwt.tools.lib}/guava/guava-33.0/failureaccess-1.0.2-rebased.jar"/>
         <pathelement location="${gwt.tools.lib}/gson/gson-2.6.2.jar"/>
-        <pathelement location="${gwt.tools.lib}/jscomp/20160315/sourcemap-rebased.jar"/>
+        <pathelement location="${gwt.tools.lib}/jscomp/20231112/sourcemap-rebased.jar"/>
         <pathelement location="${gwt.tools.lib}/json/android-sdk-19.1/json-android-rebased.jar"/>
         <pathelement location="${gwt.tools.lib}/jsr305/jsr305.jar"/>
         <pathelement
--- a/dev/core/src/com/google/gwt/core/ext/linker/impl/StandardLinkerContext.java
+++ b/dev/core/src/com/google/gwt/core/ext/linker/impl/StandardLinkerContext.java
@@ -131,6 +131,7 @@ public class StandardLinkerContext extends Linker implements LinkerContext {
   private ResourceOracle publicResourceOracle;
 
   private final SortedSet<SelectionProperty> selectionProperties;
+  private final ModuleDef module;
 
   public StandardLinkerContext(TreeLogger logger, ModuleDef module,
       ResourceOracle publicResourceOracle, JsOutputOption outputOption)
@@ -143,6 +144,7 @@ public class StandardLinkerContext extends Linker implements LinkerContext {
     this.moduleLastModified = module.lastModified();
     this.publicResourceOracle = publicResourceOracle;
     this.outputOption = outputOption;
+    this.module = module;
 
     // Sort the linkers into the order they should actually run.
     linkerClasses = new ArrayList<Class<? extends Linker>>();
@@ -535,6 +537,10 @@ public class StandardLinkerContext extends Linker implements LinkerContext {
         Utility.close(artifactStream);
       }
     }
+  }
+
+  public ModuleDef getModule() {
+    return module;
   }
 
   /**
--- a/dev/core/src/com/google/gwt/core/linker/SymbolMapsLinker.java
+++ b/dev/core/src/com/google/gwt/core/linker/SymbolMapsLinker.java
@@ -15,6 +15,9 @@
  */
 package com.google.gwt.core.linker;
 
+import static java.util.Objects.isNull;
+import static java.util.Objects.nonNull;
+
 import com.google.gwt.core.ext.LinkerContext;
 import com.google.gwt.core.ext.TreeLogger;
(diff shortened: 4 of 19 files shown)


dev/build.xml
```diff
@@ -142,8 +143,9 @@
           <zipfileset
               src="${gwt.tools.lib}/eclipse/org.eclipse.jdt.core_3.17.0.v20190306-2240.jar"/>
           <zipfileset
-              src="${gwt.tools.lib}/eclipse/jdtCompilerAdapter_3.17.0.v20190306-2240.jar"/>
-          <zipfileset src="${gwt.tools.lib}/guava/guava-19.0/guava-19.0-rebased.jar"/>
+              src="${gwt.tools.lib}/eclipse/jdtCompilerAdapter_3.32.0.v20221108-1853.jar"/>
```
niloc132, line 146: if you're fixing this, should fix the line above too...

dev/core/src/com/google/gwt/core/linker/SymbolMapsLinker.java
```diff
@@ -313,13 +326,46 @@ public ArtifactSet link(TreeLogger logger, LinkerContext context,
...
+
+            Map<String, EmittedArtifact> generatedSources = new java.util.HashMap<>();
+            artifacts.find(EmittedArtifact.class)
+                .forEach(emittedArtifact -> {
+                    if (Visibility.Source == emittedArtifact.getVisibility()) {
+                      generatedSources.put(emittedArtifact.getPartialPath(), emittedArtifact);
+                    }
                 });
+
+            StandardLinkerContext stdContext = (StandardLinkerContext) context;
```
niloc132, line 352: As long as there is a cast here, might as well just access the module itself just once, rather than each time through the loop?

niloc132 (commented):
Some updates from private chat working on this feature:
 * We've learned that SDM won't be improved by this, as the various linking activity doesn't copy source files for in SDM, only in prod code with sourcemaps enabled. The "sourcesContent" part of this patch will only improve SDM in cases where the debugger loads many many sources, to the point where that network traffic could be time better spent reading a giant sourcemap file.
 * This is still a legit way to slightly improve prod build times for projects that deploy with sourcemaps (not many, but they exist), though it might be counterproductive for those only using sourcemaps for stack trace deobfuscation
 * For these reasons, we're leaning towards a config property in CompilerProperties.gwt.xml that would activate this new feature.
 * The patch still has value simply by updating the sourcemap library: both to potentially fix any issues in the last 8ish years since this was updated, and to drop the server-side runtime dependency on protobuf once and for all.

There is also other cleanup possible for compiler.useSourceMaps (since it should work in all browsers today, and not need extra wiring in CoreWithUserAgent), but let's leave that to another patch.

niloc132:
Description should be updated to indicate that this fixes https://github.com/gwtproject/gwt/issues/9752.

dev/core/src/com/google/gwt/core/linker/SymbolMapsLinker.java
```diff
@@ -59,6 +72,7 @@
 public class SymbolMapsLinker extends AbstractLinker {
 
   public static final String MAKE_SYMBOL_MAPS = "compiler.useSymbolMaps";
+  public static final String EMBED_SOURCE_MAPS_SOURCES = "compiler.embedSourceMapsSources";
```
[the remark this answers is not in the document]
vegegoku: This actually needs to be removed, this is now is read from the [ModuleDef](https://github.com/vegegoku/gwt/blob/b39f264bdf9fac87d93c034b9f83fd28d7c6efd8/dev/core/src/com/google/gwt/dev/cfg/ModuleDef.java#L637) which defaults to `false` if the property is not present.

dev/core/src/com/google/gwt/core/linker/SymbolMapsLinker.java
```diff
@@ -335,6 +358,70 @@ public Object merge(String extKey, Object oldVal, Object newVal) {
...
+                                               SourceMapGeneratorV3 sourceMapGenerator,
+                                               int totalPrefixLines, String sourceMapString,
+                                               String partialPath)
+      throws SourceMapParseException {
+    sourceMapGenerator.setStartingPosition(totalPrefixLines, 0);
+    SourceMapConsumerV3 section = new SourceMapConsumerV3();
+    section.parse(sourceMapString);
+    section.visitMappings(sourceMapGenerator::addMapping);
+
+    Iterator extensions = section.getExtensions().entrySet().iterator();
```
niloc132, line 372: fix raw generics here, then the loop below won't need the cast?

dev/core/src/com/google/gwt/core/linker/SymbolMapsLinker.java
```diff
@@ -335,6 +358,70 @@ public Object merge(String extKey, Object oldVal, Object newVal) {
...
+
+    while (extensions.hasNext()) {
+      Entry<String, Object> entry = (Entry) extensions.next();
+      String extensionKey = (String) entry.getKey();
+      sourceMapGenerator.addExtension(extensionKey, entry.getValue());
+    }
+
+    ResourceLoader resourceLoader = ResourceLoaders.fromContextClassLoader();
+
+    Map<String, EmittedArtifact> generatedSources = new java.util.HashMap<>();
```
niloc132, line 382: import HashMap - or probably more consistent with the rest of the codebase, use the factory method in Maps?

dev/core/src/com/google/gwt/core/linker/SymbolMapsLinker.java
```diff
@@ -335,6 +358,70 @@ public Object merge(String extKey, Object oldVal, Object newVal) {
...
+            if (Visibility.Source == emittedArtifact.getVisibility()) {
+              generatedSources.put(emittedArtifact.getPartialPath(), emittedArtifact);
+            }
+        });
+
+    for (String sourceFileName : section.getOriginalSources()) {
+      String content;
+      try {
+        InputStream cis = loadSource(logger, sourceFileName, generatedSources,
+            resourceLoader);
```
niloc132, line 394: Should be try-with-resources, right, so we are sure to close the file?

dev/core/src/com/google/gwt/core/linker/SymbolMapsLinker.java
```diff
@@ -335,6 +358,70 @@ public Object merge(String extKey, Object oldVal, Object newVal) {
...
+                                        ResourceLoader resourceLoader)
+      throws UnableToCompleteException, URISyntaxException, IOException {
+    if (generatedSources.containsKey(sourceFileName)) {
+      return generatedSources.get(sourceFileName).getContents(logger);
+    } else {
+      // ask the resourceOracle for the file contents and add it
+      URL resource = resourceLoader.getResource(sourceFileName);
+      if (nonNull(resource)) {
+        URI uri = resource.toURI();
+        // Use Files.readAllBytes and Paths.get for Java 8 compatibility
```
niloc132, line 418: Sorry, what is this comment saying?

dev/core/src/com/google/gwt/core/linker/SymbolMapsLinker.java
vegegoku: The line after that is a check for a null and try to use a difference resource for loading the content, so I guess using the try with resource wont work here, but we can add a finally block to ensure we close the InputStream.

dev/core/src/com/google/gwt/core/linker/SymbolMapsLinker.java
vegegoku: the line above should be actually removed as it is unused.

dev/core/src/com/google/gwt/core/ext/linker/impl/StandardLinkerContext.java
```diff
@@ -537,6 +539,10 @@ public void produceOutput(TreeLogger logger, ArtifactSet artifacts,
     }
   }
 
+  public ModuleDef getModule() {
```
niloc132, line 542: As we discussed when first working out how to plumb this, I don't really like exposing this here. I don't have a good enough answer to ask you to change it, so just documenting my thoughts in case something strikes one of us (or someone sees this PR in a few years and wants to know why on earth we made this change).

One option is to make a specific getter, `shouldEmbedSourcesInModuleSourcemaps()` sort of thing that just does `return module.isEmbedSourcesContent()`. That would better match getModuleName()/getModuleFunctionName()/getModuleLastModified() - though those three are part of LinkerContext itself. I don't think this should be added to LinkerContext. 

I dislike this option the least. However, it does suggest that we should do the same in CompilerContext - but CompilerContext has a bunch of other calls to getModule() already, so it feels a bit inconsistent.

Another option would be to remove the other three getters, and have all the callers instead call `getModule().getName()` and so on. This provides access to too much of ModuleDef I think, and it isn't immutable, so I really don't care for this. 

---

Separately, I don't love the name `isEmbedSourcesContent` as a method on ModuleDef, nor do I love that the method exists there at all. No other properties exist on ModuleDef itself, so it seems silly to have this one be special enough to be there. That said, it is needed by both the various pre/compile stages and linkers too, so it either needs to be somewhere very accessible, or needs to be duplicated - so I think I'm okay with it staying in ModuleDef _or_ pushing it down into Properties (so you have to call `getModule().getProperties.isBlahBlah()` - that does happen in other places, so it wouldn't be terrible). Renaming it: how about we get SourceMaps in the name somehow - shouldEmbedSourceMapContents() perhaps? sourceMapContentsEmbedEnabled()?

niloc132 (commented):
Looks like there may be more classpaths that _should_ be expanded, or code that needs to be corrected to stop walking classes it should ignore (annotations mostly?)

Example from the build:
```
requestfactory-server:
     [java] Could not find class file for com/google/auto/value/AutoValue
     [java] Could not find class file for com/google/auto/value/AutoValue$Builder
     [java] May 12, 2024 4:55:17 AM com.google.auto.value.AutoValue
     [java] SEVERE: Could not find class file for com/google/auto/value/AutoValue
     [java] May 12, 2024 4:55:17 AM com.google.auto.value.AutoValue.Builder
     [java] SEVERE: Could not find class file for com/google/auto/value/AutoValue$Builder
     [java] Could not find class file for com/google/errorprone/annotations/CanIgnoreReturnValue
     [java] May 12, 2024 4:55:17 AM com.google.errorprone.annotations.CanIgnoreReturnValue
     [java] SEVERE: Could not find class file for com/google/errorprone/annotations/CanIgnoreReturnValue
     [java] Could not find class file for com/google/errorprone/annotations/DoNotCall
     [java] May 12, 2024 4:55:17 AM com.google.errorprone.annotations.DoNotCall
     [java] SEVERE: Could not find class file for com/google/errorprone/annotations/DoNotCall
     [java] Could not find class file for com/google/errorprone/annotations/InlineMe
     [java] May 12, 2024 4:55:17 AM com.google.errorprone.annotations.InlineMe
     [java] SEVERE: Could not find class file for com/google/errorprone/annotations/InlineMe
     [java] Could not find class file for com/google/errorprone/annotations/DoNotMock
     [java] May 12, 2024 4:55:17 AM com.google.errorprone.annotations.DoNotMock
     [java] SEVERE: Could not find class file for com/google/errorprone/annotations/DoNotMock
     [java] Could not find class file for com/google/errorprone/annotations/InlineMeValidationDisabled
     [java] May 12, 2024 4:55:17 AM com.google.errorprone.annotations.InlineMeValidationDisabled
     [java] SEVERE: Could not find class file for com/google/errorprone/annotations/InlineMeValidationDisabled
     [java] Could not find class file for com/google/errorprone/annotations/CompatibleWith
     [java] May 12, 2024 4:55:17 AM com.google.errorprone.annotations.CompatibleWith
     [java] SEVERE: Could not find class file for com/google/errorprone/annotations/CompatibleWith
     [java] Could not find class file for com/google/errorprone/annotations/Immutable
     [java] May 12, 2024 4:55:17 AM com.google.errorprone.annotations.Immutable
     [java] SEVERE: Could not find class file for com/google/errorprone/annotations/Immutable
     [java] Could not find class file for com/google/errorprone/annotations/ForOverride
     [java] Could not find class file for com/google/errorprone/annotations/CheckReturnValue
     [java] Could not find class file for com/google/errorprone/annotations/RestrictedApi
     [java] May 12, 2024 4:55:17 AM com.google.errorprone.annotations.ForOverride
     [java] SEVERE: Could not find class file for com/google/errorprone/annotations/ForOverride
     [java] May 12, 2024 4:55:17 AM com.google.errorprone.annotations.CheckReturnValue
     [java] SEVERE: Could not find class file for com/google/errorprone/annotations/CheckReturnValue
     [java] May 12, 2024 4:55:17 AM com.google.errorprone.annotations.RestrictedApi
     [java] SEVERE: Could not find class file for com/google/errorprone/annotations/RestrictedApi
     [java] Could not find class file for com/google/errorprone/annotations/concurrent/GuardedBy
     [java] May 12, 2024 4:55:17 AM com.google.errorprone.annotations.concurrent.GuardedBy
     [java] SEVERE: Could not find class file for com/google/errorprone/annotations/concurrent/GuardedBy
```

niloc132 (approved):
Approved, let's merge the tools PR, and remove the CI changes here, and give everyone else another day or so to see if they have any last minute feedback.

niloc132:
Sorry, still lots of build warnings, I think more dependencies need to be added (or the RF jar extractor tool updated to otherwise handle this case): https://github.com/gwtproject/gwt/actions/runs/9196862375/job/25295956029?pr=9936#step:6:592

niloc132:
Confirmed that this is fixed, waiting for full build to merge.

niloc132:
Passing build: https://github.com/vegegoku/gwt/actions/runs/9975626940/job/27565849685

.github/workflows/full-check.yml
```diff
@@ -29,7 +29,8 @@ jobs:
       - name: Checkout GWT tools into a sibling directory
         uses: actions/checkout@v4
         with:
-          repository: 'gwtproject/tools'
+          repository: 'vegegoku/tools'
```
niloc132, line 32: With the tools change merged, these need to be reverted to land this.

niloc132:
Full build after dropping `gwtproject/tools` workaround https://github.com/vegegoku/gwt/actions/runs/9978374423

jnehlmeier:
Running GWT including this change in production sometimes results in a NPE. See comment in https://github.com/gwtproject/gwt/issues/9931#issuecomment-2283421092

Issue #9939, opened 2024-03-19
SuperDevListener is not correctly build includeJsInteropExports/excludeJsInteropExports args


MCMicS:
**GWT version: 2.11.0** 
**Browser (with version):** 
**Operating System:** 

---

##### Description
includeJsInteropExports/excludeJsInteropExports arguments cannot be used when launching application with com.google.gwt.dev.DevMode. It comes because The [RegexFiler](https://github.com/gwtproject/gwt/blob/d77e85c2a80d086e387d2413932499831c1718ea/dev/core/src/com/google/gwt/util/regexfilter/RegexFilter.java#L81) add an `+`or `-` and in [SuperDevListener](https://github.com/gwtproject/gwt/blob/d77e85c2a80d086e387d2413932499831c1718ea/dev/core/src/com/google/gwt/dev/shell/SuperDevListener.java#L177) the additional `+`will not be removed which leads to an error because +/- is not allowed there.
This happen for example in Intellij with GWT Run Configration because it uses the DevMode

Code with issue:
```
    for (String regex : options.getJsInteropExportFilter().getValues()) {
      if (regex.startsWith("-")) {
        args.add("-excludeJsInteropExports");
        args.add(regex.substring(1));
      } else {
        args.add("-includeJsInteropExports");
        args.add(regex);
      }
    }
```

Possible solution:
```
    for (String regex : options.getJsInteropExportFilter().getValues()) {
      if (regex.startsWith("-")) {
        args.add("-excludeJsInteropExports");
        args.add(regex.substring(1));
      } else {
        args.add("-includeJsInteropExports");
        args.add(regex.startsWith("+") ? regex.substring(1) : regex);
      }
    }
```
IC will test it the days and can create a Pull Request then if the solution sounds ok.

##### Steps to reproduce

1. Download GWT Zip (like gwt-2.11.0.zip) and extract it
2. Running following command
```shell
java -cp gwt-dev.jar;gwt-user.jar;samples/Hello/src com.google.gwt.dev.DevMode -logdir logs -startupUrl Hello.html -war samples/Hello/war -generateJsInteropExports -includeJsInteropExports com.google.* com.google.gwt.sample.hello.Hello
```
3. The DevMode will not start and shows that following args are transferred to CodeServer (found in logs/main.log):
```
[...]

PR #9940, opened 2024-03-19
remove first character for delegating args to CodeServer

MCMicS:
fix for #9939


niloc132 (approved):
LGTM, and I don't see another clear way to achieve this. The code as it stands doesn't appear very testable, so I'm leaning to merge as is, with a round of manual testing of the fix (maybe added to the release test plan too).

MCMicS:
I like the solution not so because I check if it starts with a `+`. Maybe I can remove it always but I not know all possible values at this point. So this is IMO a safe solution

I can try to write a test to check that the generated `codeServerArgs` are correct after call the constructor. I can do this the days.

niloc132 (approved):

Issue #9759, opened 2022-06-27
Using collapse-all-properties causes StackTraceDeobfuscator to not work anymore


jnehlmeier:
**GWT version:** HEAD
**Browser (with version):** all
**Operating System:** all

---

##### Description

When using `<collapse-all-properties/>`, client exceptions send back to the server for deobfuscation do not have any line and column number information anymore. In that case symbol maps and source maps based deobfuscation will not work.

Not sure if it is a regression since I never tried using `<collapse-all-properties/>` previously.

##### Known workarounds

Do not use `<collapse-all-properties/>`

PR #9941, opened 2024-03-19
Undo the disable of source maps for non-chrome browsers

vegegoku:
Nowadays, all modern browsers support sourcemaps by default, so there is no need to have exception configuration anymore.

### DIFF
--- a/user/src/com/google/gwt/core/CoreWithUserAgent.gwt.xml
+++ b/user/src/com/google/gwt/core/CoreWithUserAgent.gwt.xml
@@ -16,14 +16,6 @@
 <module>
   <inherits name="com.google.gwt.core.Core"/>
 
-  <!-- Disable source maps for non-Chrome browsers -->
-  <!-- TODO(goktug): enable source maps for non-Chrome browsers -->
-  <set-property name="compiler.useSourceMaps" value="false">
-    <none> 
-      <when-property-is name="user.agent" value="safari"/>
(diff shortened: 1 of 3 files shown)


MCMicS (approved):

niloc132 (commented):
Worth testing if this is what was breaking https://github.com/gwtproject/gwt/issues/9759 - since this claims that only chrome/safari support sourcemaps (clearly untrue, hasn't been true for a very long time), this might prevent the symbolmaplinker from correctly producing output when collapsed?

user/src/com/google/gwt/core/CoreWithUserAgent.gwt.xml
```diff
@@ -12,22 +12,13 @@
 <!-- implied. License for the specific language governing permissions and   -->
 <!-- limitations under the License.                                         -->
 
-<!-- Deferred binding rules for core classes based on user agent. -->
+<!-- Deferred binding rules for core classes. -->
```
niloc132, line 15: If this is no longer based on user agent, can we move the remaining replace-with to StackTrace.gwt.xml where it probably belongs?

That would leave this file empty, and we can let users remove it from their own projects, eventually removing it here.

niloc132 (approved):

niloc132:
Update description to list any issues fixed by this, and summarize changes, then we can merge?

PR #9952, opened 2024-04-24
Fix SDM (dev_mode_on.js) to work with strong CSP

### LINKED ISSUE #9951: SDM (Super Dev Mode) does not work with a strong CSP (Content Security Policy)
SDM (Super Dev Mode) does not work with a strong CSP (Content Security Policy)

**GWT version: 2.11.0** 

##### Description
SDM does not work with a CSP, that does not allow "unsafe-inline", because the "dev_mode_on.js" contains some inline javascript call:

`result.setAttribute('href', 'javascript:' + encodeURIComponent(javascript));
`
##### Steps to reproduce
Enable CSP w/o "unsafe-inline" and try to use SDM.

##### Known workarounds
Disable CSP or allow "unsafe-inline".

lgemeinhardt:
This PR converts the inline javascript call (within dev_mode_on.js in SDM) to a javascript function, to allow CSP (Content Security Policy) to avoid "unsafe-inline".

Fixes #9951

### DIFF
--- a/dev/codeserver/java/com/google/gwt/dev/codeserver/dev_mode_on.js
+++ b/dev/codeserver/java/com/google/gwt/dev/codeserver/dev_mode_on.js
@@ -122,19 +122,22 @@
     result.style.background = '#ddd';
     result.style.border = '2px outset #ddd';
     result.style.padding = '3pt';
-    result.setAttribute('href', 'javascript:' + encodeURIComponent(javascript));
+    result.onclick = javascript;
     result.title = 'Tip: drag this button to the bookmark bar';
     return result;
   }
 
   function makeCompileBookmarklet(codeserver_url, module_name) {
     var bookmarklets_js = codeserver_url + 'dev_mode_on.js';
-    var javascript = '{ window.__gwt_bookmarklet_params = {'
-        + 'server_url:\'' + codeserver_url + '\','
-        + 'module_name:\'' + module_name + '\'};'
-        + ' var s = document.createElement(\'script\');'
-        + ' s.src = \'' + bookmarklets_js + '\';'
-        + ' void(document.getElementsByTagName(\'head\')[0].appendChild(s));}';
+    var javascript = function() { 
+	  window.__gwt_bookmarklet_params = {
+        server_url: codeserver_url,
+        module_name: module_name
+	  };
+       var s = document.createElement('script');
+       s.src = bookmarklets_js;
+       void(document.getElementsByTagName('head')[0].appendChild(s));
+	};
     return makeBookmarklet('Compile', javascript);
   }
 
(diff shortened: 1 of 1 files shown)


dev/codeserver/java/com/google/gwt/dev/codeserver/dev_mode_on.js
```diff
@@ -122,19 +122,22 @@
...
-        + ' void(document.getElementsByTagName(\'head\')[0].appendChild(s));}';
+    var javascript = function() { 
+	  window.__gwt_bookmarklet_params = {
+        server_url: codeserver_url,
+        module_name: module_name
+	  };
+       var s = document.createElement('script');
+       s.src = bookmarklets_js;
+       void(document.getElementsByTagName('head')[0].appendChild(s));
+	};
```
niloc132, line 140: Indentation is kind of all over the place here, can you make it consistent?

dev/codeserver/java/com/google/gwt/dev/codeserver/dev_mode_on.js
```diff
@@ -122,19 +122,22 @@
     result.style.background = '#ddd';
     result.style.border = '2px outset #ddd';
     result.style.padding = '3pt';
-    result.setAttribute('href', 'javascript:' + encodeURIComponent(javascript));
+    result.onclick = javascript;
     result.title = 'Tip: drag this button to the bookmark bar';
```
niloc132, line 126: I don't think this will function as a bookmarklet at all if the href is replaced by a lambda - lambdas aren't copied when the bookmarklet URL's contents are... Is there a way we can detect that CSP is in use here and just don't offer the bookmarklet button?

dev/codeserver/java/com/google/gwt/dev/codeserver/dev_mode_on.js
lgemeinhardt: Fixed now

dev/codeserver/java/com/google/gwt/dev/codeserver/dev_mode_on.js
lgemeinhardt: You are right. The button has a dual function (direct clicking or use as bookmarklet - I never used the second one, because it's server specific). Now I added both (CSP and bookmarklet) cases and it works for both cases (even if the browser shows a CSP issue). I could not think of a better way to separate the two cases better, while using only one button.

dev/codeserver/java/com/google/gwt/dev/codeserver/dev_mode_on.js
```diff
@@ -115,14 +115,15 @@
     return dialog;
   }
 
-  function makeBookmarklet(name, javascript) {
+  function makeBookmarklet(name, javascript, javascriptFunction) {
```
zbynek, line 118: Instead of passing `javascript` as a separate argument causing duplication we could compute it
```
var javascript = '('+(javascriptFunction+"").replace(/\s+/g,' ')+')()';
```
(the replace is just to keep it shorter)

lgemeinhardt: @zbynek Thanks for this proposal. The missing part would be to replace the "placeholder" variables ("codeserver_url", "module_name" and "bookmarklets_js") later on with real (string) values. Not sure, if this would be more readable at the end.

niloc132 (approved):

zbynek (approved):

PR #9953, opened 2024-05-05
Add Java 9 BigInteger constructors

### LINKED ISSUE #9901: Java 9 BigInteger constructors with offset/length params missing
In Java 9, two new constructors were added to BigInteger:

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/math/BigInteger.html#%3Cinit%3E(byte%5B%5D,int,int)

vegegoku:
Emulate the `BigInteger(byte[] bytes, int offset, int length)` and `BigInteger​(int signum, byte[] magnitude, int offset, int length)` constructors that were added in java 9.

Fixes #9901

### DIFF
--- a/user/super/com/google/gwt/emul/java/math/BigInteger.java
+++ b/user/super/com/google/gwt/emul/java/math/BigInteger.java
@@ -280,7 +280,7 @@ public class BigInteger extends Number implements Comparable<BigInteger>,
       // math.12=Zero length BigInteger
       throw new NumberFormatException("Zero length BigInteger"); //$NON-NLS-1$
     }
-    if (length < 0 || offset < 0 || val.length < 0 || length > val.length - offset) {
+    if (length < 0 || offset < 0 || length > val.length - offset) {
       throw new IndexOutOfBoundsException("Range check failed: offset=" + offset + ", length="
           + length + ", val.length=" + val.length);
     }
@@ -337,7 +337,7 @@ public class BigInteger extends Number implements Comparable<BigInteger>,
   public BigInteger(int signum, byte[] magnitude, int offset, int length) {
     checkNotNull(magnitude);
 
-    if (length < 0 || offset < 0 || magnitude.length < 0 || length > magnitude.length - offset) {
+    if (length < 0 || offset < 0 || length > magnitude.length - offset) {
       throw new IndexOutOfBoundsException("Range check failed: offset=" + offset + ", length="
           + length + ", val.length=" + magnitude.length);
     }
--- a/user/super/com/google/gwt/emul/java/math/BigInteger.java
+++ b/user/super/com/google/gwt/emul/java/math/BigInteger.java
@@ -260,16 +260,40 @@ public class BigInteger extends Number implements Comparable<BigInteger>,
    * @throws NumberFormatException if the length of {@code val} is zero.
    */
   public BigInteger(byte[] val) {
+    this(val, 0, val.length);
+  }
+
+  /**
+   * Constructs a new {@code BigInteger} from the given two's complement
+   * representation. The most significant byte is the entry at index 0. The most
+   * significant bit of this entry determines the sign of the new {@code
+   * BigInteger} instance. The given array must not be empty.
+   *
+   * @param val two's complement representation of the new {@code BigInteger}.
+   * @throws NullPointerException if {@code val == null}.
+   * @throws NumberFormatException if the length of {@code val} is zero.
+   */
+  public BigInteger(byte[] val, int offset, int length) {
     if (val.length == 0) {
       // math.12=Zero length BigInteger
       throw new NumberFormatException("Zero length BigInteger"); //$NON-NLS-1$
     }
-    if (val[0] < 0) {
+    if (length < 0 || offset < 0 || val.length < 0 || length > val.length - offset) {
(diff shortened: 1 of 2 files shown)


user/super/com/google/gwt/emul/java/math/BigInteger.java
```diff
@@ -260,16 +260,40 @@ private static void setFromString(BigInteger bi, String val, int radix) {
...
     if (val.length == 0) {
       // math.12=Zero length BigInteger
       throw new NumberFormatException("Zero length BigInteger"); //$NON-NLS-1$
     }
-    if (val[0] < 0) {
+    if (length < 0 || offset < 0 || val.length < 0 || length > val.length - offset) {
+      throw new IndexOutOfBoundsException("Range check failed: offset=" + offset + ", length="
+          + length + ", val.length=" + val.length);
+    }
+    if (length == 0) {
```
niloc132, line 285: It surprises me that this is supported - that's effectively the same thing as passing an empty byte array, but sure enough, it makes a difference:
```
jshell> new BigInteger(new byte[]{127, 127, 127, 127}, 0, 0)
$6 ==> 0

jshell> new BigInteger(new byte[]{}, 0, 0)
|  Exception java.lang.NumberFormatException: Zero length BigInteger
|        at BigInteger.<init> (BigInteger.java:315)
|        at (#7:1)
```

user/test/com/google/gwt/emultest/java/math/BigIntegerConstructorsTest.java
```diff
@@ -605,6 +606,62 @@ public void testConstructorSignBytesZeroNull3() {
     assertEquals("incorrect sign", 0, aNumber.signum());
   }
 
+  @Test
```
niloc132, line 609: This annotation isn't required/necessary, all tests are Junit3 :(.

Could you add tests that confirm working behavior, in addition to these negative test cases? It is probably enough to take a few tests above that use the old constructor and just pad them out with other numbers (0 or 127 so we have all bits and no bits set), and set offset/length to confirm that only the correct values are used?

user/test/com/google/gwt/emultest/java/math/BigIntegerConstructorsTest.java
```diff
@@ -605,6 +606,62 @@ public void testConstructorSignBytesZeroNull3() {
...
+    try {
+      new BigInteger(aByte, 2, 4);
+      fail("NumberFormatException has not been caught");
+    } catch (NumberFormatException e) {
+    }
+  }
+
+  @Test
+  public void testConstructorBytes_Offset_ZeroLength() {
+    byte[] aByte = {0, 0, 12, 34, 56, 78, 0, 0};
```
niloc132, line 621: Just keep in mind that if the implementation was wrong and it was reading the 0th or 1st position, the value would still be zero... might be better to replace the zeros here with something non-zero.

niloc132 (commented):
In more closely looking at the Javadoc for BigInteger, I see there is another new constructor:
https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/math/BigInteger.html#%3Cinit%3E(int,byte%5B%5D,int,int)

There are also other new methods (`sqrt`, `sqrtAndRemainder`), and a new constant TWO. I'll make another issue for these new members, but as long as the other constructor is similar work, it seems worth adding now?

user/super/com/google/gwt/emul/java/math/BigInteger.java
```diff
@@ -289,27 +315,53 @@ public BigInteger(byte[] val) {
...
+   * @param offset the start offset of the binary representation.
+   * @param length the number of bytes to use.
+   * @throws NullPointerException if {@code magnitude == null}.
+   * @throws NumberFormatException if the sign is not one of -1, 0, 1 or if the
+   *           sign is zero and the magnitude contains non-zero entries.
+   */
+  public BigInteger(int signum, byte[] magnitude, int offset, int length) {
     checkNotNull(magnitude);
 
+    if (length < 0 || offset < 0 || magnitude.length < 0 || length > magnitude.length - offset) {
```
niloc132, line 340:
```suggestion
    if (length < 0 || offset < 0 || length > magnitude.length - offset) {
```
I don't think we need the magnitude length check?

user/super/com/google/gwt/emul/java/math/BigInteger.java
```diff
@@ -260,16 +260,42 @@ private static void setFromString(BigInteger bi, String val, int radix) {
...
+   * @throws NullPointerException if {@code val == null}.
+   * @throws NumberFormatException if the length of {@code val} is zero.
+   */
+  public BigInteger(byte[] val, int offset, int length) {
     if (val.length == 0) {
       // math.12=Zero length BigInteger
       throw new NumberFormatException("Zero length BigInteger"); //$NON-NLS-1$
     }
-    if (val[0] < 0) {
+    if (length < 0 || offset < 0 || val.length < 0 || length > val.length - offset) {
```
niloc132, line 283:
```suggestion
    if (length < 0 || offset < 0 || length > val.length - offset) {
```
We shouldn't ever need the array length check

niloc132 (commented):
Full build passed: https://github.com/vegegoku/gwt/actions/runs/9192280404

niloc132 (approved):

PR #9955, opened 2024-05-11
Update assumptions in LoggingRPCSuite and fix missing class name

niloc132:
CollectorChrome was renamed about 10 years ago, and sourcemaps are well supported now, so this test can be adjusted to always handle the optimized case.

Fixes the recent build failure.


niloc132:
Full CI run after discovering https://github.com/gwtproject/gwt/issues/9956 and adjusting the test to go back to 'native' stack mode (for all browsers, not just chrome).

https://github.com/niloc132/gwt/actions/runs/9048067967

vegegoku (approved):

Issue #9675, opened 2019-11-28
Interfaces marked as @JsFunction have property length set to 0


vasvir:
The issue is in current git of gwt and it is not browser related. The problematic code is at https://github.com/gwtproject/gwt/blob/master/dev/core/super/com/google/gwt/dev/jjs/intrinsic/com/google/gwt/lang/Runtime.java#L162

Looks like that some javascript libraries are checking on the argument length of the callbacks that are passed to them.

So for example D3 bisector can take a function with one or two arguments. If it is one then it is an accessor if it is two then it is a comparator.

##### Steps to reproduce
`

    @JsFunction
    public static interface SomeFunction {
        public double call(double x);
    }

    @Override
    public void onModuleLoad() {
        final SomeFunction y = new SomeFunction() {
            @Override
            public double call(double x) {
                return 2 * x;
            }
        };
        DomGlobal.console.log(y);
        return;    
    }
`
Note in the attached image that the reported length of the generated function is 0 instead of 1.

![image](https://user-images.githubusercontent.com/4067640/69808493-46733500-11f0-11ea-8904-18cf959d66c4.png)

##### Known workarounds
There is no known workaround.

##### Links to further discussions

This issue has been reported at https://groups.google.com/forum/#!topic/Google-Web-Toolkit/-qaL4dzZm_E

vasvir:
There is a fiddle with a bit more experimentation and benchmarks at https://jsfiddle.net/vasvir/forwtzqd/

- makeLambdaFunction0 is the original method used in the GWT project
- makeLambdaFunction1 is the method (hardwired switch statement) that is implemented in the submitted patch. It is based on https://stackoverflow.com/questions/7316688/how-to-programmatically-set-the-length-of-a-function and it has the same speed as the original code.
- makeLambdaFunction2 is a fully generic, argument count agnostic, based on eval, implementation. It is based on  https://stackoverflow.com/questions/51953812/is-it-possible-to-change-a-lexical-scope-programmatically-in-javascript and it is around 600 times slower than the original code.
- makeLambdaFunction3 is an improvement of makeLambdaFunction2. It eliminates one eval function out of two and reduces the overhead to 300 times compared to the original code. Still however it is not considered for patch submission.

The current patch implements a hardwired switch statement with 10 branches. So an argument count of up to10 is supported with this implementation. If the function has more than 10 arguments then the argument count is set to 10.

[...]

PR #9957, opened 2024-05-15
Fix native JsMethods with varargs called from Java varargs methods

### LINKED ISSUE #9932: Regression in jsinterop varargs causing internal compiler error
**GWT version:** HEAD-SNAPSHOT
**Operating System:** linux

---

##### Description
It looks like the JDT update might have changed how we build JMethodCall instances, and broken some jsinterop vararg usages. I'm still working out exactly what is wrong, this ticket is a work in progress, but it looks like the `JMethodCall` now has multiple arguments where previously it had an array of arguments. 

##### Steps to reproduce
The sample line of code is
```
DomGlobal.console.error("Request", req, " failed with reason ", reason);`
```
This should be compiled to (in GWT 2.11.0) something like
```javascript
$wnd.goog.global.console.error('Request ',b,Mkc,d);
```
but in HEAD-SNAPSHOT we see an exception instead:
```
com.google.gwt.dev.jjs.InternalCompilerException: Unexpected error during visit.
        at com.google.gwt.dev.js.ast.JsVisitor.translateException(JsVisitor.java:482)
        at com.google.gwt.dev.js.ast.JsVisitor.doTraverse(JsVisitor.java:469)
        at com.google.gwt.dev.js.ast.JsVisitor.doAccept(JsVisitor.java:444)
        at com.google.gwt.dev.js.ast.JsVisitor.accept(JsVisitor.java:109)
        at com.google.gwt.dev.js.ast.JsNameRef.traverse(JsNameRef.java:115)
        at com.google.gwt.dev.js.ast.JsVisitor.doTraverse(JsVisitor.java:467)
        at com.google.gwt.dev.js.ast.JsVisitor.doAccept(JsVisitor.java:444)
        at com.google.gwt.dev.js.ast.JsVisitor.accept(JsVisitor.java:109)
        at com.google.gwt.dev.js.ast.JsNameRef.traverse(JsNameRef.java:115)
        at com.google.gwt.dev.js.ast.JsVisitor.doTraverse(JsVisitor.java:467)
        at com.google.gwt.dev.js.ast.JsVisitor.doAcceptList(JsVisitor.java:450)
        at com.google.gwt.dev.js.ast.JsVisitor.acceptList(JsVisitor.java:113)
        at com.google.gwt.dev.js.ast.JsInvocation.traverse(JsInvocation.java:89)
        at com.google.gwt.dev.js.ast.JsVisitor.doTraverse(JsVisitor.java:467)
        at com.google.gwt.dev.js.ast.JsVisitor.doAccept(JsVisitor.java:444)
        at com.google.gwt.dev.js.ast.JsVisitor.accept(JsVisitor.java:109)
        at com.google.gwt.dev.js.ast.JsConditional.traverse(JsConditional.java:85)
        at com.google.gwt.dev.js.ast.JsVisitor.doTraverse(JsVisitor.java:467)
        at com.google.gwt.dev.js.ast.JsVisitor.doAccept(JsVisitor.java:444)
        at com.google.gwt.dev.js.ast.JsVisitor.accept(JsVisitor.java:109)
        at com.google.gwt.dev.js.ast.JsExprStmt.traverse(JsExprStmt.java:42)
        at com.google.gwt.dev.js.ast.JsVisitor.doTraverse(JsVisitor.java:467)
        at com.google.gwt.dev.js.ast.JsVisitor.doAcceptWithInsertRemove(JsVisitor.java:461)
        at com.google.gwt.dev.js.ast.JsVisitor.acceptWithInsertRemove(JsVisitor.java:121)
        at com.google.gwt.dev.js.ast.JsBlock.traverse(JsBlock.java:53)
        at com.google.gwt.dev.js.ast.JsVisitor.doTraverse(JsVisitor.java:467)
        at com.google.gwt.dev.js.ast.JsVisitor.doAccept(JsVisitor.java:444)
        at com.google.gwt.dev.js.ast.JsVisitor.accept(JsVisitor.java:109)
        at com.google.gwt.dev.js.ast.JsFunction.traverse(JsFunction.java:181)
        at com.google.gwt.dev.js.ast.JsVisitor.doTraverse(JsVisitor.java:467)
        at com.google.gwt.dev.js.ast.JsVisitor.doAccept(JsVisitor.java:444)
        at com.google.gwt.dev.js.ast.JsVisitor.accept(JsVisitor.java:109)
        at com.google.gwt.dev.js.ast.JsExprStmt.traverse(JsExprStmt.java:42)
        at com.google.gwt.dev.js.ast.JsVisitor.doTraverse(JsVisitor.java:467)
        at com.google.gwt.dev.js.ast.JsVisitor.doAcceptWithInsertRemove(JsVisitor.java:461)
        at com.google.gwt.dev.js.ast.JsVisitor.acceptWithInsertRemove(JsVisitor.java:121)
        at com.google.gwt.dev.js.ast.JsBlock.traverse(JsBlock.java:53)
        at com.google.gwt.dev.js.ast.JsVisitor.doTraverse(JsVisitor.java:467)
        at com.google.gwt.dev.js.ast.JsVisitor.doAccept(JsVisitor.java:444)
        at com.google.gwt.dev.js.ast.JsVisitor.accept(JsVisitor.java:109)
        at com.google.gwt.dev.js.ast.JsProgramFragment.traverse(JsProgramFragment.java:41)
        at com.google.gwt.dev.js.ast.JsVisitor.doTraverse(JsVisitor.java:467)
        at com.google.gwt.dev.js.ast.JsVisitor.doAccept(JsVisitor.java:444)
        at com.google.gwt.dev.js.ast.JsVisitor.accept(JsVisitor.java:109)
        at com.google.gwt.dev.js.ast.JsProgram.traverse(JsProgram.java:120)
        at com.google.gwt.dev.js.ast.JsVisitor.doTraverse(JsVisitor.java:467)
        at com.google.gwt.dev.js.ast.JsVisitor.doAccept(JsVisitor.java:444)
        at com.google.gwt.dev.js.ast.JsVisitor.accept(JsVisitor.java:109)
        at com.google.gwt.dev.js.JsSymbolResolver.execImpl(JsSymbolResolver.java:69)
        at com.google.gwt.dev.js.JsSymbolResolver.exec(JsSymbolResolver.java:58)
        at com.google.gwt.dev.jjs.JavaToJavaScriptCompiler.compilePermutation(JavaToJavaScriptCompiler.java:391)
        at com.google.gwt.dev.jjs.JavaToJavaScriptCompiler.compilePermutation(JavaToJavaScriptCompiler.java:272)
        at com.google.gwt.dev.CompilePerms.compile(CompilePerms.java:198)
        at com.google.gwt.dev.ThreadedPermutationWorkerFactory$ThreadedPermutationWorker.compile(ThreadedPermutationWorkerFactory.java:50)
        at com.google.gwt.dev.PermutationWorkerFactory$Manager$WorkerThread.run(PermutationWorkerFactory.java:74)
        at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: java.lang.NullPointerException
        at com.google.gwt.thirdparty.guava.common.base.Preconditions.checkNotNull(Preconditions.java:212)
        at com.google.gwt.thirdparty.guava.common.collect.MapMakerInternalMap.putIfAbsent(MapMakerInternalMap.java:3521)
        at com.google.gwt.thirdparty.guava.common.collect.Interners$WeakInterner.intern(Interners.java:88)
        at com.google.gwt.dev.js.ast.JsNormalScope.doCreateName(JsNormalScope.java:49)
        at com.google.gwt.dev.js.ast.JsScope.declareName(JsScope.java:60)
        at com.google.gwt.dev.js.ast.JsScope.declareUnobfuscatableName(JsScope.java:91)
        at com.google.gwt.dev.js.JsSymbolResolver$JsResolveSymbolsVisitor.resolveQualifiedName(JsSymbolResolver.java:38)
        at com.google.gwt.dev.js.JsAbstractSymbolResolver.endVisit(JsAbstractSymbolResolver.java:52)
        at com.google.gwt.dev.js.ast.JsNameRef.traverse(JsNameRef.java:118)
        at com.google.gwt.dev.js.ast.JsVisitor.doTraverse(JsVisitor.java:467)
        ... 54 more
         [ERROR] at Callbacks.java(64): <source info not available>
         [ERROR] at Callbacks.java(64): <source info not available>
         [ERROR] at Callbacks.java(64): <source info not available>
         [ERROR] at Callbacks.java(64): <source info not available>
         [ERROR] at Callbacks.java(61): <source info not available>
         [ERROR] at Callbacks.java(61): <source info not available>
         [ERROR] at Callbacks.java(42): <source info not available>
         [ERROR] at Callbacks.java(42): <source info not available>
         [ERROR] at Callbacks.java(42): <source info not available>
         [ERROR] at com.google.gwt.dev.js.ast.JsProgram(0): var _, prototypesByTypeId, initFnList, permutationId = -1;
function getPermutationId(){
  return permutationId;
}

function $$init(){
}

function setGwtProperty(propertyName, propertyValue){
  if (typeof window === 'object' && typeof window['$gwt'] === 'object') {
    window['$gwt'][propertyName] = propertyValue;
  }
}

[...]

         [ERROR] at com.google.gwt.dev.js.ast.JsProgram(0): <JsProgramFragment>
         [ERROR] at com.google.gwt.dev.js.ast.JsProgram(0): <JsProgram>
      [ERROR] Unrecoverable exception, shutting down
com.google.gwt.core.ext.UnableToCompleteException: (see previous log entries)
        at com.google.gwt.dev.javac.CompilationProblemReporter.logAndTranslateException(CompilationProblemReporter.java:106)
        at com.google.gwt.dev.jjs.JavaToJavaScriptCompiler.compilePermutation(JavaToJavaScriptCompiler.java:459)
        at com.google.gwt.dev.jjs.JavaToJavaScriptCompiler.compilePermutation(JavaToJavaScriptCompiler.java:272)
        at com.google.gwt.dev.CompilePerms.compile(CompilePerms.java:198)
        at com.google.gwt.dev.ThreadedPermutationWorkerFactory$ThreadedPermutationWorker.compile(ThreadedPermutationWorkerFactory.java:50)
        at com.google.gwt.dev.PermutationWorkerFactory$Manager$WorkerThread.run(PermutationWorkerFactory.java:74)
        at java.base/java.lang.Thread.run(Thread.java:1583)
      [ERROR] Not all permutation were compiled , completed (0/1)
```
That exception seems to come from the fact that the JsSafeCloner only considers the `name` property of a `JsNameRef` and not the `ident`
https://github.com/gwtproject/gwt/blob/d77e85c2a80d086e387d2413932499831c1718ea/dev/core/src/com/google/gwt/dev/js/JsSafeCloner.java#L145-L157

However, there are cases where `name` can be null, and `ident` then has the relevant information. Applying this fix gets around that stack trace:
```
-      JsNameRef toReturn = new JsNameRef(x.getSourceInfo(), x.getName());
+      final JsNameRef toReturn;
+      if (x.getName() == null) {
+        toReturn = new JsNameRef(x.getSourceInfo(), x.getIdent(), x.getQualifier());
+      } else {
+        toReturn = new JsNameRef(x.getSourceInfo(), x.getName());
+      }
```
This at least leads us to emit correct code:
```javascript
$wnd.goog.global.console.error.apply($wnd.goog.global.console, ['Request ',b,Mkc].concat(d));
```
however, this is unnecessarily using the `USING_APPLY_FOR_VARARGS_ARRAY` call style - it appears that the first three vararg elements were detected as being earlier parameters (which do not exist), rather than just components of the vararg parameter. Seeming to confirm that, when debugging 2.11.0, `lastArgument` ends up being a `JsArrayLiteral` with all four arguments, so that `DIRECT` call style can be used entirely with non-vararg arguments.
https://github.com/gwtproject/gwt/blob/d77e85c2a80d086e387d2413932499831c1718ea/dev/core/src/com/google/gwt/dev/js/JsUtils.java#L257-L270
Debugging this same code in HEAD-SNAPSHOT ends up with the first three args in `nonVarargArguments`, and `lastArgument` is a single local reference.

Both issues should be tested, fixed as part of this ticket.

niloc132:
Native jsinterop varargs calls of the form `instance.method(argsArray)` sometimes implemented using `instance.method.apply(instance, argsArray)`, where instance represents some expression.  JsUtils.createApplyInvocation assumes that ImplementJsVarargs has already created a local variable to ensure that no side effects can take place when cloning the "instance", but ImplementJsVarargs doesn't actually check if a side effect was able to happen. 

This fix only adds an additional check, creating a local variable in the case where the instance expression has side effects, and clarifies that a native JsProperty might have sideeffects (as it could call a JS property accessor).

An additional fix is made to JsSafeCloner, to let it successfully clone a qualified reference (e.g. `$wnd.console`) rather than produce an invalid JS AST. This ended up not being strictly required for the fix, but appears to be more correct and will at least generate code rather than a NullPointerException in a later location.

Fixes #9932


niloc132:
Full build https://github.com/niloc132/gwt/actions/runs/9088542441

dev/core/src/com/google/gwt/dev/jjs/ast/JFieldRef.java
```diff
@@ -92,6 +92,9 @@ public boolean hasSideEffects() {
     if (hasClinit()) {
       return true;
     }
+    if (getField().isJsNative()) {
+      return false;
+    }
```
niloc132, line 97: This is just plain backwards - I think I inverted it at some point for testing purposes - but I think I'd prefer to remove it entirely.

The idea was supposed to be as a way to trigger ImplementJsVarargs own "if this has side effects I should only eval it once" in a nicer way than assuming it recognizes all qualified name references (since its failure to do so caused the original bug). However, this ends up having other side effects, like preventing a lot of dead code elimination from working correctly.

The "hasSideEffects" check itself in ImplementJsVarargs isn't wrong, but this particular attempt at a safeguard is. The comment in ImplementJsVarargs should also be corrected.

niloc132:
This patch has a few issues with it - the native property ref issue is one, discussed above, but also that the fix is incomplete. While this does appear to address the long-standing issue (at least since 2.9, probably older), it seems that the recent JDT update changes how varargs are handled in some small way. Perhaps that issue could be handled separately, but they seem somewhat linked - when I understand what is happening there I'll see if it makes sense to land them together or apart.

niloc132:
I think the concerns above are addressed - the native property ref sideeffect has been backed out, and the JDT issue is resolved, though more tests need to be written to confirm. Here's the fully build succeeding with the latest JDT change: https://github.com/niloc132/gwt/actions/runs/9586615514

The fix revolves around the fact that there are subtypes of MethodBinding that didn't previously exist - using `MethodBinding.original()` we can get back the underlying method so that we see that the method is varargs (so may need an array creation synthesized). However, `original()` also discards generic parameterization, so types no longer match, so this must be done somewhat carefully, and more tests are needed to confirm that the behavior and output is correct. At the very least, tests should include:
 * varargs with `Object...`  (as this will accept all objects), tested with `null`, empty, one argument, several arguments, array 
 * repeat with a non-vararg param
 * varargs with type args, where the typearg is unrelated to the vararg param
 * varargs with type args, where the typearg is used to specify the vararg param (like Arrays.asList)

Unfortunately, GWT has effectively no tests for simple Java features, only indirect tests that often expose problems. I assume either that these tests exist but were never ported to open source, or that they were not needed since Google had enough code written using GWT that such problems were discovered on their own. Nevertheless, it makes bugs like this and switch/case hard to hunt down, since simple tests like "are varargs methods called correctly" or "does the correct case get used in a switch" don't seem to exist in their plainest form.

user/test/com/google/gwt/core/interop/JsTypeVarargsTest.java
```diff
@@ -321,4 +322,31 @@ public static int varargJsMethodUninstantiatedVararg(
...
+
+  // Java declaration of globally available instance method that takes varargs
+  @JsType(namespace = JsPackage.GLOBAL)
+  public static class VarArgsQualifiedInstanceMethod {
+    @JsProperty(namespace = JsPackage.GLOBAL)
+    public static VarArgsQualifiedInstanceMethod INSTANCE = new VarArgsQualifiedInstanceMethod();
+    public int getLength(Object... values) {
+      return values.length;
+    }
+  }
```
vegegoku, line 339: new line here.

dev/core/src/com/google/gwt/dev/jjs/impl/ImplementJsVarargs.java
```diff
@@ -462,8 +462,8 @@ public void endVisit(JMethodCall x, Context ctx) {
       // Passed as an array to varargs method will result in an apply call, in which case hoist the
       // qualifier to make sure it is only evaluated once.
       JExpression instance = x.getInstance();
-      if (x.getTarget().needsDynamicDispatch() && !x.isStaticDispatchOnly()
-          && instance != null && !(instance instanceof JVariableRef)) {
+      if (x.getTarget().needsDynamicDispatch() && !x.isStaticDispatchOnly() && instance != null
```
vegegoku, line 465: Since we dropped support for java 7 and now we only support java 8+ wont it make sense to start using newer syntax like Objects.nonNull(...) and Objects.isNull(...)

dev/core/src/com/google/gwt/dev/js/JsSafeCloner.java
```diff
@@ -143,13 +143,17 @@ public void endVisit(JsNameOf x, JsContext ctx) {
      */
     @Override
     public void endVisit(JsNameRef x, JsContext ctx) {
-      if (x.getQualifier() == null && x.getIdent() == "arguments") {
+      if (x.getQualifier() == null && "arguments".equals(x.getIdent())) {
```
vegegoku, line 146: Same here Objects.isNull

dev/core/src/com/google/gwt/dev/jjs/impl/ImplementJsVarargs.java
niloc132: Specifically, we dropped support for _running this test_ on Java 7 - the code may or may not still run, but the test won't pass. We are not yet compiling code to use Java 9+ APIs, but could use Java features like that.

That said, two thoughts:
 * isNull() is great to use as a lambda, but it any better than the != null? It appears to be the same number of chars (assuming you add spaces, and assuming you do a static import), but will add the cost of a method call (until it is inlined by the jit, etc). As I understand it, the purpose is mostly for passing as a method reference...?
  * If I'm not changing the logic (I only moved the null check up a line), I'd rather not change the code to another way of doing the same thing, all else equals.

dev/core/src/com/google/gwt/dev/js/JsSafeCloner.java
niloc132: Objects.equals() probably? I think that would be better than what I have here now, as a replacement for `==`. I've never seen this fail, but it certainly could, and then would break here, which is why I changed it.

vegegoku (approved):

vasvir:
Sorry if this is irrelevant:

Is this patch close to the issue #9675 ? Or is it in a different part of the code?

jnehlmeier:
@vasvir This patch fixes JsMethod varargs methods, e.g. `console.log(Object... o)`, which caused compiler errors. This patch does not change how GWT generates JS code for Java lambdas.

niloc132:
@vasvir I think the issue you linked has an accurate summary here - there isn't a nice way to do this outside of the ones you've presented.

Getting into the details of "what does the Function object say about how you can call it" seem fraught - either we need an extra wrapper per function (Roberto's solution of a new wrapper per func), a huge runtime cost (your option 2 or 3 solution) or we need to be satisfied with only 0-10 args, and never preserving the name of the arguments (your option 1). While 0-10 (or maybe 0-5? 0-50?) might suffice for your own work, it would break JS projects/tools that inspect the names params would still break (angular does this for example for its DI feature).

Another option not included in your list is to get rid of the runtime arg count check and instead have your 11 (or 51, etc) makeLambdaArgumentN functions in Runtime.java, and tweak GenerateJavaScriptAST.constructJsFunctionObject
https://github.com/gwtproject/gwt/blob/a53bb51669ed7dbb894b1ef0d9aa6dd50ff2e126/dev/core/src/com/google/gwt/dev/jjs/impl/GenerateJavaScriptAST.java#L1010-L1020
to instead dispatch based on arg count, so the work is done at compile time.

In theory this could be an optional feature this way, and the 99% of apps that have never wanted this don't need to be impacted by this case.

Issue #9671, opened 2019-10-10
Support Mouse wheel in firefox from version 69


eTouchard:
**GWT version:** 2.8.2
**Browser (with version):** firefox 69.0.2
**Operating System:** All

---

##### Description

Mouse wheel is no longer taken into account in firefox

##### Steps to reproduce

Create an MouseWheelHandler and add it to a widget using addDomHandler(handler, MouseDownEvent.getType());

when turning the wheel the handler is not called

##### Known Correction
The dom event now is wheel and not DOMMouseScroll
in MouseWheelEvent.java add the following line in the static part

 new Type<MouseWheelHandler>("wheel", new MouseWheelEvent());

rdeangelis83:
`com.google.gwt.user.client.ui.Widget#addDomHandler(handler, MouseWheelEvent.getType())` works correctly also in Firefox.
Can you provide simple example project how to reproduce this issue?

zbynek:
Technically the wheel events fire, but implementing a handler that checks if it's scrolling up or down is impossible, because https://github.com/gwtproject/gwt/blob/ebc32a0364d64ad7531e3fcfa51362f95f2398d4/user/src/com/google/gwt/dom/client/DOMImplStandardBase.java#L205 rounds the delta to zero for many input devices and if you try to circumvent that by extracting wheelDelta directly, you run into problems with GWT using two different event types
* https://developer.mozilla.org/en-US/docs/Web/API/Element/DOMMouseScroll_event for Gecko
* https://developer.mozilla.org/en-US/docs/Web/API/Element/mousewheel_event for Blink
Both of them are deprecated.

Issue #8806, opened 2015-06-10
onMouseWheel not triggered on IE11


dankurka:
Originally reported on Google Code with ID 8845
```
Found in GWT Release (e.g. 2.4.0, 2.5.1, trunk):

2.6.1.

Encountered on OS / Browser (e.g. WinXP, IE9, FF10):

Windows 7, Internet Explorer 11. Works on firefox, safari, chrome on Linux, Windows.

Detailed description (please be as specific as possible):

Registering a mouse wheel event on an HTML element does not trigger an event on Internet
Explorer 11. It does work, though, on all other browsers, and it also works for Internet
Explorer 10. This is probably related to the fact that IE11 "fakes" its identity to
be a Mozilla type of browser.

Shortest code snippet which demonstrates issue (please indicate where
actual result differs from expected result):

Register a mouse wheel handler on a Canvas.

Workaround if you have one:

Insert
<meta http-equiv="X-UA-Compatible" content="IE=10" />
into the web page. This makes the mouse wheel work - but breaks other things.

Links to relevant GWT Developer Forum posts:

There seem to be a related defect in the google maps API, probably due to the same
issue:

http://code.google.com/p/gmaps-api-issues/issues/detail?id=5944

Note that this is probably because IE uses onWheel() instead of onMouseWheel() to register
events?

Link to patch posted at http://gwt-review.googlesource.com

No patch posted. However, you may find a demo for reproduction at

https://nflexpvipecs.rus.uni-stuttgart.de/vip/ViPLab.html


```

Reported by `trichter@accusoft.com` on 2014-08-03 19:31:18

stehlifn:
this exists under gwt 2.8 - no mouse wheel when using
- Internet Explorer Version: 11.0.9600.18500
- Internet Explorer Update-Version: 11.0.36

tbroyer:
I can't reproduce. Using the project generated by webAppCreator, I just added:
```java
RootPanel.get().addDomHandler(e -> nameField.setValue("" + e.getDeltaY()), MouseWheelEvent.getType());
```
and when I "mousewheel" above the document, the text field's value is changed, so the event works. (It's changed to "0" though, whereas in Edge it's changed to "3" or "-3" depending on the scroll direction; so `getDeltaY` seems to be broken; also note that you need to "mousewheel" above the _document_, which may not cover the entire viewport).

IE Version: 11.0.10240.16384

PR #9971, opened 2024-06-19
Use wheel event type instead of deprected mousewheel

### LINKED ISSUE #8009: Support standardized DOM WheelEvent
Originally reported on Google Code with ID 8012

```
Currently GWT 2.5 supports only old mousewheel and DOMMouseScroll events and there is
a lack of supporting standardized WheelEvent:

https://developer.mozilla.org/en-US/docs/Mozilla_event_reference/wheel
http://www.w3.org/TR/DOM-Level-3-Events/#event-type-wheel

```

Reported by `a.korzhevskiy` on 2013-02-14 12:17:27

zbynek:
Fixes #8009
Fixes the velocity detection problem mentioned in #9671 and  #8806 

This is a bit more direct than https://gwt-review.googlesource.com/c/gwt/+/3171 as it removes the deprecated events where possible.

Teste in FF and Chrome using a mouse.

### DIFF
--- a/user/src/com/google/gwt/dom/client/DOMImpl.java
+++ b/user/src/com/google/gwt/dom/client/DOMImpl.java
@@ -139,7 +139,7 @@ abstract class DOMImpl {
   }-*/;
 
   public int eventGetMouseWheelVelocityY(NativeEvent evt) {
-    return evt.getDeltaY() == 0.0 ? 0 : (evt.getDeltaY() > 0 ? 1 : -1);
+    return (int) Math.signum(evt.getDeltaY());
   }
 
   public abstract EventTarget eventGetRelatedTarget(NativeEvent nativeEvent);
--- a/user/src/com/google/gwt/dom/client/DOMImpl.java
+++ b/user/src/com/google/gwt/dom/client/DOMImpl.java
@@ -138,9 +138,9 @@ abstract class DOMImpl {
     return !!evt.metaKey;
   }-*/;
 
-  public native int eventGetMouseWheelVelocityY(NativeEvent evt) /*-{
-    return Math.sign(evt.deltaY);
-  }-*/;
+  public int eventGetMouseWheelVelocityY(NativeEvent evt) {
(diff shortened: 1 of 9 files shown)


zbynek:
@niloc132 thoughts on including this in 2.12? I'll try to do some testing asap to remove the draft status.

niloc132:
Sorry, I wasn't taking a look yet until the draft marker was removed (or until I was ping'd directly to ask me to look). I'll try to take a glance in the next few days.

user/src/com/google/gwt/user/client/impl/DOMImpl.java
```diff
@@ -102,7 +102,7 @@ public native int eventGetTypeInt(String eventType) /*-{
     case "scroll": return 0x04000;
     case "error": return 0x10000;
     case "mousewheel": return 0x20000;
```
niloc132, line 104: To confirm, the `mousewheel` case is left here in case someone creates their own mouse event type using `MouseEvents.MOUSEWHEEL`?

zbynek: Yes.

user/src/com/google/gwt/dom/client/DOMImpl.java
```diff
@@ -138,7 +138,9 @@ public native boolean eventGetMetaKey(NativeEvent evt) /*-{
     return !!evt.metaKey;
   }-*/;
 
-  public abstract int eventGetMouseWheelVelocityY(NativeEvent evt);
+  public int eventGetMouseWheelVelocityY(NativeEvent evt) {
+    return evt.getDeltaY() == 0.0 ? 0 : (evt.getDeltaY() > 0 ? 1 : -1);
```
niloc132, line 142: That works, but I was thinking `return (int) Math.signum(evt.getDeltaY());` - at worst it'll do the same thing you're doing, at best it'll share the same pair of conditionals with other code using that impl.

This isn't meant to be a criticism, as I might not have thought this through fully, just pointing it out in case there's a path here that makes sense.

niloc132 (approved):

niloc132 (approved):

PR #9975, opened 2024-07-06
String method emulation for Java 11, align Character.isWhitespace with Java 11

### LINKED ISSUE #9973: Whitespace check inconsistency
**GWT version: 2.11** 
**Browser (with version): Chrome** 
**Operating System: N/A** 

---

##### Description

GWT emulation of isWhitespace is listing characters explicitly to avoid discrepancies between Java and JS notion of "whitespace", but there is one inconsistency with Java 8 and one inconsistency with Java 11.
For comparison the table below also shows which characters are recognized as whitespace using `chr.match(/\s/)` in JS (Chrome).

|        range | java 8 | java 11 | GWT | JS  |
|--------------|--------|---------|-----|-----|
|\u0009-\u000D, \u0020, \u1680, \u2000-\u2006, \u2008-\u200A, \u2028, \u2029, \u205F, \u3000  | yes    | yes     | yes | yes |
|\u001C-\u001F | yes    | yes     | yes | no  |
|\u00A0,\u2007,\u202F        | no     | no      | no  | yes |
|\u180E        | yes    | no     | yes | no |
|\uFEFF        | no     | no      | yes | yes |

##### Steps to reproduce
* Call `Character.isWhitespace('\uFEFF')` in GWT and JVM (version 8 or 11), observe different results.
* Call `Character.isWhitespace('\u180E')` in GWT and JVM (version 11, Temurin), observe different results.

##### Known workarounds
Implement own utility method for whitespace detection.

##### Links
* https://www.compart.com/en/unicode/U+FEFF zero width no break space
* https://www.compart.com/en/unicode/U+180E Mongolian vowel separator
* https://stackoverflow.com/a/4731164/2270026 explicit regex for Java 8
* j2objc emulation https://github.com/google/j2objc/blob/add4718cf098c3565f49576a5f642173cac0c884/jre_emul/android/platform/libcore/ojluni/src/main/java/java/lang/Character.java#L7350

zbynek:
Adds emulation for java.lang.String methods added in Java 11, partially completing #9831:
* isBlank
* lines
* repeat
* strip
* stripLeading
* stripTrailing

Fixes #9973
Partial #9831

### DIFF
--- a/user/super/com/google/gwt/emul/java/lang/Character.java
+++ b/user/super/com/google/gwt/emul/java/lang/Character.java
@@ -322,7 +322,7 @@ public final class Character implements Comparable<Character>, Serializable {
       // the Java definition includes separators.
       whitespaceRegex =
           new NativeRegExp(
-              "[\\u1680\\u180E\\u2000-\\u2006\\u2008-\\u200A\\u2028\\u2029\\u205F\\u3000\\uFEFF]"
+              "[\\u1680\\u2000-\\u2006\\u2008-\\u200A\\u2028\\u2029\\u205F\\u3000]"
                   + "|[\\t-\\r ]"
                   + "|[\\x1C-\\x1F]");
     }
--- a/user/super/com/google/gwt/emul/java/lang/String.java
+++ b/user/super/com/google/gwt/emul/java/lang/String.java
@@ -27,7 +27,12 @@ import java.nio.charset.Charset;
 import java.nio.charset.UnsupportedCharsetException;
 import java.util.Comparator;
 import java.util.Locale;
+import java.util.Spliterator;
+import java.util.Spliterators;
 import java.util.StringJoiner;
+import java.util.function.Consumer;
+import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 import javaemul.internal.ArrayHelper;
 import javaemul.internal.Coercions;
 import javaemul.internal.EmulatedCharset;
@@ -755,6 +760,82 @@ public final class String implements Comparable<String>, CharSequence,
     return start > 0 || end < length ? substring(start, end) : this;
   }
 
+  public String strip() {
+    int length = length();
+    int start = getLeadingWhitespaceLength();
+    if (start == length) {
+      return "";
+    }
+    return substring(start, length - getTrailingWhitespaceLength());
+  }
+
+  public String stripLeading() {
+    return substring(getLeadingWhitespaceLength());
+  }
+
+  public String stripTrailing() {
+    return substring(0, length() - getTrailingWhitespaceLength());
+  }
+
+  public boolean isBlank() {
+    return length() == getLeadingWhitespaceLength();
+  }
+
+  public Stream<String> lines() {
+    return StreamSupport.stream(new LinesSpliterator(this), false);
+  }
+
+  public String repeat(int count) {
+    if (count < 0) {
+      throw new IllegalArgumentException("count is negative: " + count);
+    }
+    return asNativeString().repeat(count);
+  }
+
+  private int getLeadingWhitespaceLength() {
+    int length = length();
+    for (int i = 0; i < length; i++) {
+      if (!Character.isWhitespace(charAt(i))) {
+        return i;
+      }
+    }
+    return length;
+  }
+
+  private int getTrailingWhitespaceLength() {
+    int length = length();
+    for (int i = length - 1; i >= 0; i--) {
+      if (!Character.isWhitespace(charAt(i))) {
+        return length - 1 - i;
+      }
+    }
+    return length;
+  }
+
+  private static class LinesSpliterator extends Spliterators.AbstractSpliterator<String> {
+    private int processed = 0;
+    private String content;
+
+    private LinesSpliterator(String content) {
+      super(Long.MAX_VALUE, Spliterator.IMMUTABLE | Spliterator.ORDERED);
+      this.content = content;
+    }
+
+    @Override
+    public boolean tryAdvance(Consumer<? super String> action) {
+      int rPosition = content.indexOf('\r', processed);
+      int nPosition = content.indexOf('\n', processed);
+      int lineEnd = nPosition == -1 ? rPosition : Math.min(nPosition, rPosition);
+      action.accept(lineEnd == -1 ? content.substring(processed)
+              : content.substring(processed, lineEnd));
+      processed = lineEnd + 1;
+      if (nPosition == rPosition + 1 && rPosition != -1) {
+        processed++;
+      }
+      return lineEnd != -1 && processed < content.length();
+    }
+  }
+
   @JsType(isNative = true, name = "String", namespace = "<window>")
   private static class NativeString {
     public static native String fromCharCode(char x);
@@ -771,6 +852,7 @@ public final class String implements Comparable<String>, CharSequence,
     public native String toLocaleUpperCase();
     public native String toLowerCase();
     public native String toUpperCase();
+    public native String repeat(int count);
   }
 
   // CHECKSTYLE_OFF: Utility Methods for unboxed String.
--- a/user/super/com/google/gwt/emul/java/lang/String.java
+++ b/user/super/com/google/gwt/emul/java/lang/String.java
@@ -16,6 +16,7 @@
 
 package java.lang;
 
+import static javaemul.internal.InternalPreconditions.checkArgument;
 import static javaemul.internal.InternalPreconditions.checkCriticalStringBounds;
 import static javaemul.internal.InternalPreconditions.checkNotNull;
 import static javaemul.internal.InternalPreconditions.checkStringBounds;
@@ -782,13 +783,11 @@ public final class String implements Comparable<String>, CharSequence,
   }
 
   public Stream<String> lines() {
-    return StreamSupport.stream(new LinesSpliterator(this), false);
+    return StreamSupport.stream(new LinesSpliterator(), false);
   }
 
   public String repeat(int count) {
-    if (count < 0) {
-      throw new IllegalArgumentException("count is negative: " + count);
-    }
+    checkArgument(count >= 0, "count is negative: " + count);
     return asNativeString().repeat(count);
   }
 
@@ -812,15 +811,13 @@ public final class String implements Comparable<String>, CharSequence,
     return length;
   }
 
-  private static class LinesSpliterator extends Spliterators.AbstractSpliterator<String> {
+  private class LinesSpliterator extends Spliterators.AbstractSpliterator<String> {
     private int nextIndex = 0;
-    private String content;
-    int rPosition = -2;
-    int nPosition = -2;
+    private int rPosition = -1;
+    private int nPosition = -1;
 
-    private LinesSpliterator(String content) {
+    private LinesSpliterator() {
       super(Long.MAX_VALUE, Spliterator.IMMUTABLE | Spliterator.ORDERED);
-      this.content = content;
     }
 
     @Override
@@ -832,17 +829,17 @@ public final class String implements Comparable<String>, CharSequence,
         nPosition = cappedIndexOf('\n');
       }
       int lineEnd = Math.min(nPosition, rPosition);
-      action.accept(content.substring(nextIndex, lineEnd));
+      action.accept(substring(nextIndex, lineEnd));
       nextIndex = lineEnd + 1;
       if (nPosition == rPosition + 1) {
         nextIndex++;
       }
-      return nextIndex < content.length();
+      return nextIndex < length();
     }
 
     private int cappedIndexOf(char c) {
-      int index = content.indexOf(c);
-      return index == -1 ? content.length() : index;
+      int index = indexOf(c);
+      return index == -1 ? length() : index;
     }
   }
 
--- a/user/super/com/google/gwt/emul/java/lang/String.java
+++ b/user/super/com/google/gwt/emul/java/lang/String.java
@@ -822,6 +822,9 @@ public final class String implements Comparable<String>, CharSequence,
 
     @Override
     public boolean tryAdvance(Consumer<? super String> action) {
+      if (isEmpty()) {
+        return false;
+      }
       if (rPosition < nextIndex) {
         rPosition = cappedIndexOf('\r');
       }
@@ -838,7 +841,7 @@ public final class String implements Comparable<String>, CharSequence,
     }
 
     private int cappedIndexOf(char c) {
-      int index = indexOf(c);
+      int index = indexOf(c, nextIndex);
       return index == -1 ? length() : index;
     }
   }
--- a/user/super/com/google/gwt/emul/java/lang/String.java
+++ b/user/super/com/google/gwt/emul/java/lang/String.java
@@ -813,8 +813,10 @@ public final class String implements Comparable<String>, CharSequence,
   }
 
   private static class LinesSpliterator extends Spliterators.AbstractSpliterator<String> {
-    private int processed = 0;
+    private int nextIndex = 0;
     private String content;
+    int rPosition = -2;
+    int nPosition = -2;
 
     private LinesSpliterator(String content) {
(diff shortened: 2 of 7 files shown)


user/super/com/google/gwt/emul/java/lang/String.java
```diff
@@ -755,6 +760,82 @@ public String trim() {
...
+      int rPosition = content.indexOf('\r', processed);
+      int nPosition = content.indexOf('\n', processed);
+      int lineEnd = nPosition == -1 ? rPosition : Math.min(nPosition, rPosition);
+      action.accept(lineEnd == -1 ? content.substring(processed)
+              : content.substring(processed, lineEnd));
+      processed = lineEnd + 1;
+      if (nPosition == rPosition + 1 && rPosition != -1) {
+        processed++;
+      }
+      return lineEnd != -1 && processed < content.length();
```
niloc132, line 835: if you expand the ternary into an if, you can avoid running the other 4 lines of code at all when lineEnd == -1, roughly:
```suggestion
      if (lineEnd == -1) {
        action.accept(content.substring(processed));
        processed = content.length();
        return false;
      }
      action.accept(content.substring(processed, lineEnd));
      processed = lineEnd + 1;
      if (nPosition == rPosition + 1 && rPosition != -1) {
        processed++;
      }
      return processed < content.length();
```

I'm not sure if this makes any meaningful difference, but it was at least a bit easier to read... If we had a loop here it could matter, but all the "hard" work is already done by this point.

---

That said, do we need to do both indexOf()s unconditionally? Can we not instead scan from `processed` to find `'\r'` - if success test next char for `'\n'`, if fail re-scan from `processed`?

user/test/com/google/gwt/emultest/java11/lang/StringTest.java
```diff
@@ -0,0 +1,81 @@
...
+            String noFoo = "foo".repeat(-1);
+            throw new Error("Should fail with negative arg");
+        } catch (IllegalArgumentException ex) {
+            assertEquals("count is negative: -1", ex.getMessage());
+        }
+    }
+
+    public void testLines() {
+        assertEquals(Arrays.asList("a", "b", "c", "d"),
+                "a\rb\nc\r\nd".lines().collect(Collectors.toList()));
```
niloc132, line 73: Worth testing a \n\r case (that is, the opposite of \r\n)? I think you've got this right, but can't be sure just from looking at it. \r\n should be one line separator, but \n\r should be two. Expanding on this \r\n\r and \n\r\n should be two, but \r\n\r\n should be two while \n\r\n\r should be three (\n, \r\n, \r)

user/test/com/google/gwt/emultest/java11/lang/StringTest.java
```diff
@@ -0,0 +1,81 @@
...
+public class StringTest extends EmulTestBase {
+    public void testIsBlank() {
+        assertTrue("".isBlank());
+        assertTrue("  ".isBlank());
+        assertFalse("x ".isBlank());
+        assertTrue("\u001c".isBlank());
+        assertFalse("\u00a0".isBlank());
+    }
+
+    public void testStrip() {
```
niloc132, line 35: Esp given https://github.com/gwtproject/gwt/issues/9973, would you mind also checking that the trim() tests correctly leave these non-ascii whitespace chars intact?

Also, from the same issue, should we have tests for `\ufeff` and `\u180e` to ensure they are preserved? Granted they would fail at this time, but might be good to have them while touching relevant tests (and comment them out, link them to the new issue).

user/super/com/google/gwt/emul/java/lang/String.java
zbynek: >  Can we not instead scan from processed to find '\r' - if success test next char for '\n', if fail re-scan from processed?

We have to be careful about `a\nb\r\n` where 'scanning for `\r` first' skips the first `\n`, but your point is valid -- currently for the string `\n\n\n\n\n\n` we'd check `O(length^2)` times if a character is `\r`, which is terrible. If we don't recompute the positions that are still valid from last iteration we can only check each character once for `\r` and once for `\n` which is optimal.

niloc132 (commented):
Thanks for your patience while I found time to review. Added a lot of comments, none are critical to merging, but mostly making sure I understand and that there isn't a reason to make changes now while we're thinking about it.

user/super/com/google/gwt/emul/java/lang/String.java
```diff
@@ -755,6 +760,92 @@ public String trim() {
...
+      }
+    }
+    return length;
+  }
+
+  private static class LinesSpliterator extends Spliterators.AbstractSpliterator<String> {
+    private int nextIndex = 0;
+    private String content;
+    int rPosition = -2;
+    int nPosition = -2;
```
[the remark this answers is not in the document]
zbynek: Some version of the code needed to distinguish between "has no more newlines" and "didn't look for newlines yet" but that's no longer needed and -2 looks confusing, changing to -1.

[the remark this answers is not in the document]
niloc132: If it is deliberate as a "this isn't -1", I don't object, just want to be sure I understand.

user/test/com/google/gwt/emultest/java11/lang/StringTest.java
zbynek: I added more testcases, then I stumbled upon
```

  /**
   * Helper method for testTrim to avoid compiler optimizations.
   *
   * TODO: insufficient, compiler now inlines.
   */
  public void trimRightAssertEquals(String left, String right) {
    assertEquals(left, right.trim());
  }
  ```
  Is there a way to check the compiler output to see if the test is actually testing the emulation?

File: user/test/com/google/gwt/emultest/java11/lang/StringTest.java [hunk 5]
```diff
@@ -0,0 +1,102 @@
...
+  public void testStripTrailing() {
+    stripRightTrailingAsssertEquals("", "");
+    stripRightTrailingAsssertEquals("", "  ");
+    stripRightTrailingAsssertEquals(" x", " x ");
+    stripRightTrailingAsssertEquals("\u001cx", "\u001cx\u001c");
+    stripRightTrailingAsssertEquals("\u00a0x\u00a0", "\u00a0x\u00a0 ");
+  }
+
+  private void stripRightAsssertEquals(String expected, String arg) {
+    assertEquals(expected, arg.strip())
```
niloc132, line 61:
```suggestion
    assertEquals(expected, arg.strip());
```

user/test/com/google/gwt/emultest/java11/lang/StringTest.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  private void stripRightLeadingAsssertEquals(String expected, String arg) {
+    assertEquals(expected, arg.stripLeading())
```
niloc132, line 65:
```suggestion
    assertEquals(expected, arg.stripLeading());
```

user/test/com/google/gwt/emultest/java11/lang/StringTest.java
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  private void stripRightTrailingAsssertEquals(String expected, String arg) {
+    assertEquals(expected, arg.stripTrailing())
```
niloc132, line 69:
```suggestion
    assertEquals(expected, arg.stripTrailing());
```

user/test/com/google/gwt/emultest/java11/lang/StringTest.java
```diff
@@ -0,0 +1,102 @@
...
+    assertEquals(Arrays.asList("a", "b", "c", "d"),
+        "a\rb\nc\r\nd".lines().collect(Collectors.toList()));
+    assertEquals(Arrays.asList("a"),
+        "a\n".lines().collect(Collectors.toList()));
+    assertEquals(Arrays.asList("a"),
+        "a\r\n".lines().collect(Collectors.toList()));
+    assertEquals(Arrays.asList(),
+        "".lines().collect(Collectors.toList()));
+    assertEquals(Arrays.asList(""),
+        "\n".lines().collect(Collectors.toList()));
```
niloc132, line 94:
```
Testcase: testLines took 0.001 sec
	Caused an ERROR
fromIndex: 2, toIndex: 1, length: 8
java.lang.StringIndexOutOfBoundsException: fromIndex: 2, toIndex: 1, length: 8
	at java.lang.Throwable.Throwable(Throwable.java:66)
	at java.lang.Exception.Exception(Exception.java:29)
	at java.lang.RuntimeException.RuntimeException(RuntimeException.java:29)
	at java.lang.IndexOutOfBoundsException.IndexOutOfBoundsException(IndexOutOfBoundsException.java:29)
	at java.lang.StringIndexOutOfBoundsException.StringIndexOutOfBoundsException(StringIndexOutOfBoundsException.java:30)
	at javaemul.internal.InternalPreconditions.checkCriticalStringBounds(InternalPreconditions.java:585)
	at java.lang.String.$substring(InternalPreconditions.java:568)
	at java.lang.String$LinesSpliterator.tryAdvance(String.java:832)
	at java.util.Spliterator.$forEachRemaining(Spliterator.java:52)
	at java.util.Spliterators$BaseSpliterator.forEachRemaining(Spliterator.java:51)
	at java.util.stream.StreamImpl.$reduce(StreamImpl.java:663)
	at java.util.stream.StreamImpl.$collect(StreamImpl.java:592)
	at com.google.gwt.emultest.java11.lang.StringTest.testLines(StringTest.java:94)
```

user/test/com/google/gwt/emultest/java11/lang/StringTest.java
niloc132: Good catch, here's what I tend to do: 
 * add `-setProperty compiler.useSourceMaps=true` to `test.args`, this disables the stack trace emulation that makes the output hard to read. Note that this may slightly break your stack traces, the line numbers won't quite be accurate (but close)
 * add `-style PRETTY` also to `test.args`, this makes the output more or less readable
 * add some apparent nondeterminism, or go through JS to your code to make the compiler unable to optimize - for example, an always-true condition like `Math.random() < 5` in an `if` or ternary, or a JSNI unchecked cast wrapping a value

From the first two, when I look at the JS file in build/out/user/test/web-htmlunit/www/com.google.gwt.emultest.EmulSuite.JUnit/ I see
```
_.testStrip = function testStrip(){
  assertEquals_1('', '');
  assertEquals_1('', '');
  assertEquals_1('x', 'x');
  assertEquals_1('x', 'x');
  assertEquals_1('\xA0x\xA0', '\xA0x\xA0');
  assertEquals_1('\uFFEFx\u180E', '\uFFEFx\u180E');
}
```
Clearly that's no good - testRepeat at least is the same:
```
_.testRepeat = function testRepeat(){
  var ex;
  assertEquals_1('', '');
  assertEquals_1('foo', 'foo');
  assertEquals_1('foofoofoo', 'foofoofoo');
  try {
    checkCriticalArgument(false, 'count is negative: -1');
    'foo'.repeat(-1);
    throw toJs(new Error_0('Should fail with negative arg'));
  }
  // ...
```

Note however that the draft tests do better, since the compiler doesnt optimize as heavily:
```
_.stripRightAsssertEquals = function stripRightAsssertEquals(expected, arg){
  jf.assertEquals_1(expected, jl.strip__Ljava_lang_String___devirtual$(arg));
}
;
//...
_.testStrip = function testStrip(){
  this.stripRightAsssertEquals('', '');
  this.stripRightAsssertEquals('', '  ');
  this.stripRightAsssertEquals('x', ' x ');
  this.stripRightAsssertEquals('x', '\x1Cx\x1C');
  this.stripRightAsssertEquals('\xA0x\xA0', '\xA0x\xA0 ');
  this.stripRightAsssertEquals('\uFFEFx\u180E', '\uFFEFx\u180E ');
}
;
```
```
_.testRepeat = function testRepeat(){
  var ex, noFoo;
  jf.assertEquals_1('', jl.repeat_I_Ljava_lang_String___devirtual$('foo', 0));
  jf.assertEquals_1('foo', jl.repeat_I_Ljava_lang_String___devirtual$('foo', 1));
  jf.assertEquals_1('foofoofoo', jl.repeat_I_Ljava_lang_String___devirtual$('foo', 3));
```

This is caused by DeadCodeElimination.tryOptimizeStringCall, where it finds a string literal with a plain method call on it, and calls it reflectively in the JVM at compile time, saving the compiled JS from needing to do it.

So we need to "hide" the string constants from the compiler so it doesnt just call trim() or repeat() on it, for example, StringTest.java does this (but doesn't use it for trim tests?):
https://github.com/gwtproject/gwt/blob/4aee3bd9df97e2b96a444751ea26a3a9091223f5/user/test/com/google/gwt/emultest/java/lang/StringTest.java#L957-L963

Used like this:
```
  private void stripRightAsssertEquals(String expected, String arg) {
    assertEquals(expected, hideFromCompiler(arg).strip());
  }
```
This makes for some awful output:
```
_.testStrip = function testStrip(){
  assertEquals_1('', $strip(($wnd.Math.random() < -1 && fail() , '')));
  assertEquals_1('', $strip(($wnd.Math.random() < -1 && fail() , '  ')));
  assertEquals_1('x', $strip(($wnd.Math.random() < -1 && fail() , ' x ')));
  assertEquals_1('x', $strip(($wnd.Math.random() < -1 && fail() , '\x1Cx\x1C')));
  assertEquals_1('\xA0x\xA0', $strip(($wnd.Math.random() < -1 && fail() , '\xA0x\xA0 ')));
  assertEquals_1('\uFFEFx\u180E', $strip(($wnd.Math.random() < -1 && fail() , '\uFFEFx\u180E ')));
}
;
_.testStripLeading = function testStripLeading(){
  assertEquals_1('', $stripLeading(($wnd.Math.random() < -1 && fail() , '')));
  assertEquals_1('', $stripLeading(($wnd.Math.random() < -1 && fail() , '  ')));
  assertEquals_1('x ', $stripLeading(($wnd.Math.random() < -1 && fail() , ' x ')));
  assertEquals_1('x\x1C', $stripLeading(($wnd.Math.random() < -1 && fail() , '\x1Cx\x1C')));
  assertEquals_1('\xA0x\xA0', $stripLeading(($wnd.Math.random() < -1 && fail() , '\xA0x\xA0')));
}
;
_.testStripTrailing = function testStripTrailing(){
  assertEquals_1('', $stripTrailing(($wnd.Math.random() < -1 && fail() , '')));
  assertEquals_1('', $stripTrailing(($wnd.Math.random() < -1 && fail() , '  ')));
  assertEquals_1(' x', $stripTrailing(($wnd.Math.random() < -1 && fail() , ' x ')));
  assertEquals_1('\x1Cx', $stripTrailing(($wnd.Math.random() < -1 && fail() , '\x1Cx\x1C')));
  assertEquals_1('\xA0x\xA0', $stripTrailing(($wnd.Math.random() < -1 && fail() , '\xA0x\xA0 ')));
}
;
```
but we can clearly see that it is no longer able to optimize away the call. I'll put up the change for the plain StringTest class, no idea how that was ignored for so long...

niloc132 (commented):
Some notes from the build, testing locally.

niloc132:
Please give https://github.com/gwtproject/gwt/pull/9986 a look to resolve those other TODOs.

niloc132 (commented):
All tests pass at this time.

user/super/com/google/gwt/emul/java/lang/Character.java
```diff
@@ -322,7 +322,7 @@ private static boolean isWhitespace(String ch) {
       // the Java definition includes separators.
       whitespaceRegex =
           new NativeRegExp(
-              "[\\u1680\\u180E\\u2000-\\u2006\\u2008-\\u200A\\u2028\\u2029\\u205F\\u3000\\uFEFF]"
+              "[\\u1680\\u2000-\\u2006\\u2008-\\u200A\\u2028\\u2029\\u205F\\u3000]"
```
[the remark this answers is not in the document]
zbynek: yes 👍

niloc132 (approved):
Approved, will give it a day or so for other reviewers, then merge if there are no other comments.

PR #9979, opened 2024-07-11
Update CI for Java 22, improved PR checks

niloc132:
Updates nightly builds and PR checks to verify they run cleanly with Java 22, and ensures that all tests compile for PR checks.

Also tries to fix reviewdog annotations by adding a newly required parameter.

### DIFF
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


niloc132:
It looks like the bytecode being generated by javac is somewhat different (even with the same --release level) with Java22, which is throwing off two tests in `GeneratedClassnameFinderTest`. This doesn't look like a real bug, just that previous javac versions omitted some references that are now generated. 

I'll do more testing/investigation and reopen this when ready.

vegegoku (approved):

PR #9981, opened 2024-07-17
Implement Java 14 switch expressions

### LINKED ISSUE #9878: Add compiler implementation, tests for switch expressions for Java 14
See https://openjdk.org/jeps/361.

At first glance, this seems like a fairly easy construct to consider, but at a minimum it requires executing statements inside of expressions - this seems to either require decomposing the enclosing expression(s) into statements and executing a switch/case or if/else, or extracting statements to synthetic lambdas or methods (closing over appropriate other locals, etc). As a pathological example, consider this, a switch expression with only one case that executes unconditionally, with embedded statements, passed as an argument to a method call with other expressions that may have side effects:

```java
someMethod(obj.affectState(), switch(0) {
  default -> {
    List<String> list = new ArrayList<>();
    list.add("a");
    list.addAll(obj.affectState());
    yield list;
  }
});
```

Rewriting away just the switch expression while staying in Java could look like
```java
// All sibling expressions (but not children/parents as far as I can tell?) must
// be extracted to ensure they are executed in the correct order. These can/should
// be final.
final Collection<String> arg1 = obj.affectState();
// Provide another local for the switch expression's value, to be assigned in
// each case/default
final Collection<String> arg2;
switch(0) {// switch expression becomes switch statement, with child expressions
  default: {
    List<String> list = new ArrayList<>();
    list.add("a");
    list.addAll(obj.affectState());
    // yield <value> is rewritten to assignment and break
    arg2 = list;
    break;
  }
}
// Actual parent express resumes
someMethod(arg1, arg2);
```

Or, again as a synthetic method:
```java
// notice that we close over `obj`, but not .affectState(), as we might not invoke it
someMethod(obj.affectState(), syntheticSwitchMethod(obj));

...

private static Collection<String> syntheticSwitchMethods(OtherObject obj) {
  switch(0) {// switch expression becomes switch statement, with child expressions
    default: {
      List<String> list = new ArrayList<>();
      list.add("a");
      list.addAll(obj.affectState());
      // yield <value> is rewritten to return <value>
      return list;
    }
  }
}
```

Supporting multiple expressions values in each case, and the `->` label both appear to purely be syntactic sugar, and the JDT should automatically cover this for us.

niloc132:
Java switch expressions are implemented in JS as a function wrapped around a switch block, with `return` used for `yield`, and multi-valued `case` statements split into multiple entries. Most other optimizations that work on a switch statement at this time will not apply to switch expressions, these can be re-added at a later time, but other optimizations that have some specialization for switch statements have been tested or updated to handle switch expressions.

JEP https://openjdk.org/jeps/361
Fixes #9878


dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
```diff
@@ -2675,6 +2696,16 @@ public boolean visit(SwitchStatement x, BlockScope scope) {
       return true;
     }
 
+    @Override
+    public boolean visit(SwitchExpression x, BlockScope blockScope) {
+//      InternalCompilerException exception =
+//              new InternalCompilerException("Switch expressions not yet supported");
+//      exception.addNode(new JCaseStatement(makeSourceInfo(x), null));
+//      throw exception;
```
niloc132, line 2704: Delete this

dev/core/src/com/google/gwt/dev/jjs/impl/DeadCodeElimination.java
```diff
@@ -1206,17 +1221,21 @@ private boolean tryReduceSwitchWithConstantInput(JSwitchStatement s, Context ctx
...
-            if (caseValue.getValueObj().equals(targetValue.getValueObj())) {
-              matchingCase = caseStatement;
-              break;
+          } else {
+            for (JExpression expr : caseStatement.getExprs()) {
+              if (expr instanceof JValueLiteral) {
+                JValueLiteral caseValue = (JValueLiteral) expr;
+                if (caseValue.getValueObj().equals(targetValue.getValueObj())) {
+                  matchingCase = caseStatement;
+                  break statements;
```
niloc132, line 1236: See about rewriting to avoid the labeled break, maybe a stream/optional operation?

niloc132:
Nightly build is passing https://github.com/niloc132/gwt/actions/runs/9978594294

jnehlmeier:
I assume it is ready for review or do you want to address your second code comment first?

niloc132:
I'd love some review yes, so we can get this shipped ASAP. I'm trying to finish up the closer-to-merging PRs, give people a chance to go over this, if interested, as its not as small as I had hoped.

dev/core/src/com/google/gwt/dev/jjs/ast/JCaseStatement.java
```diff
@@ -17,28 +17,68 @@
...
+    if (exprs.size() > 1) {
+      throw new IllegalStateException(
+              "JCaseStatement.getExpr() called on a node with multiple expressions " + exprs);
+    }
+    return exprs.isEmpty() ? null : exprs.get(0);
+  }
+
+  public List<JExpression> getExprs() {
+    return exprs;
+  }
```
vegegoku, line 57: It does not seems like this is being used anywhere now in gwt itself,  but if this is used internally by the compiler then why not just remove it instead of deprecation? is there any external usage for this?

dev/core/src/com/google/gwt/dev/jjs/ast/JCaseStatement.java
```diff
@@ -17,28 +17,68 @@
...
+      if (compareOperation == null) {
+        compareOperation = caseComparison;
+      } else {
+        compareOperation = new JBinaryOperation(getSourceInfo(), JPrimitiveType.BOOLEAN,
+                JBinaryOperator.OR, compareOperation, caseComparison);
+      }
+    }
+    assert compareOperation != null : this;
+    return compareOperation;
   }
```
vegegoku, line 76: Except for the default check - which I think is a guard since the places where this is called we should never get here - this is method seems to be generic and not specific to the switch case statements, So I am not sure if this the best place for it, seems like it belong somewhere in the GwtAstBuilder. WDYT?

dev/core/src/com/google/gwt/dev/jjs/impl/gflow/cfg/CfgBuilder.java
```diff
@@ -393,16 +377,14 @@ public boolean visit(JBreakStatement x, Context ctx) {
     @Override
     public boolean visit(JCaseStatement x, Context ctx) {
       pushNode(new CfgStatementNode<JStatement>(parent, x));
-      if (x.getExpr() != null) {
+      if (x.isDefault()) {
+        // default label
+      } else {
```
vegegoku, line 382: Should we invert this if and remove the empty block?

dev/core/src/com/google/gwt/dev/jjs/ast/JCaseStatement.java
niloc132: I only moved it here because I ended up having the logic in several places, or calling into some place like GwtAstBuilder.

The GwtAstBuilder type is intended for converting JDT ast into GWT's Java ast - but we only use this in code that optimizes out the GWT Java ast - DeadCodeElimination and CfgBuilder.

I'll Javadoc it so it is more clear, but this is definitely specific to a CaseStatement - the `getExprs()` call in the loop is what this is about. The goal is to rewrite a single `case` into an `if` so we can remove the `switch` entirely, which applies when optimizing code (after some case blocks are optimized out or enum values are removed, etc).

I don't see an obvious class to put this in as a static method, do you?

dev/core/src/com/google/gwt/dev/jjs/impl/gflow/cfg/CfgBuilder.java
niloc132: Yes, I will do that.

dev/core/src/com/google/gwt/dev/jjs/impl/DeadCodeElimination.java
```diff
@@ -1996,6 +2021,7 @@ private Object tryTranslateLiteral(JExpression maybeLiteral, Class<?> type) {
    * labeled break break to another context.
    */
   public static class FindBreakContinueStatementsVisitor extends JVisitor {
+    // TODO add yield support?
```
jnehlmeier, line 2024: I don't think so. Looks like it is used with `do...while` and switch statements. continue/break can only occur in a switch expression case body within loops or switch statements. In both cases the switch expression case is not terminated. So I think if we ever need to search for `yield` we probably create a new visitor.

dev/core/src/com/google/gwt/dev/jjs/impl/gflow/cfg/CfgBuilder.java
```diff
@@ -645,6 +627,16 @@ public boolean visit(JStatement x, Context ctx) {
       throw new UnsupportedNodeException(x.getClass().toString());
     }
 
+    @Override
+    public boolean visit(JSwitchExpression x, Context ctx) {
```
jnehlmeier, line 631: Wondering about the consequences given that JSwitchStatement does quite a bit.

jnehlmeier (commented):
Seems to look good except that empty method in CfgBuilder. It is a blind spot now for the compiler analysis, right?

dev/core/src/com/google/gwt/dev/jjs/ast/JTransformer.java
```diff
@@ -574,6 +586,10 @@ public final void endVisit(JType x, Context ctx) {
     public final void endVisit(JUnaryOperation x, Context ctx) {
     }
 
+    @Override
+    public void endVisit(JUnsafeTypeCoercion x, Context ctx) {
```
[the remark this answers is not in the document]
niloc132: It was missing, probably didn't matter (since the only implementation of JTransformer is GenerateJavaScriptTransformer, and by definition an unsafe type coercion in JS is a no-op), but there are enough other little oversights like that that when I find them I try to fix them, or at least queue them up locally.

This one I ended up including since I was adding other methods to the same file, will remove if you'd prefer, please let me know.

dev/core/src/com/google/gwt/dev/jjs/ast/JCaseStatement.java
```diff
@@ -17,28 +17,68 @@
...
 
   public JCaseStatement(SourceInfo info, JExpression expr) {
     super(info);
-    this.expr = expr;
+    this.exprs = Collections.singletonList(expr);
+  }
+
+  public JCaseStatement(SourceInfo info, Collection<JExpression> exprs) {
+    super(info);
+    this.exprs = Collections.unmodifiableList(new ArrayList<>(exprs));
```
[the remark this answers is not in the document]
niloc132: We are still compiling with `--release 8`, so that's a no-go without more changers to the build files, probably best done separately (and then start chasing all of the other places that should be cleaned up in this way if we think it matters). 

This is in an effort to keep gwt-servlet etc compatible on the server with Java 8.

dev/core/src/com/google/gwt/dev/jjs/ast/JSwitchStatement.java
```diff
@@ -18,34 +18,34 @@
 import com.google.gwt.dev.jjs.SourceInfo;
 
 /**
- * Java switch statement.
+ * Wrapper to represent a Java switch statement as a JStatement.
```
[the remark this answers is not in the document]
niloc132: Yes, correct, thank you.

dev/core/src/com/google/gwt/dev/jjs/impl/CloneExpressionVisitor.java
```diff
@@ -255,6 +256,11 @@ public void endVisit(JPermutationDependentValue x, Context ctx) {
         "this point but contains " + x);
   }
 
+  @Override
+  public boolean visit(JSwitchExpression x, Context ctx) {
```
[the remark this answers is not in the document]
niloc132: Really intentional. 

Consider that the switch expression is the _only_ kind of java expression that includes new statements which exist in the same scope as the statement in which the expression itself lives. As a result, if you include switch-expressions in the cloner, suddenly the cloner must be a general purpose Node cloner, since any statement can be cloned as well.

dev/core/src/com/google/gwt/dev/jjs/impl/DeadCodeElimination.java
niloc132: I think you're right, but hadn't made the argument quite as convincingly to myself as you did.

dev/core/src/com/google/gwt/dev/jjs/impl/gflow/cfg/CfgBuilder.java
jnehlmeier: Oh I see. I just found https://github.com/gwtproject/gwt/commit/ea7228a75491f76b1240a7648eb48d198b19acc4

So maybe we should officially deprecate it for removal (or fix its bugs and keep it).

dev/core/src/com/google/gwt/dev/jjs/ast/JTransformer.java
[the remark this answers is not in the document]
jnehlmeier: Keep it in. We won't revert this PR anyways in the future.

dev/core/src/com/google/gwt/dev/jjs/ast/JCaseStatement.java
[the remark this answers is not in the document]
jnehlmeier: Hm ok. I only had the release notes of 2.11 in mind which states that it is likely the last minor release that supports running on Java 8.

jnehlmeier (approved):

PR #9984, opened 2024-07-20
Deprecate and avoid unload event handler

### LINKED ISSUE #9908: Remove general usage of Window `unload` event
**GWT version:** GWT 2.11 and below
**Browser (with version):** Chrome 120 (soft rollout, see chrome link below)
**Operating System:** any

---

##### Description
Browsers are removing the `unload` event on the browser, as there are replacements available, and the unload event can interfere with newer browser features. See the [MDN docs on the unload event](https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event) for more specifics.

Chrome specifically has a deprecation policy that is warning that this will be removed in the future. For GWT's own usage of unload, this shouldn't matter:
 * Other window events are removed when this event goes off, to avoid memory leaks (only applies to ancient browsers)
 * RootPanel itself adds a close handler to ensure that all widgets are disconnected, to avoid memory leaks (as above, no longer applies)
 * The long-deprecated `WindowCloseListener` handling also uses this feature.

##### Steps to reproduce
Use some Window/RootPanel API directly to monitor unload events, or indirectly with window resize/scroll/beforeunload events (such as with PlaceController, DialogBox, RootLayoutPanel). In chrome, you may see a warning/errors that unload events are deprecated for removal.

To fix this, we can probably remove this class of "memory-leak prevention" (which no longer applies in any browser created in the last ten years) - this means RootPanel need not use it, and Window need not initialize close handlers for window resize/scroll events. Likewise, split unload/beforeunload handling to separate initializers. 

Window.addCloseHandler should also be deprecated.

This should probably be backported, given Chrome's timeline for this deprecation, unless we get 2.12 to a finished state (and it is barely started) in the next few weeks.

https://groups.google.com/g/google-web-toolkit/c/ZVc9yztBurs

##### Implications around Widget.onUnload and detach event handlers
With the removal of this event from Chrome and the feature from GWT itself, it is even less reliably than before to let a Widget handle its own "onDetach" or to let an Attach(attached=false) event handler observe it. 

The browser's [pagehide](https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event) event is one possible option here, but care must be taken:
 * This event is also not reliable
 * This event may be followed by a [pageshow](https://developer.mozilla.org/en-US/docs/Web/API/Window/pageshow_event) event indicating that the page has been displayed again, _without reloading_. This means that the application is still running, without re-creating DOM elements and widgets, so all widgets must be re-attached to continue handling events. If `event.persisted` is false, this is not a concern, there will be no subsequent `pageshow` event, but if it is true, the page _might_ be cached in its current state and later restored.

Developers should consider other approaches to this, such as directly handling `bfcache` `pageshow`/`pagehide` events in the handlers that need to know about the page status changing, rather than behaving as though the DOM elements were disassembled and reassembled.

niloc132:
Event handling for `unload` and `beforeunload` now is split, so that only explicit calls to `WindowImpl.initWindowCloseHandler()`, `Window.addCloseHandler()`, or `Window.addWindowCloseListener()` can cause the `unload` event handler to be attached - all of which are now deprecated. The `beforeunload` event wiring is now the only event handler set up when `Window.addWindowClosingHandler` etc are called.

RootPanel methods formerly intended to prevent memory leaks during the `unload` event are deprecated, since modern browsers do not leak these handlers, and may prevent correct behavior of apps using bfcache.

Documentation changed on using various static wrap(Element) methods to reflect how to detach, if desired, when the page is closed or no longer in use.

Fixes #9908

### DIFF
--- a/user/src/com/google/gwt/user/client/impl/WindowImpl.java
+++ b/user/src/com/google/gwt/user/client/impl/WindowImpl.java
@@ -62,12 +62,7 @@ public class WindowImpl {
       try {
         ret = $entry(@com.google.gwt.user.client.Window::onClosing())();
       } finally {
-        try {
-          oldRet = oldOnBeforeUnload && oldOnBeforeUnload(evt);
-        } catch (unhandled) {
-          // Pass to uncaught exception handler, if registered
-          $wnd.onerror(unhandled);
-        }
+        oldRet = oldOnBeforeUnload && oldOnBeforeUnload(evt);
       }
       // Avoid returning null as IE6 will coerce it into a string.
       // Ensure that "" gets returned properly.
--- a/user/src/com/google/gwt/user/client/impl/WindowImpl.java
+++ b/user/src/com/google/gwt/user/client/impl/WindowImpl.java
@@ -64,7 +64,10 @@ public class WindowImpl {
       } finally {
         try {
           oldRet = oldOnBeforeUnload && oldOnBeforeUnload(evt);
-        } catch (ignored) {}
+        } catch (unhandled) {
+          // Pass to uncaught exception handler, if registered
+          $wnd.onerror(unhandled);
+        }
       }
       // Avoid returning null as IE6 will coerce it into a string.
       // Ensure that "" gets returned properly.
--- a/user/src/com/google/gwt/user/client/impl/WindowImpl.java
+++ b/user/src/com/google/gwt/user/client/impl/WindowImpl.java
@@ -29,11 +29,30 @@ public class WindowImpl {
   public native String getQueryString() /*-{
     return $wnd.location.search;
   }-*/;
-  
-  public native void initWindowCloseHandler() /*-{
+
+  @Deprecated
+  public void initWindowCloseHandler() {
+    initWindowUnloadHandler();
+    initWindowBeforeUnloadHandler();
+  }
+
+  @Deprecated
+  public native void initWindowUnloadHandler() /*-{
+    var oldOnUnload = $wnd.onunload;
+
+    $wnd.onunload = $entry(function(evt) {
+      try {
+        @com.google.gwt.user.client.Window::onClosed()();
+      } finally {
+        oldOnUnload && oldOnUnload(evt);
+        $wnd.onunload = null;
+      }
+    });
+  }-*/;
+
+  public native void initWindowBeforeUnloadHandler() /*-{
     var oldOnBeforeUnload = $wnd.onbeforeunload;
-    var oldOnUnload =  $wnd.onunload;
-    
+
     // Old mozilla doesn't like $entry's explicit return statement and
     // will always pop up a confirmation dialog.  This is worked around by
     // just wrapping the call to onClosing(), which still has the semantics
@@ -43,7 +62,9 @@ public class WindowImpl {
       try {
         ret = $entry(@com.google.gwt.user.client.Window::onClosing())();
       } finally {
-        oldRet = oldOnBeforeUnload && oldOnBeforeUnload(evt);
+        try {
+          oldRet = oldOnBeforeUnload && oldOnBeforeUnload(evt);
+        } catch (ignored) {}
       }
       // Avoid returning null as IE6 will coerce it into a string.
       // Ensure that "" gets returned properly.
@@ -55,18 +76,6 @@ public class WindowImpl {
       }
       // returns undefined.
     };
-    
-    $wnd.onunload = $entry(function(evt) {
-      try {
-        @com.google.gwt.user.client.Window::onClosed()();
-      } finally {
-        oldOnUnload && oldOnUnload(evt);
-        $wnd.onresize = null;
-        $wnd.onscroll = null;
-        $wnd.onbeforeunload = null;
-        $wnd.onunload = null;
-      }
-    });
   }-*/;
 
   public native void initWindowResizeHandler() /*-{
--- a/user/src/com/google/gwt/user/client/ui/RootPanel.java
+++ b/user/src/com/google/gwt/user/client/ui/RootPanel.java
@@ -89,11 +89,18 @@ public class RootPanel extends AbsolutePanel {
    * This method may only be called per widget, and only for widgets that were
    * originally passed to {@link #detachOnWindowClose(Widget)}.
    * </p>
-   * 
+   * <p>
+   * Note that modern browsers do not have the memory leaks that originally required use of this
+   * feature - it is retained only to support application-specific detach events.
+   * </p>
+   *
    * @param widget the widget that no longer needs to be cleaned up when the
    *          page closes
    * @see #detachOnWindowClose(Widget)
+   * @deprecated Instead, prefer to use the {@code bfcache}, {@code pagehide},
+   * {@code visibilitychange} events, etc.
    */
+  @Deprecated
   public static void detachNow(Widget widget) {
     assert widgetsToDetach.contains(widget) : "detachNow() called on a widget "
         + "not currently in the detach list";
@@ -123,10 +130,17 @@ public class RootPanel extends AbsolutePanel {
    * contained in another widget. This is to ensure that the DOM and Widget
    * hierarchies cannot get into an inconsistent state.
    * </p>
-   * 
+   * <p>
+   * Note that modern browsers do not have the memory leaks that originally required use of this
+   * feature - it is retained only to support application-specific detach events.
+   * </p>
+   *
    * @param widget the widget to be cleaned up when the page closes
    * @see #detachNow(Widget)
+   * @deprecated Instead, prefer to use the {@code bfcache}, {@code pagehide},
+   * {@code visibilitychange} events, etc.
    */
+  @Deprecated
   public static void detachOnWindowClose(Widget widget) {
     assert !widgetsToDetach.contains(widget) : "detachOnUnload() called twice "
         + "for the same widget";
@@ -218,16 +232,36 @@ public class RootPanel extends AbsolutePanel {
 
   /**
    * Determines whether the given widget is in the detach list.
-   * 
+   * <p>
+   * Note that modern browsers do not have the memory leaks that originally required use of this
+   * feature - it is retained only to support application-specific detach events.
+   * </p>
+   *
    * @param widget the widget to be checked
    * @return <code>true</code> if the widget is in the detach list
+   * @deprecated Instead, prefer to use the {@code bfcache}, {@code pagehide},
+   * {@code visibilitychange} events, etc.
    */
+  @Deprecated
   public static boolean isInDetachList(Widget widget) {
     return widgetsToDetach.contains(widget);
   }
 
-  // Package-protected for use by unit tests. Do not call this method directly.
-  static void detachWidgets() {
+  /**
+   * Detaches all widgets that were set to be detached on window close by a call to
+   * {@link #detachOnWindowClose}. Formerly was package-protected, now can be called by projects
+   * that required the old behavior and are willing to set up their own onclose handler on the
+   * window.
+   * <p>
+   * Note that modern browsers do not have the memory leaks that originally required use of this
+   * feature - it is retained only to support application-specific detach events.
+   * </p>
+   *
+   * @deprecated Instead, prefer to use the {@code bfcache}, {@code pagehide},
+   * {@code visibilitychange} events, etc.
+   */
+  @Deprecated
+  public static void detachWidgets() {
     // When the window is closing, detach all widgets that need to be
     // cleaned up. This will cause all of their event listeners
     // to be unhooked, which will avoid potential memory leaks.
--- a/user/src/com/google/gwt/user/client/ui/RootPanel.java
+++ b/user/src/com/google/gwt/user/client/ui/RootPanel.java
@@ -89,16 +89,11 @@ public class RootPanel extends AbsolutePanel {
    * This method may only be called per widget, and only for widgets that were
    * originally passed to {@link #detachOnWindowClose(Widget)}.
    * </p>
-   * <p>
-   * Note that modern browsers do not have the memory leaks that originally required use of this
-   * feature - it is retained only to support application-specific detach events.
-   * </p>
    *
    * @param widget the widget that no longer needs to be cleaned up when the
    *          page closes
    * @see #detachOnWindowClose(Widget)
-   * @deprecated Instead, prefer to use the {@code bfcache}, {@code pagehide},
-   * {@code visibilitychange} events, etc.
+   * @deprecated Instead, use {@link Widget#removeFromParent()}.
    */
   @Deprecated
   public static void detachNow(Widget widget) {
@@ -115,30 +110,28 @@ public class RootPanel extends AbsolutePanel {
   /**
    * Adds a widget to the detach list. This is the list of widgets to be
    * detached when the page unloads.
-   * 
+   *
    * <p>
    * This method must be called for all widgets that have no parent widgets.
    * These are most commonly {@link RootPanel RootPanels}, but can also be any
    * widget used to wrap an existing element on the page. Failing to do this may
    * cause these widgets to leak memory. This method is called automatically by
    * widgets' wrap methods (e.g.
-   * {@link Button#wrap(com.google.gwt.dom.client.Element)}).
+   * {@link Button#wrap(Element)}).
    * </p>
-   * 
+   *
    * <p>
    * This method may <em>not</em> be called on any widget whose element is
    * contained in another widget. This is to ensure that the DOM and Widget
    * hierarchies cannot get into an inconsistent state.
    * </p>
-   * <p>
-   * Note that modern browsers do not have the memory leaks that originally required use of this
-   * feature - it is retained only to support application-specific detach events.
-   * </p>
    *
    * @param widget the widget to be cleaned up when the page closes
    * @see #detachNow(Widget)
-   * @deprecated Instead, prefer to use the {@code bfcache}, {@code pagehide},
-   * {@code visibilitychange} events, etc.
+   * @deprecated While originally introduced to combat memory leaks in old browsers, this is no
+   * longer necessary, and the unload event used by this method is being removed from browsers.
+   * Additionally, it is unreliable as a means to ensure calls to {@link Widget#onUnload()}. See
+   * <a href="https://github.com/gwtproject/gwt/issues/9908">Issue 9908</a> for more information.
    */
   @Deprecated
   public static void detachOnWindowClose(Widget widget) {
@@ -239,8 +232,9 @@ public class RootPanel extends AbsolutePanel {
    *
    * @param widget the widget to be checked
    * @return <code>true</code> if the widget is in the detach list
-   * @deprecated Instead, prefer to use the {@code bfcache}, {@code pagehide},
-   * {@code visibilitychange} events, etc.
+   * @deprecated Use of the detach list requires the unload event, which is planned to be removed
+   * in future browser updates. See notice on {@link #detachOnWindowClose(Widget)} and
+   * <a href="https://github.com/gwtproject/gwt/issues/9908">Issue 9908</a> for more information.
    */
   @Deprecated
   public static boolean isInDetachList(Widget widget) {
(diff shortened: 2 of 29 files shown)


user/src/com/google/gwt/user/client/impl/WindowImpl.java
```diff
@@ -29,11 +29,30 @@ public native String getHash() /*-{
...
+  @Deprecated
+  public native void initWindowUnloadHandler() /*-{
+    var oldOnUnload = $wnd.onunload;
+
+    $wnd.onunload = $entry(function(evt) {
+      try {
+        @com.google.gwt.user.client.Window::onClosed()();
+      } finally {
+        oldOnUnload && oldOnUnload(evt);
+        $wnd.onunload = null;
```
jnehlmeier, line 48: Is there a reason why you kept this reset to `null` but removed the ones for `onresize`, `onscroll` and `onbeforeunload`?

user/src/com/google/gwt/user/client/impl/WindowImpl.java
```diff
@@ -43,7 +62,9 @@ public native void initWindowCloseHandler() /*-{
       try {
         ret = $entry(@com.google.gwt.user.client.Window::onClosing())();
       } finally {
-        oldRet = oldOnBeforeUnload && oldOnBeforeUnload(evt);
+        try {
+          oldRet = oldOnBeforeUnload && oldOnBeforeUnload(evt);
+        } catch (ignored) {}
```
jnehlmeier, line 67: Should we be brave enough and log it? In previous code the exception escaped here and might end up on developer console. Now it would definitely be swallowed.

user/src/com/google/gwt/user/client/ui/RootPanel.java
```diff
@@ -258,15 +253,6 @@ private static native Element getRootElement() /*-{
     return $doc;
   }-*/;
 
-  private static void hookWindowClosing() {
```
jnehlmeier, line 261: I think we should somehow keep that logic. Introducing `pagehide` event first and then use it here could be a solution. It fires unreliably on mobile, just like `beforeunload`, but it is fired after `beforeunload` so we can be sure the user hasn't canceled unloading the page.

If we fully delete this logic here, then `RootPanel.detachOnWindowClose()` is now a pretty bad method name and all Widgets created using `wrap(Element)` methods (also custom widgets) do not receive calls to `Widget.onDetach` anymore. As a result some code in end user applications isn't executed anymore.

user/src/com/google/gwt/user/client/impl/WindowImpl.java
niloc132: Since unload is not guaranteed to be run anyway, there is no benefit for other events - it won't consistently be invoked.

It will at least consistently clean itself up though - I'm _fairly_ certain that this only exists for (long since fixed) memory leaks, but it seemed like just in case there was some other clean up code somewhere, we'd want to still do this for itself. It does seem reasonable to fully remove though - open to your advice here.

user/src/com/google/gwt/user/client/impl/WindowImpl.java
niloc132: That's a fair point - I had intended to ensure that if `ret` was set we wouldn't cancel it, but I'm changing too much here, and should preserve those semantics.

user/src/com/google/gwt/user/client/ui/RootPanel.java
niloc132: Since it is unreliable, is it important to keep? The goal according to javadoc for detachNow and detachOnWindowClose is to ensure that objects are consistently GC'd, which is already certain to be true (since IE9 and FF 1.6 are long dead).

I agree that for (inconsistent) consistency we _could_ replace this with `pagehide` - but `pagehide` comes with different semantics that may confused application authors who previously assumed this behavior. For example, if using the `bfcache`, the app may continue to function after `pagehide`, and each of those widgets that was removed will need to be re-attached to resume handling events.  That's a pivot to a much bigger feature than what RootPanel is trying to support - and it may not be consistent with the current API, right?

jnehlmeier: Hmm yeah it would be unfortunate if `bfcache` restores the page from memory and nothing is in attached state anymore. The code could check `pagehide.persisted` and only detach if it is `false`. But who knows how often it will really be `false`.

My primary thinking was that we don't know what people really do in `onDetach` or as a consequence in `onUnload`, e.g. `postMessage` or something like that. Calling it unreliably might be better than not calling it at all. But I don't really have a strong opinion about it. However if we remove it here, then `RootPanel.detachOnWindowClose()` should probably be marked deprecated for removal with updated JavaDoc and `RootPanel.addToDetachList()` should be added as a replacement.

niloc132: I'm really loathe to expand the API here, but would instead prefer to focus efforts on the updated modules and enhancing them - no jsni, no generators, no rebind rules, and slowly "strangle" the old gwt-user code until it is phased out.

niloc132: On that note, I guess that either detachWidgets() should be removed, or made public and deprecated so that users can still call into it if they really want the old behavior? I've put up a patch doing this, with some documentation on alternatives.

user/src/com/google/gwt/user/client/ui/RootPanel.java
```diff
@@ -89,11 +89,18 @@ public void execute(Widget w) {
...
+   * feature - it is retained only to support application-specific detach events.
+   * </p>
+   *
    * @param widget the widget that no longer needs to be cleaned up when the
    *          page closes
    * @see #detachOnWindowClose(Widget)
+   * @deprecated Instead, prefer to use the {@code bfcache}, {@code pagehide},
+   * {@code visibilitychange} events, etc.
    */
+  @Deprecated
```
jnehlmeier, line 103: These changes go too far as we should not make these methods deprecated unless we want `Widget.onDetach` to become public.

I have looked through our app and we use `onLoad` and `onUnload` to:
- register/deregister event handlers for widget events (DOM as well as logical widget events). This was for old browsers but could be skipped today.
- register/deregister events on an app wide singleton eventbus
- register/deregister with other singleton classes, e.g. an autosave manager which handles timers internally to fire autosave events for registered widgets
- start/stop timer for a given widget for various tasks, e.g. something simple as a clock.
- reset internal widget state if we want to reuse a widget, so it starts with fresh state on attach
- probably a couple of other things

Now imagine you have a GWT application that loads HTML from the server, renders the HTML using `setInnerHTML` and then uses the various `*.wrap(Element)` methods of GWT widgets as well as custom widgets to make that HTML interactive. If the application loads and unloads many such HTML fragments it is important to cleanup the widgets corresponding to a single HTML fragment using `detachNow`. Otherwise they will stay in the static widget list in `RootPanel` until you leave the page causing memory to increase and increase.
Now you could argue that these `wrap` widgets should not put themselves into the detach list anymore because they don't have to (seen from a DOM event perspective). But then `onUnload` lifecycle methods won't work anymore and it will cause trouble for most of the above use cases.

This change is only about what happens if we close the window and it should not interfere with widget lifecycle in general while the application runs normally and just loads/unloads elements/widgets.

So I do think we cannot get rid of `detachNow` and as a consequence `isInDetachList` and `detachOnWindowClose` (possibly renamed to `addToDetachList`) in the future unless we decide to make `Widget.onDetach` public as a non-deprecated replacement for `detachNow` to support widget lifecycle methods for `Widget.wrap(Element)` widgets. Instead of `detachNow` users would then need to call `onDetach` directly and might need to reimplement a detachlist themselves if they need `isInDetachList` for some reason.

jnehlmeier (changes requested):

user/src/com/google/gwt/user/client/ui/RootPanel.java
niloc132: I think there's a compromise here that still lets us get rid of all of this (deprecating for now, probably leaving it all for at least a year, probably longer), but first let's clarify exactly what we're taking away and what it implies:

 * We're _not_ removing attach/detach or load/unload. At that point, we might as well get rid of Widget itself - definitely a choice that projects can make, but one we don't want to make for them.
 * Instead, the root here is the onclose handler that Window itself maintained, and provided for classes like RootPanel. This must be removed since otherwise we're actively preventing projects from using `bfcache`.
 * With that removed, RootPanel in turn must stop listening for the close, since likewise it would actively prevent projects from using `bfcache`. 
 * With RootPanel.detachOnClose gone, these other calls end up turning into a memory leak _especially_ in the case of `bfcache` being functional, so we want to actively discourage them - if a project is using one of the `wrap` factory methods with `detachNow`, they should switch instead to `widget.removeFromParent()`. This is not clear in the current version of the patch. This is strictly superior to exposing `onDetach` publicly, since it does checks to see how to detach the widget.
 * We definitely don't want projects to _start_ using `detachNow`, but existing projects can continue to use it and should expect it and `isInDetachList` to continue to behave properly.

Proposed migration path:
 * In GWT 2.N, we will deprecate these methods and improve documentation. Teams can continue to use these methods for some time, but documentation will encourage migrating from `detachNow(foo)` to `foo.removeFromParent()`. To continue to handle the memory leak in `RootPanel.widgetsToDetach`, `removeFromParent` will continue to call `detachNow(this)` as it has done. The patch already documents that calling `detachWidgets()` is required as the page closes if that "best effort" call is important to their app (but since it wasnt reliable to begin with, it shouldn't be assumed to work). 
 * In GWT 2.(N+1) or so, we will update Widget.removeFromParent() to stop calling detatchNow(this), and instead directly call onDetach - after an assert on `attached` to ensure that the widget should even be detached to begin with. The combined "is attached" and "parent is null" check should correctly account for the case of a widget that was created via `wrap()`, has not yet been detached, and may or may not still be in the dom. 
 * At this time, `detachNow` should be a no-op, and likewise `detachOnWindowClose` etc - the entire `widgetsToDetach` collection should go away, and calls to these methods should warn  - there's no memory leak, but detach isn't getting called any more here.
 * In GWT 2.(N+2), remove these deprecated no-op methods, to ensure swift migration and no surprises when things don't work.

I believe `RootPanel.get()` should be treated like `wrap()` calls through all of this - if you elect to call detachWidgets at some point, they will get cleaned up together. Alternatively, a RootPanel can be cleaned up directly through a call to removeFromParent().

Risks I see:
 * No one reads docs or release notes fully (obviously myself included). Some people check their build log when they see deprecations. Most people check runtime logs when bugs start happening. It is possible we should move the warnings to GWT 2.N so that we make noise a little sooner.
 * Projects that incorrectly call detachWidgets() while using the bfcache will be surprised that their events no longer work. Javadoc and release notes should at least point them in the right direction here.

With all of that said, attach/detach on page close will still do no better than best effort, but requires active participation in one place in a project after this patch - and it not called, these widgets should function correctly when using `bfcache` automatically, as they weren't detached when the page was hidden (and so you don't want unload/load going off). But when rearranging widgets or removing them from their `HasWidgets` parents, they will correctly continue to call `onAttach`/`onDetach` and fire their attach events.

---

If this clears up the issue, I have the meat of this implemented locally and a few new unit tests to confirm, but am still working on updating documentation.

niloc132: Pushed an update reflecting this, please take a look.

user/src/com/google/gwt/user/client/ui/RootPanel.java
```diff
@@ -126,10 +130,17 @@ public static void detachNow(Widget widget) {
    * contained in another widget. This is to ensure that the DOM and Widget
    * hierarchies cannot get into an inconsistent state.
    * </p>
-   *
```
jnehlmeier, line 129: Alternative:

Remove the added paragraph and then:

`@deprecated` This method was introduced to prevent memory leaks in very old browsers which is not relevant anymore. Also `unload` events are not fired reliably on mobile browsers and are not compatible with the back/forward cache (bfcache) in modern browers. If your application has code in `Widget.onUnload()` methods that needs to be executed if the window has been closed then consider using `pagehide` event instead. Note that `pagehide` events are compatible with bfcache but are still not fired reliably on mobile devices. Also note that bfcache effectively makes a memory dump on `pagehide` and a memory restore on `pageshow`. If you blindly execute cleanup code during `pagehide` your widget might not work correctly anymore once restored from bfcache. You should check `pagehide.persisted` before executing cleanup code and possibly also use `pageshow` together with `pageshow.persisted` to execute initialization code.

user/src/com/google/gwt/user/client/ui/RootPanel.java
```diff
@@ -223,16 +232,36 @@ public static native com.google.gwt.user.client.Element getBodyElement() /*-{
 
   /**
    * Determines whether the given widget is in the detach list.
-   *
```
jnehlmeier, line 226: Alternative: 

Remove added paragraph and then:

`@deprecated The detach list was used to prevent memory leaks in old browsers which is not relevant anymore. See deprecation notice in {@link #detachOnWindowClose} for further information.`

user/src/com/google/gwt/user/client/ui/RootPanel.java
```diff
@@ -223,16 +232,36 @@ public static native com.google.gwt.user.client.Element getBodyElement() /*-{
...
+   * Detaches all widgets that were set to be detached on window close by a call to
+   * {@link #detachOnWindowClose}. Formerly was package-protected, now can be called by projects
+   * that required the old behavior and are willing to set up their own onclose handler on the
+   * window.
+   * <p>
+   * Note that modern browsers do not have the memory leaks that originally required use of this
+   * feature - it is retained only to support application-specific detach events.
+   * </p>
+   *
+   * @deprecated Instead, prefer to use the {@code bfcache}, {@code pagehide},
```
jnehlmeier, line 260: `@deprecated See deprecation notice in {@link #detachOnWindowClose} for further information.`

user/src/com/google/gwt/user/client/impl/WindowImpl.java
jnehlmeier: I like the fact that `ret` now goes through the code path no matter what. Just log the ignored error and it should be fine.

user/src/com/google/gwt/user/client/ui/RootPanel.java
jnehlmeier: Can we use `@deprecated(forRemoval = true)`? It was added in Java 9 and we compile using `--release 8`. But given it is not new syntax, it might work? Then CI tooling might create a little noise.

user/src/com/google/gwt/user/client/impl/WindowImpl.java
niloc132: Sorry, how do you want that to be logged? I wouldn't want to directly use console.log, and uncaught ex handler requires wrapping in Throwable... I'm going with reporting to window.onerror (which will then typically be handled by the UCEH)? Reverting seems like a better option, but let me know what you think.

user/src/com/google/gwt/user/client/ui/RootPanel.java
niloc132: GWT's own implementation of Override doesn't yet support forRemoval/etc. Maybe hit that in the next release, and we can sprinkle it around?

user/test/com/google/gwt/user/client/ui/ImageTest.java
```diff
@@ -794,6 +794,23 @@ public void testWrapOfSubclass() {
...
+    String uid = Document.get().createUniqueId();
+    DivElement div = Document.get().createDivElement();
+    div.setInnerHTML("<img id='" + uid + "' src='counting-forwards.png'>");
+    Document.get().getBody().appendChild(div);
+
+    final TestImage image = TestImage.wrap(Document.get().getElementById(uid));
+    assertNotNull(image);
+
+    // Cleanup.
+    Document.get().getBody().appendChild(div);
```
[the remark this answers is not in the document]
niloc132: Oh that's weird - the original test does it too, I wonder why... I've removed both, tests still pass.

user/src/com/google/gwt/user/client/ui/RootPanel.java
niloc132: That's a bit heavy for a deprecated tag, how do you feel about an abbreviated version of that, plus a link to 9908 for more info/discussion?

user/src/com/google/gwt/user/client/ui/RootPanel.java
jnehlmeier: Oh ok. Yes then some other day.

user/src/com/google/gwt/user/client/ui/RootPanel.java
jnehlmeier: I am fine with either way. I just wanted to ensure that a reader also knows the pitfalls of using `pagehide` as a replacement.

user/src/com/google/gwt/user/client/impl/WindowImpl.java
```diff
@@ -43,7 +62,12 @@ public native void initWindowCloseHandler() /*-{
       try {
         ret = $entry(@com.google.gwt.user.client.Window::onClosing())();
       } finally {
-        oldRet = oldOnBeforeUnload && oldOnBeforeUnload(evt);
+        try {
+          oldRet = oldOnBeforeUnload && oldOnBeforeUnload(evt);
+        } catch (unhandled) {
+          // Pass to uncaught exception handler, if registered
+          $wnd.onerror(unhandled);
```
jnehlmeier, line 69: I actually had `console.log()` in mind. Calling `onerror` probably makes more sense as the exception would have ended up in that error handler as well, if it exists. Although personally I don't like it that much because the code doesn't stop executing here. So it is more like a warning log.

However calling `onerror` is more involved. First you need to check if `onerror` exists to not cause an error yourself and then you need to check if `unhandled` is an `Error` or not. If it is, you can pass in all five parameters `onerror(msg, sourceFileUrl, line, col, errorObj)` taken from `Error` otherwise only `onerror(unhandled)` is possible.

If `$wnd.onerror` does not exist the code could still fallback to `console.log(unhandled)`.

user/test/com/google/gwt/user/client/ui/ImageTest.java
```diff
@@ -790,10 +790,25 @@ public void testWrapOfSubclass() {
     assertNotNull(image);
 
     // Cleanup.
-    Document.get().getBody().appendChild(div);
```
jnehlmeier, line 793: I actually thought that it should be `removeChild()` and that the test would then cover that `detachNow()` does not fail if the element is already removed from the DOM. Reason is that you usually call `detachNow` AFTER the element has been removed from the DOM through some other means, just like the JavaDoc of `wrap()` states.

Same for the test you added with calling `removeFromParent()`. It should not fail if the Widget element isn't in the DOM anymore.

user/src/com/google/gwt/user/client/impl/WindowImpl.java
niloc132: The other option is to do like reportUncaughtException does: `setTimeout(function() {throw unhandled;},0)`. This will re-throw after the current code completes, and will either reach a registered onerror handler, or the browser will log it.

My preferences, to keep it simpler, from most preferred to least:
 * revert to original, just let it throw instead of returning a value
 * throw in a setTimeout
 * just console.error()
 * test the type of the error, and either invokie onerror or console.error()

zbynek: > throw in a setTimeout

Sounds good, but could we throw the JS exception and not the Java one? (c.f. https://github.com/gwtproject/gwt-core/issues/20)

jnehlmeier: Would re-throwing in a setTimeout execute given that this is beforeUnload and once beforeUnload has been completed the page will be shutdown by the browser? If that is somewhat brittle I would lean towards reverting the change. The developer then has to fix the old beforeUnload handler to make ours work.

niloc132: Reverted back to how this was originally done.

user/test/com/google/gwt/user/client/ui/ImageTest.java
niloc132: Ah, that makes more sense. I'll add a test so that we also cover the "detach even though it is connected" case, as the test used to do. That historically worked (since the old test was adding twice instead of never removing).

niloc132: All tests passing https://github.com/niloc132/gwt/actions/runs/11038034684

jnehlmeier (approved):

PR #9995, opened 2024-08-28
Make MAX_STRING_NODE_LENGTH configurable

### LINKED ISSUE #9578: RPC format relies on eval for large payloads and long strings
**GWT version:2.8.2** 
**Browser (with version):any** 
**Operating System:any** 

---

##### Description
From https://github.com/gwtproject/gwt/issues/8197#issuecomment-349342472, it turns out that RPC's server serialization stream writer still assumes that the browser needs concatenated strings and arrays. 

However, the strings were concatenated to support hosted mode (which apparently at the time couldn't support larger than 64kb strings), but that appears to have been fixed, and arrays were split to support a bug in IE6/7 where long arrays couldn't be eval'd, but again, we don't support IE6/7 any more.

So, we don't really need it. Additionally, this is an issue since it violates CSP, as eval isn't particularly safe, and some sites would like to forbid its use to further protect their data and their users.

I'll try removing these hacks and confirm that eval is no longer called (except of course in browsers that do not support `JSON.parse`), and confirm that running dev mode can handle giant strings correctly as well. This should remove the ability of the server to produce streams older than version 8 (since `ServerSerializationStreamWriter.writeHeader` is the only code that should be writing out a version for the client). Will leave in support on the client for reading older streams, as we've done in the past.

lgemeinhardt:
Make `ServerSerializationStreamWriter.MAX_STRING_NODE_LENGTH` configurable, to better work around #9578 and allow real JSON in all cases. 
This PR is a follow-up to #9961 (making the rpc serialization chunk size configurable).

Workaround #9578 
Followup #9961

### DIFF
--- a/user/src/com/google/gwt/user/server/rpc/impl/ServerSerializationStreamWriter.java
+++ b/user/src/com/google/gwt/user/server/rpc/impl/ServerSerializationStreamWriter.java
@@ -345,7 +345,7 @@ public final class ServerSerializationStreamWriter extends
    * This exists to work around a Rhino parser bug in the hosted mode client
    * that limits string node lengths to 64KB.
    */
-  private static final int MAX_STRING_NODE_LENGTH =  = Integer.getInteger("gwt.rpc.maxStringNodeLength", 0xFFFF);
+  private static final int MAX_STRING_NODE_LENGTH =  Integer.getInteger("gwt.rpc.maxStringNodeLength", 0xFFFF);
 
   static {
     /*
--- a/user/src/com/google/gwt/user/server/rpc/impl/ServerSerializationStreamWriter.java
+++ b/user/src/com/google/gwt/user/server/rpc/impl/ServerSerializationStreamWriter.java
@@ -345,7 +345,7 @@ public final class ServerSerializationStreamWriter extends
    * This exists to work around a Rhino parser bug in the hosted mode client
    * that limits string node lengths to 64KB.
    */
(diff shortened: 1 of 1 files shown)


user/src/com/google/gwt/user/server/rpc/impl/ServerSerializationStreamWriter.java
```diff
@@ -345,7 +345,7 @@ abstract void write(ServerSerializationStreamWriter stream, Object instance)
    * This exists to work around a Rhino parser bug in the hosted mode client
    * that limits string node lengths to 64KB.
    */
-  private static final int MAX_STRING_NODE_LENGTH = 0xFFFF;
+  private static final int MAX_STRING_NODE_LENGTH =  Integer.getInteger("gwt.rpc.maxStringNodeLength", 0xFFFF);
```
niloc132, line 348: (Sorry, reviewdog still isn't supplying annotations automatically...)

> /home/runner/work/gwt/gwt/gwt/user/src/com/google/gwt/user/server/rpc/impl/ServerSerializationStreamWriter.java:348:0: warning: Line is longer than 100 characters (found 111). (com.puppycrawl.tools.checkstyle.checks.sizes.LineLengthCheck)

lgemeinhardt: I added the line break

niloc132 (approved):

PR #9997, opened 2024-08-28
StackTraceDeobfuscator should never provide null methodName

### LINKED ISSUE #9996: NPE in StackTraceDeobfuscator for missing identifiers, plus incorrect methodNames from chromium
**GWT version:** HEAD-SNAPSHOT
**Browser (with version):** any/chrome
**Operating System:** any

This issue is discussing two related issues, one a regression in GWT since #9936 and the other a regression in the past few years in chromium.

* The GWT regression is mentioned at https://github.com/gwtproject/gwt/issues/9931#issuecomment-2283421092:

    > So if the outer if is entered, there is a possibility that
    >
    >      methodName = mappingForLine.getIdentifier().orElse(null);
    >
    > sets methodName to null.

    This results in an NPE while creating the replacement StackTraceElement, preventing reading the exception.

* The Chromium issue is that when building a stack element of the form `at method (file:row:col)`, the method name now may have a `new ` prefix. This is not found in the symbol map file, so cannot be deobfuscated (though sourcemaps can try to take the row/col and figure out which file it should have come from). Example JS stack trace (with -style PRETTY):
    ```
    java.lang.RuntimeException: Runtime Exception
        at RuntimeException_0.createError (com.google.gwt.user.LoggingRPCSuite.JUnit-0.js:912:10)
        at RuntimeException_0.initializeBackingError (com.google.gwt.user.LoggingRPCSuite.JUnit-0.js:946:40)
        at RuntimeException_0.Throwable_0 (com.google.gwt.user.LoggingRPCSuite.JUnit-0.js:877:8)
        at RuntimeException_0.Exception_0 (com.google.gwt.user.LoggingRPCSuite.JUnit-0.js:961:15)
        at new RuntimeException_0 (com.google.gwt.user.LoggingRPCSuite.JUnit-0.js:970:15)
        at $throwException (com.google.gwt.user.LoggingRPCSuite.JUnit-0.js:4315:14)
        at LoggingRPCTest.testStackMapDeobfuscation (com.google.gwt.user.LoggingRPCSuite.JUnit-0.js:4369:5)
    ```
    Instead of `new RuntimeException_0`, our symbol maps expect `RuntimeException_0`, so this prefix needs to be removed.

niloc132:
Improves accuracy of deobfuscated stack traces by using the best answer between original JS names, symbol maps, and source maps, by both avoiding taking an identifier from the sourcemap if none is provided, and normalizing away chromium specific prefixes in stack trace elements.

Fixes #9996
Follow-up #9936

### DIFF
--- a/user/src/com/google/gwt/core/server/StackTraceDeobfuscator.java
+++ b/user/src/com/google/gwt/core/server/StackTraceDeobfuscator.java
@@ -26,6 +26,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.URL;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
@@ -186,9 +187,9 @@ public abstract class StackTraceDeobfuscator {
       return null;
     }
     // Warm the symbol cache for all symbols in this stack trace.
-    Set<String> requiredSymbols = new HashSet<String>();
+    Set<String> requiredSymbols = new HashSet<>();
     for (StackTraceElement ste : st) {
-      requiredSymbols.add(ste.getMethodName());
+      requiredSymbols.add(normalizeSymbol(ste.getMethodName()));
     }
     loadSymbolMap(strongName, requiredSymbols);
 
@@ -197,6 +198,17 @@ public abstract class StackTraceDeobfuscator {
       newSt[i] = resymbolize(st[i], strongName);
     }
     return newSt;
+  }
+
+  /**
+   * Helper to clean up client-provided symbols.
+   */
+  private static String normalizeSymbol(String symbol) {
+    // Chrome prefixes "new " for constructor calls, eliminate the prefix
+    if (symbol.startsWith("new ")) {
+      symbol = symbol.substring(4);
+    }
+    return symbol;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


user/src/com/google/gwt/core/server/StackTraceDeobfuscator.java
```diff
@@ -199,6 +200,17 @@ public final StackTraceElement[] resymbolize(StackTraceElement[] st, String stro
     return newSt;
   }
 
+  /**
+   * Helper to clean up client-provided symbols.
+   */
+  private static String normalizeSymbol(String symbol) {
+    // Chrome prefixes "new " for constructor calls, eliminate the prefix
+    if (symbol.startsWith("new ")) {
```
niloc132, line 208: This could also be done in the client StackTraceCreator.ModernCollector, but it seemed lower impact to include it here instead.

niloc132:
Full CI build: https://github.com/niloc132/gwt/actions/runs/10600496224

user/src/com/google/gwt/core/server/StackTraceDeobfuscator.java
```diff
@@ -305,7 +317,13 @@ public final StackTraceElement resymbolize(StackTraceElement ste, String strongN
 
           if (declaringClass == null || declaringClass.equals(ste.getClassName())) {
             declaringClass = mappingForLine.getOriginalFile();
```
jnehlmeier, line 319: In the linked issue if JS produces method name `<anonymous>` the `declaringClass` now becomes a filename via sourcemaps. The deobfuscated stack trace then had

```java
at com/google/gwt/core/client/impl/Impl.java.METHODNAME(com/google/gwt/core/client/impl/Impl.java:383)
```
Maybe we should try stripping `.java` and replacing `/` with `.`?

jnehlmeier: Sure

jnehlmeier (approved):

PR #10014, opened 2024-10-18
Remove all user agent checks for MSIE or documentMode

niloc132:
This removes all checks of MSIE in the user agent string, and all checks of documentMode to verify which IE version is being used, as these browsers are long since unsupported.

Most of these changes are specific to tests or devmode, but there is also a change to simplify Throwable slightly (allowing it to work in node.js and other non-browser environments), and simplification of startup user agent checks.


jnehlmeier (approved):

zbynek:
Should `ScrollImpl.ScrollIImplTrident` and `ResizeLayoutPanel.ImplTrident` be also removed to get rid of all IE specific code here?

niloc132:
> Should `ScrollImpl.ScrollIImplTrident` and `ResizeLayoutPanel.ImplTrident` be also removed to get rid of all IE specific code here?

My thinking here is "yes, eventually" - the goal of this commit was to pick a few specific keywords to search and remove all traces of them (actually I skipped one that I think only appears in legacy dev mode), so as to be exhaustive about those terms. There are so many different ways to refer to old browsers that all must eventually be removed, so I would slightly prefer to go at this in stages.

So, next passes might remove all "Trident" or go after ff3.5, safari 5, android browser, etc issues, in reasonably small pieces. Some, like gwt-rpc de/serialization might be better to tackle as an entire use case at once, rather than specific keywords to search.

zbynek (approved):

PR #10021, opened 2024-10-29
Ignore all annotations in all scopes if source is missing

### LINKED ISSUE #10020: Missing sources for type annotations trip the compiler
**GWT version:** 2.11
**Operating System:** Windows

---

##### Description
Some usages of annotations lead to missing source errrors.

##### Steps to reproduce

**Usecase 1)**
* Add jsinterop 1.0.2 to your project
* Compile with GWT 2.11

```
[ERROR] Errors in 'jar:file:/.../base-1.0.2.jar!/jsinterop/base/Js.java'
         [ERROR] Line 140: Nullable cannot be resolved to a type
```

**Usecase 2**
Same  issue also reproducible with [j2objc annotations 2.8](https://mvnrepository.com/artifact/com.google.j2objc/j2objc-annotations/2.8), depending on placement. Annotating a field is OK

```java
@WeakOuter Object foo = new Object();
```

Annotating an anonymous class is not

```java
foo =  new @WeakOuter Object() {
   @Override
   public String toString() {
      return "Foo";
   }
}

```

##### Known workarounds

* Include all annotation sources

https://github.com/google/jsinterop-base/issues/19

niloc132:
JDT supports three kinds of annotations, and the ASTVisitor lets them be encountered in two different scopes. This patch ensures that all six cases are handled, and adds a test that uses each of the six cases - removing any one of the `visit` implementations will fail the test by incorrectly reporting the annotation as not having sources available.

Fixes #10020

### DIFF
--- a/dev/core/test/com/google/gwt/dev/jjs/impl/AnnotationCompilerTest.java
+++ b/dev/core/test/com/google/gwt/dev/jjs/impl/AnnotationCompilerTest.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2024 GWT Project Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under the License
+ * is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.gwt.dev.jjs.impl;
+
+import com.google.gwt.core.ext.UnableToCompleteException;
+import com.google.gwt.thirdparty.guava.common.base.Joiner;
+
+public class AnnotationCompilerTest extends FullCompileTestBase {
+  @Override
+  protected void optimizeJava() {
+  }
+
+  public void testUseAnnotationWithoutSource() throws UnableToCompleteException {
+    // The JDT represents annotations as three kinds: markers, which have no members,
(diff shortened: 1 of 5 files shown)


dev/core/test/com/google/gwt/dev/jjs/impl/AnnotationCompilerTest.java
```diff
@@ -0,0 +1,49 @@
...
+    String annotations = Joiner.on(" ").join(
+        "@SampleBytecodeOnlyMarkerAnnotation",
+        "@SampleBytecodeOnlySingleMemberAnnotation(\"abc\")",
+        "@SampleBytecodeOnlyNormalAnnotation(a=1, b=2)"
+    );
+
+    String code = Joiner.on('\n').join(
+        "package test;",
+        "import com.google.gwt.dev.jjs.impl.*;",
+        "public class EntryPoint<T extends " + annotations + " Object> {",
```
jnehlmeier, line 39: Should we go all-in here and add annotations everywhere allowed in source code so the test is basically a catch all test and not tailored to the available methods of JDT ASTVisitor (BinaryTypeReferenceVisitor)? 

JDT knows additional scopes (Module, CompilationUnit, Method) and if we update JDT some visitor might change and we might not detect it here given visitors have default implementations. For example block scope could become method scope.

niloc132: Worth considering, but the ASTVisitor doesn't check those, so the particular code in question that is actively looking for binary type refs so that it can complain about them in the logs (and throw an exception) isn't going to fail, at least as I read it.

Would you contribute a test that does that? I'm guessing there is an example out there of all the different ways you can stack annotations that can be used to extend this. We'd need (at least) three instances of every location, to hit marker/normal/single.

vegegoku (approved):

PR #10022, opened 2024-10-30
Remove broken JSNI references from MSIE cleanup

niloc132:
The previous PR broke three tests which still had JSNI references to removed methods - not caught by the "compile.tests" task. This patch finishes removing these checks in tests, so that we never skip certain assertions based on old browsers, which no longer exist.

Follow-up #10014


niloc132:
Passing CI build: https://github.com/niloc132/gwt/actions/runs/11581798862

PR #10027, opened 2024-11-02
Correct handling for SwitchStatements without a containing Block

### LINKED ISSUE #10024: Using a switch statement together with if- or for-statement without using {} crashes
Using a switch-statement with an if- or for-statement without using {}, the GWT compilation will throw an exception. Surrounding the switch with {} will fix that exception.

```
if ((vpos != null))
    switch (vpos) {
     case TOP :
         return "text-before-edge";
     case CENTER :
         return "central";
     case BASELINE :
         return "baseline";
     case BOTTOM :
         return "text-after-edge";
     }
```
will crash with this exception:

```
[INFO] Caused by: java.lang.ClassCastException: class com.google.gwt.dev.jjs.ast.JSwitchStatement cannot be cast to class com.google.gwt.dev.jjs.ast.JExpression (com.google.gwt.dev.jjs.ast.JSwitchStatement and com.google.gwt.dev.jjs.ast.JExpression are in unnamed module of loader 'app')
[INFO] 	at com.google.gwt.dev.jjs.impl.GwtAstBuilder$AstVisitor.pop(GwtAstBuilder.java:2816)
[INFO] 	at com.google.gwt.dev.jjs.impl.GwtAstBuilder$AstVisitor.endVisit(GwtAstBuilder.java:1087)
[INFO] 	... 25 more
[INFO]    [ERROR] at SvgNodePeer.java(201): if ((vpos != null))
[INFO]     switch (vpos) {
[INFO]     case TOP :
[INFO]         return "text-before-edge";
[INFO]     case CENTER :
[INFO]         return "central";
[INFO]     case BASELINE :
[INFO]         return "baseline";
[INFO]     case BOTTOM :
[INFO]         return "text-after-edge";
[INFO]     }
[INFO]       org.eclipse.jdt.internal.compiler.ast.IfStatement
```

while this will work: 

```
if ((vpos != null)) {
    switch (vpos) {
     case TOP :
         return "text-before-edge";
     case CENTER :
         return "central";
     case BASELINE :
         return "baseline";
     case BOTTOM :
         return "text-after-edge";
     }
}
```

Same issue with this code:

```
                for (InMemoryAuthorizationRule rule : rules)
                    switch (rule.computeRuleResult(operationRequest)) {
                        case DENIED:  result = AuthorizationRuleResult.DENIED; break; // Breaking as it's a final decision
                        case GRANTED: result = AuthorizationRuleResult.GRANTED; // Not breaking, as we need to check there is not another denying rule (which is priority)
                        case OUT_OF_RULE_CONTEXT: // just ignoring it and looping to the next
                    }
```

**GWT version:** 2.12.0
**Browser (with version):** all browser 
**Operating System:** all OS

---

niloc132:
Also fixed incorrect generics for related code, spotted while auditing other incomplete typechecks.

Fixes #10024

### DIFF
--- a/dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
@@ -2812,26 +2812,31 @@ public class GwtAstBuilder {
 
     protected JStatement pop(Statement x) {
       JNode pop = (x == null) ? null : pop();
-      if (x instanceof Expression) {
+      if (x instanceof Expression && pop instanceof JExpression) {
         return maybeBoxOrUnbox((JExpression) pop, (Expression) x).makeStatement();
       }
       return (JStatement) pop;
     }
 
-    @SuppressWarnings("unchecked")
     protected <T extends JStatement> List<T> pop(Statement[] statements) {
       if (statements == null) {
         return Collections.emptyList();
       }
-      List<T> result = (List<T>) popList(statements.length);
+      List<T> result = Lists.newArrayList();
+      List<? extends JNode> stack = popList(statements.length);
       int i = 0;
-      for (ListIterator<T> it = result.listIterator(); it.hasNext(); ++i) {
+      for (ListIterator<? extends JNode> it = stack.listIterator(); it.hasNext(); ++i) {
         Object element = it.next();
-        if (element == null) {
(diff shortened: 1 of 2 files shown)


niloc132:
Full CI build: https://github.com/niloc132/gwt/actions/runs/11638557464

vegegoku (approved):

PR #10035, opened 2024-11-07
Ignore all annotations in all scopes if source is missing

### LINKED ISSUE #10020: Missing sources for type annotations trip the compiler

niloc132:
JDT supports three kinds of annotations, and the ASTVisitor lets them be encountered in two different scopes. This patch ensures that all six cases are handled, and adds a test that uses each of the six cases - removing any one of the `visit` implementations will fail the test by incorrectly reporting the annotation as not having sources available.

Fixes #10020
See #10021

### DIFF
--- a/dev/core/src/com/google/gwt/dev/javac/BinaryTypeReferenceRestrictionsChecker.java
+++ b/dev/core/src/com/google/gwt/dev/javac/BinaryTypeReferenceRestrictionsChecker.java
@@ -25,6 +25,7 @@ import org.eclipse.jdt.internal.compiler.ast.SingleMemberAnnotation;
 import org.eclipse.jdt.internal.compiler.impl.Constant;
 import org.eclipse.jdt.internal.compiler.lookup.BinaryTypeBinding;
 import org.eclipse.jdt.internal.compiler.lookup.BlockScope;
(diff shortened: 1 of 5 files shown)


niloc132:
This is not a regression, but does impact projects trying to use jsinterop-base 1.0.2+.

vegegoku (approved):

PR #10059, opened 2024-12-13
Support JDT's implicit yield in switch statements

### LINKED ISSUE #10044: An single-assigment case-statement without {} within a switch generates a superfluous return, leading to function termination
**GWT version: 2.12.1
**Browser (with version): all
**Operating System: all

---

##### Description
Single assignments as case statements should behave the same whether within {} or not.

Instead, with {} around them, these assignments generate correct JS, but if not surrounded with {}, the generated case statements have superfluous return statements that lead to function termination

##### Steps to reproduce
```
public void onModuleLoad() {
  String s = "a";
  switch (s) {
    case "a" -> result = 1;
    default -> result = 2;
   }
   Window.alert("result:" + result); // is never reached 
}
```

generated JS:
```
_.onModuleLoad_0_g$ = function z_g$(){
  var s_0_g$;
  s_0_g$ = 'a';
  switch (s_0_g$) {
    case 'a':
      return this.result_1_g$ = 1;
    default:return this.result_1_g$ = 2;
  }
  X9c_g$('result:' + this.result_1_g$);
}
;
```

##### Known workarounds
Put {} around case-statements

```
public void onModuleLoad() {
  String s = "a";
  switch (s) {
    case "a" -> { result = 1; }
    default -> { result = 2; }
   }
   Window.alert("result:" + result); // is reached 
}
```

generated JS:
```
_.onModuleLoad_0_g$ = function z_g$(){
  var s_0_g$;
  s_0_g$ = 'a';
  switch (s_0_g$) {
    case 'a':
      {
        this.result_1_g$ = 1;
        break;
      }

    default:{
        this.result_1_g$ = 2;
        break;
      }

  }
  X9c_g$('result:' + this.result_1_g$);
}
;
```

niloc132:
JDT generates a yield for arrow cases if the arrow points at a expression, even though the expression cannot be returned. Synthesize a block wrapping that expression as a statement, and add a break to the end of it.

Fixes #10044

### DIFF
--- a/dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
@@ -545,9 +545,17 @@ public class GwtAstBuilder {
     @Override
     public void endVisit(YieldStatement x, BlockScope scope) {
       try {
-        SourceInfo info = makeSourceInfo(x);
-        JExpression expression = pop(x.expression);
-        push(new JYieldStatement(info, expression));
+        if (x.switchExpression == null) {
+          // This is an implicit 'yield' in a case with an arrow - synthesize a break instead and
+          // wrap with a block so that the child count in JDT and GWT matches.
+          SourceInfo info = makeSourceInfo(x);
+          JExpression pop = pop(x.expression);
+          push(new JBlock(info, pop.makeStatement(), new JBreakStatement(info, null)));
+        } else {
+          SourceInfo info = makeSourceInfo(x);
+          JExpression expression = pop(x.expression);
+          push(new JYieldStatement(info, expression));
+        }
       } catch (Throwable e) {
         throw translateException(x, e);
       }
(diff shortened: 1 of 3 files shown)


niloc132:
Passing nightly CI run https://github.com/niloc132/gwt/actions/runs/12319013728

dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
```diff
@@ -545,9 +545,17 @@ public void endVisit(BreakStatement x, BlockScope scope) {
...
     public void endVisit(YieldStatement x, BlockScope scope) {
       try {
-        SourceInfo info = makeSourceInfo(x);
-        JExpression expression = pop(x.expression);
-        push(new JYieldStatement(info, expression));
+        if (x.switchExpression == null) {
+          // This is an implicit 'yield' in a case with an arrow - synthesize a break instead and
+          // wrap with a block so that the child count in JDT and GWT matches.
+          SourceInfo info = makeSourceInfo(x);
+          JExpression pop = pop(x.expression);
```
zbynek, line 552: These assignments can be extracted from the `if`.

zbynek (approved):

PR #10061, opened 2024-12-15
Remove extra assignment to gwtOnLoad

### LINKED ISSUE #10060: gwtOnLoad is assigned twice in a single statement
**GWT version: 2.7 to 2.12.1
**Operating System:** any

---

##### Description
Discovered while exploring how compiled GWT modules start up, in looking for some ways to simplify this output slightly.

The current implementation generates code that assigns gwtOnLoad to itself, from the current init method. From the showcase JS in 2.12.1:
```javascript
var gwtOnLoad = gwtOnLoad = sDb;
```
This is the comment in latest that describes what this block of code is supposed to do:
https://github.com/gwtproject/gwt/blob/339188d7535b77472715bc4d0f54118e556e46ad/dev/core/src/com/google/gwt/dev/jjs/impl/GenerateJavaScriptAST.java#L2129-L2137
and the actual relevant code a few lines down:
https://github.com/gwtproject/gwt/blob/339188d7535b77472715bc4d0f54118e556e46ad/dev/core/src/com/google/gwt/dev/jjs/impl/GenerateJavaScriptAST.java#L2146-L2151

The init expression for the JsVar should not itself contain an assignment, unless there is some benefit to doing so - and in this case, I can't imagine any, but I might not be imaginative enough...

Reviewing git history, an earlier commit (that was never released had this code, with the apparent expectation of allowing linkers to rewrite gwtOnLoad:
https://github.com/gwtproject/gwt/blob/ec75b533b397a0911495b44667d547798c6beee4/dev/core/src/com/google/gwt/dev/jjs/impl/GenerateJavaScriptAST.java#L1957-L1971
As that was removed before it was even released, I'm inclined to ignore it as a fluke. The implementation that follows appears to do what it claims to do, but does use createAssignment to do so - perhaps in 339188d753 it was accidentally inlined instead of rewritten?

Next relevant change in history is simpler still, and dates to the beginning of of GWT's public history:
https://github.com/gwtproject/gwt/blame/ab0aa686820319e15d3adf123f4502dc3349f6de/dev/core/src/com/google/gwt/dev/jjs/impl/GenerateJavaScriptAST.java#L948-L969
No `var` or assignment at all.

---

Looking more closely at the code review for the change that introduced this https://gwt-review.googlesource.com/c/gwt/+/7821, there is some commentary modular compilation and the need to re-assign the variable, plus some bug in HtmlUnit - but the reassignment shouldn't need to bother the `a = a = b` pattern, which will unconditionally assign `b` to `a`, rather than somehow chain or redefine it. I'll do some more testing to be sure that nothing subtle is happening here in HtmlUnit or incremental compile. Note: I'm not 100% confident that "modular compile" is an earlier name for "incremental compile", but signs do point in that direction.

I'm inclined to think that this is just a bug introduced by accidentally reusing too much of the previous revision, and we can safely remove the extra assignment, and shrink every GWT app by about 10 bytes.

niloc132:
Only saves about 10 bytes per GWT application, but the extra assignment appears to serve no purpose. Also corrects some slightly inaccurate comments about how initialization behaves.

Fixes #10060


niloc132:
In reviewing other generated code, the only thing I can think of is that an application might want to avoid reassigning gwtOnLoad, in the same manner as we let a linker define goog.global:
```
$wnd.goog.global = $wnd.goog.global || $wnd
```
https://github.com/gwtproject/gwt/blob/23e73f92d741d858f9191318cebcc0214f97f485/dev/core/src/com/google/gwt/dev/jjs/impl/GenerateJavaScriptAST.java#L1295-L1305

vegegoku (approved):

PR #10063, opened 2024-12-16
Remove GWT's Map polyfill

### LINKED ISSUE #10062: Remove GWT's polyfill for Map
GWT contains a polyfill for Map, used to implement java.util.Map where the key was a string (and avoid hashing the string in JS).

The polyfill contains implementation details for sufficiently old browsers that
 * did not support Map at all
 * threw an exception instead of using the iterator protocol (ff 24-25 apparently)

Additionally the polyfill itself has special handling for browsers that didn't correctly support `Object.create(null)`.

The jsinterop mapping supports old browsers that couldn't use `delete` as a method name, which is no longer needed.

niloc132:
All currently supported browsers have a working Map implementation, so we don't need to include this in every compiled GWT application.

Fixes #10062


niloc132:
Passing nightly CI run https://github.com/niloc132/gwt/actions/runs/12354394728

zbynek (approved):

vegegoku (approved):

PR #10066, opened 2024-12-18
Correctly mark a javadoc code snippet as such

niloc132:
Prevents it from rendering as a link (which would only result in a 404).

Partial #8428


niloc132:
Corrected output:
image

zbynek (approved):

Issue #9582, opened 2017-12-22
JettyLauncher throws exception with Java 9 for "jrt:" URL


gwtuser:
**GWT version:** 2.8.2
**Browser (with version):** N/A
**Operating System:** OpenSUSE 42.2 64-bit

---

##### Description
When GWT Plugin v3 is run in SuperDevMode with Eclipse Oxygen, JDK9.0.1, and GWT 2.8.2, the built-in Jetty server (or its launcher) crashes with the following trace:

 Found resouce but unrecognized URL format: 'jrt:/java.sql/javax/sql/DataSource.class'
    Java.lang.NoClassDefFoundError: javax/sql/DataSource
	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1007)
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:174)
	at java.base/java.net.URLClassLoader.defineClass(URLClassLoader.java:545)
	at java.base/java.net.URLClassLoader.access$100(URLClassLoader.java:83)
	at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:453)
	at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:447)
	at java.base/java.security.AccessController.doPrivileged(Native Method)
	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:446)
	at org.eclipse.jetty.webapp.WebAppClassLoader.findClass(WebAppClassLoader.java:510)
	at com.google.gwt.dev.shell.jetty.JettyLauncher$WebAppContextWithReload$WebAppClassLoaderExtension.findClass(JettyLauncher.java:480)
	at org.eclipse.jetty.webapp.WebAppClassLoader.loadClass(WebAppClassLoader.java:441)
	at org.eclipse.jetty.webapp.WebAppClassLoader.loadClass(WebAppClassLoader.java:403)
	at java.base/java.lang.Class.forName0(Native Method)
	at java.base/java.lang.Class.forName(Class.java:375)

Note that due to the misspelling of "resouce" this message is easily traced back to JettyServer.java, which appears to be GWT's wrapper code.  It doesn't appear to recognize the "jrt:" format.

##### Steps to reproduce
Trigger loading of javax.sql.DataSource in JDK9.0.1

gwtuser:
[...]

Issue #10057, opened 2024-12-12
Change default dev mode server to static implementation


niloc132:
Starting in GWT 2.11 (see https://github.com/gwtproject/gwt/issues/9863 and https://github.com/gwtproject/gwt/pull/9866) a new dev mode server implementation was introduced - `com.google.gwt.dev.shell.StaticResourceServer`. At the same time, the JettyLauncher implementation was deprecated, and logs a warning to the console when used.

This ticket proposes that in GWT 2.13 the default be changed to StaticResourceServer, with the JettyLauncher left for at least one more version before being removed.

---

As an alternative, we could remove the default server entirely and change to `-noserver` by default, so that any user specifying `-server` needs to also pick the server implementation they want to use.

eliasbalasis:
I can see `JettyLauncher`, `-noserver` etc. disappearing eventually, primarily in favor of external server wiring but also in relation to maintenance headaches and related struggles.

Issue #10056, opened 2024-12-12
Proposal:  external server wiring for DevMode


eliasbalasis:
In relation to #477 and continuing from #9863:

Can we consider creating an external server wiring that lets DevMode start your server, for tomcat/jetty/etc of each version?

> This would let those different versions be decoupled from GWT itself, so that GWT doesn't somehow need to bake in each possible version of each servlet container that someone might want to use. The mechanism to do this already exists in GWT - extend the com.google.gwt.core.ext.ServletContainerLauncher type and return a com.google.gwt.core.ext.ServletContainer instance when the server is started. This will almost certainly require some classloader trickery to keep your client code correctly separated from your server, but it is definitely possible to do. The better solution though is to have proper separation between your client and server code - separate classpaths, and run with separate processes.

eliasbalasis:
see https://github.com/arquillian

niloc132:
Thanks for filing this - are you interested in developing or sponsoring this work? 

I imagine this to be an external project (or family of projects) to start with, then adopted to gwtproject and maintained collectively going forwards? It probably can start with a single implementation and generalize from what is learned there.

I think a developer would naively assume that such a tool would use a specific classloader for the server rather than something like [Cargo](https://github.com/codehaus-cargo), since that would still make it necessary to debug an external process. With that said, Cargo does have the benefit of nudging the user down the path of maintaining them separately, as they _likely_ already have a way to run their server without GWT at all - and it seems plausible that a single Cargo implementation could handle all the various implementations that it in turn supports.

Other considerations:
 * As pointed out, we'll want to make sure that these tools play nicely with IDE plugins - do we want to encourage an uberjar for each that holds the entire servlet container? Probably better to have a directory of jars, or a command to run that downloads all dependencies from maven into a single directory to point at.
 * With this on the horizon we should start considering flipping the default from https://github.com/gwtproject/gwt/issues/9863 to the new static-only server, and a version or two later removing the Jetty implementation entirely. This will enable updating Jetty in GWT itself with no side effects for users. There are more considerations here too that could lead us to consider dropping Jetty entirely - Jetty 11 is End of Community Support and tentatively planned to be EoL soon, and Jetty 12 requires a minimum of Java 17. Ticket filed as https://github.com/gwtproject/gwt/issues/10057.
[...]

PR #10075, opened 2024-12-23
JettyLauncher to allow all javax.sql.* classes to be loaded, Issue #9582

ian-stiles:
Issue #9582

Latest Jetty launcher with GWT and superdev mode fails without this when loading any classes related to JDBC access.

### DIFF
--- a/dev/core/src/com/google/gwt/dev/shell/jetty/JettyLauncher.java
+++ b/dev/core/src/com/google/gwt/dev/shell/jetty/JettyLauncher.java
@@ -295,6 +295,7 @@ public class JettyLauncher extends ServletContainerLauncher {
           // Xerces
           "org.apache.xerces.",
           "javax.xml.", // Used by Jetty for jetty-web.xml parsing
+          "javax.sql.*",
       });
 
       public WebAppClassLoaderExtension() throws IOException {


tbroyer:
javax.sql should be loaded from the WEB-INF/lib. Also, I'm not sure how it's related to the linked issue.

jnehlmeier:
@ian-stiles `JettyLauncher` is deprecated since GWT 2.11 and will eventually be removed in the future (see release notes of 2.11). This has been done because GWT DevMode still uses Jetty 9.4.44, `JettyLauncher` contains classloader hacks we do not want to maintain anymore and lastly because developers use the provided Jetty of DevMode as general purpose application server and regularly ask to update the version or make something work that `JettyLauncher` does not cover. 

Today we recommend to use `DevMode -noserver` or `CodeServer` (`DevMode -noserver` is basically the same as `CodeServer` mapping `DevMode` arguments to `CodeServer` arguments) and bring your own application server that you can configure as you like. The benefit is that it can be up-to-date and closely mimic your production server without any GWT specific classloader hacks.
Alternatively you can implement your own `ServletContainerLauncher` to start an application server of your choice via `DevMode`. You can then use your own version of `ServletContainerLauncher` using `DevMode -server com.example.MyCustomServletContainerLauncher`. 

So if you want to continue use `DevMode` to start your Jetty application server you should create your own little project, copying `JettyLauncher` and its dependencies to have a starting point and then adjust it as needed.

tbroyer:
@ian-stiles Ha ha, sorry, my Google fun failed me on mobile (and I so never/rarely directly use jdbc classes I forgot they came with the jdk)

That said, this is more a workaround than a fix, so I agree with Jens that you'd better create your own "launcher" removing all the classloader trickery that JettyLauncher inherited from legacy (or use a separate server)

jnehlmeier:
@ian-stiles 

> Wouldn't it help a large number of people hitting this issue if you just
added javax.sql to the allowed list since JettyLauncher is deprecated?

No, because it would mean that these people are then stuck on an old (unsupported) version of GWT once the new version does not ship JettyLauncher anymore and these people are then in the same situation as you now if they want to upgrade GWT: They have to provide their own implementation.

In addition to what I said before a larger reason is also the shift of javax.servlet to jakarta.servlet. While GWT 2.12 has an automatically generated gwt-servlet-jakarta.jar for server side classes provided by GWT, it is not so easy for DevMode. Jetty 12 might be an option but it requires Java 17 and GWT is just switching from 8 to 11. At the end it will be better to externalize the application server component of DevMode and let people choose which implementation they want to use, including their own. 

Starting with GWT 2.11 the SDK ships two implementations of `ServletContainerLauncher`: 

- JettyLauncher (javax.servlet based, deprecated)
- StaticResourceServer (new, can only serve static files)

Currently the proposed plan is that GWT 2.13 will either change its default to `StaticResourceServer` or force the developer to select an implementation when launching DevMode. GWT 2.13 and likely 2.14 will still ship `JettyLauncher` but you have to manually select it. After that `JettyLauncher` will likely be removed from the SDK.

See: 
- https://github.com/gwtproject/gwt/issues/10057
- https://github.com/gwtproject/gwt/issues/9863


From my own experience I can tell you that using `CodeServer` directly (instead of `DevMode`) and having a way to launch your own servlet container (local jetty installation, official or custom made jetty docker image) to deploy your *.war file to isn't too much of a burden. Once you have figured out a setup that fits your project you never have to think about server issues caused by GWT again. 
I think many people have migrated to Maven and use the gwt-maven-plugin along with a multi-module project, otherwise your linked issue that this PR wants to fix would have way more comments. 
That is why GWTs "Getting started" guide now also uses Maven: https://www.gwtproject.org/gettingstarted-v2.html

niloc132:
Are we happy to close this and expect it to be part of https://github.com/gwtproject/gwt/issues/10056?

eliasbalasis:
> Are we happy to close this and expect it to be part of #10056?

I believe we can close this, modifying JettyLauncher is a suitable workaround.
I also intend to give #10056 a try, depending on priorities and workload.

Issue #10065, opened 2024-12-17
 java.lang.NullPointerException: Cannot read field "declaringScope" because "outerLocalVariable" is null after upgrading from 2.11 to 2.12.1


koendecock:
GWT version:2.12.1 

After upgrading gwt in our application from 2.11 to 2.12.1 we get the below error during the gwt compilation process.
 java.lang.NullPointerException: Cannot read field "declaringScope" because "outerLocalVariable" is null

The full stack trace:
[INFO] [ERROR] Unexpected internal compiler error
[INFO] java.lang.NullPointerException: Cannot read field "declaringScope" because "outerLocalVariable" is null
[INFO]  at org.eclipse.jdt.internal.compiler.lookup.BlockScope.getEmulationPath(BlockScope.java:831)
[INFO]  at com.google.gwt.dev.jjs.impl.GwtAstBuilder$AstVisitor.pushLambdaExpressionLocalsIntoMethodScope(GwtAstBuilder.java:1264)
[INFO]  at com.google.gwt.dev.jjs.impl.GwtAstBuilder$AstVisitor.visit(GwtAstBuilder.java:1253)
[INFO]  at org.eclipse.jdt.internal.compiler.ast.LambdaExpression.traverse(LambdaExpression.java:787)
[INFO]  at org.eclipse.jdt.internal.compiler.ast.ExplicitConstructorCall.traverse(ExplicitConstructorCall.java:524)
[INFO]  at org.eclipse.jdt.internal.compiler.ast.ConstructorDeclaration.traverse(ConstructorDeclaration.java:704)
[INFO]  at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.traverse(TypeDeclaration.java:1831)
[INFO]  at org.eclipse.jdt.internal.compiler.ast.TypeDeclaration.traverse(TypeDeclaration.java:1683)
[INFO]  at com.google.gwt.dev.jjs.impl.GwtAstBuilder.processImpl(GwtAstBuilder.java:4122)
[INFO]  at com.google.gwt.dev.jjs.impl.GwtAstBuilder.process(GwtAstBuilder.java:4160)
[INFO]  at com.google.gwt.dev.javac.CompilationStateBuilder$CompileMoreLater$UnitProcessorImpl.process(CompilationStateBuilder.java:128)
[INFO]  at com.google.gwt.dev.javac.JdtCompiler$CompilerImpl.process(JdtCompiler.java:322)
[...]

PR #10080, opened 2025-01-08
Exceptions thrown when building GWT's AST should include context

niloc132:
Several visit and endVisit methods were missing their try/catch wrapping to translate any exception that occurs, with context of the given JDT ast node where the error took place.

Follow-up #10065


niloc132:
This PR is much easier to review if you enable the "Hide whitespace" feature in the changes tab
image

vegegoku (approved):

PR #10081, opened 2025-01-09
Update and improve help info logs

### LINKED ISSUE #10042: Update, migrate, improve log help links
**GWT version:** 2.12 and below
**Operating System:** any

---

##### Description
GWT's TreeLogger class is used in the compiler, and in Generators and Linkers, allowing library/framework code to also meaningfully contribute to log output. Usually, calls to the logger only use the `log(Type, String)` overload, but technically other overloads exist, including one that can provide an isntance of `com.google.gwt.core.ext.TreeLogger.HelpInfo`. This class is effectively just a wrapper around a URL.

There aren't many usages of this either, and most are out of date - in fact I didn't even remember that this feature existed until I was trying to clean up other usages of `com.google.gwt.util.tools.Utility` (which by its package should mostly be used in other cli tools? but instead is used in linkers, HelpInfo, and other assorted places).

Existing usages:

 * Concrete subclass `com.google.gwt.dev.util.InstalledHelpInfo`, takes a file name assumed to be in the GWT SDK zip. When the log message is written, verifies that the gwt-dev.jar came from such a zip, and finds the appropriate file, or skips linking at all. Since most GWT users these days download from Maven Central, this means the extra help is missing.
     * DevMode logs a link to https://github.com/gwtproject/gwt/blob/main/distro-source/core/src/doc/helpInfo/servletMappings.html with level DEBUG, so that if an error/warning happens when examining those tags, the user will have a helpful link to review. This document is very out of date.
     * JettyLauncher logs a link to https://github.com/gwtproject/gwt/blob/main/distro-source/core/src/doc/helpInfo/webAppClassPath.html when something wasn't in WEB-INF/lib that was accessed by something on the server classpath. Again, somewhat out of date (though less egregiously), but probably accurate enough to finish deprecating JettyLauncher until it is outside of gwt-dev itself.
     * JSORestrictionsChecker logs a link to https://github.com/gwtproject/gwt/blob/main/distro-source/core/src/doc/helpInfo/webAppClassPath.html for any failed restriction. This could easily be adapted to a section on https://www.gwtproject.org/doc/latest/DevGuideCodingBasicsJSNI.html or a related page.
     * JsniReferenceResolver logs a link to https://github.com/gwtproject/gwt/blob/main/distro-source/core/src/doc/helpInfo/webAppClassPath.html when a long value is accessible from JS. Again, could end up in the JSNI docs or a related page.
 * The only other link that exists is from BrowserChannelServer, to http://code.google.com/p/google-web-toolkit/wiki/UsingOOPHM (code.google.com is erroring out, so it apepars to be a dead link at this time). This would only be logged for an old browser plugin to a "recent" legacy dev mode server instance as a warning - probably can be safely removed.

Past releases had at least one other:
 * If GWT was found to be out of date, a link to release notes... with a null URL.

Proposed steps:

 * Create a directory specifically for "gwt-dev-help" in gwt-site, and create files there for help options as they are updated/created. Consider adding an anchor or queryparam including the GWT version, so that an anchor/etc can be added to the page with specifics for older versions (as the website might get updates as features change/improve) - this is the sort of feature we can't go back and add later when it matters.
 * Deprecate and replace InstalledHelpInfo with "GwtprojectHelpInfo". 
 * Migrate and update existing help text documents, fixing details that are no longer helpful, adding references to other pages.
 * Consider adding new documents, dealing with jsinterop, regressions, reporting bugs when they are found, etc.

vegegoku:
Update, migrate, improve log help links

Fix #10042
See also https://github.com/gwtproject/gwt-site/pull/386


dev/core/src/com/google/gwt/dev/util/GwtprojectOrgHelpInfo.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Reference to a help document on the GWT project website.
+ */
+public class GwtprojectOrgHelpInfo extends TreeLogger.HelpInfo {
+  private URL url;
+
+  public GwtprojectOrgHelpInfo(String relativeUrl) {
+    try {
+      url = new URL("https://gwtproject.org" + relativeUrl);
```
niloc132, line 31: All constructor calls seem to start with /doc/latest/gwt-dev-help/, can we include that in the path here?

If we ever want an absolute link to somewhere else, we could have a factory method for "help(pagename)" vs "article(path)"?

dev/core/src/com/google/gwt/dev/javac/JSORestrictionsChecker.java
```diff
@@ -288,8 +286,8 @@ static String errAlreadyImplemented(String intfName, String impl1,
 
   private static void errorOn(ASTNode node, CompilationUnitDeclaration cud,
       String error) {
-    GWTProblem.recordError(node, cud, error, new InstalledHelpInfo(
-        "jsoRestrictions.html"));
+    GWTProblem.recordError(node, cud, error, new GwtprojectOrgHelpInfo(
+        "/doc/latest/DevGuideCodingBasicsJSNI.html"));
```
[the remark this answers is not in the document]
vegegoku: The file was not moved as a file but as a section inside the DevGuideCodingBasicJSNI, but I now I change my commit to directly reference that section instead of the whole document and made it reflect the section title instead of the removed ffile.

dev/core/src/com/google/gwt/dev/javac/JSORestrictionsChecker.java
```diff
@@ -43,16 +43,14 @@
...
- *
+ * @see <a href="https://www.gwtproject.org/doc/latest/DevGuideCodingBasicsJSNI.html">jsoRestrictions.html</a>.
  * Any violations found are attached as errors on the
  * CompilationUnitDeclaration.
  *
  * @see <a
  *      href="http://code.google.com/p/google-web-toolkit/wiki/OverlayTypes">Overlay
  *      types design doc</a>
- * @see jsoRestrictions.html
+ * @see <a href="https://www.gwtproject.org/doc/latest/DevGuideCodingBasicsJSNI.html">jsoRestrictions.html</a>
```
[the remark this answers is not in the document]
vegegoku: same as above.

dev/core/src/com/google/gwt/dev/util/GwtprojectOrgHelpInfo.java
vegegoku: you are right I just did that

dev/core/src/com/google/gwt/dev/javac/JSORestrictionsChecker.java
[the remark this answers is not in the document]
niloc132: I don't think this absolute URL will work right when passed to GwtprojectOrgHelpInfo - maybe `GwtprojectOrgHelpInfo`'s ctor should have a check for a leading `/`, or make factory methods for "article links" vs "help links"?

dev/core/src/com/google/gwt/dev/util/GwtprojectOrgHelpInfo.java
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Reference to a help document on the GWT project website.
+ */
+public class GwtprojectOrgHelpInfo extends TreeLogger.HelpInfo {
+  private URL url;
+
+  public GwtprojectOrgHelpInfo(String relativeUrl) {
+    try {
+      url = new URL("https://gwtproject.org/doc/latest/gwt-dev-help/" + relativeUrl);
```
zbynek, line 46: Should we append `?version=2.13` as per Colin's suggestion in the issue description?

niloc132: I _think_ i'd vote an anchor rather than queryparam, since an anchor will work with just plain html, we won't need custom routing or whatever. On the other hand, if we add JS that helps handle this, we can read either queryparam or anchor?

Probably can most easily do this with About.getGwtVersion().

zbynek: With queryparam we could link to 
https://www.gwtproject.org/doc/latest/DevGuideCodingBasicsJSNI.html#important (if we wanted to)
directly by inserting the version param before the `#`. Also queryparam shows in server logs if you want to check how these pages are used. Both of these possible advantages are hypothetical and come at a cost of adding a bit of JS :shrug:

niloc132 (dismissed):

zbynek (dismissed):
Looks good, checkstyle warning might be good to resolve before merging.

niloc132 (approved):

PR #10082, opened 2025-01-09
Update JDT

### LINKED ISSUE #10026: Use JDT jars from maven central rather than packaging our own
Will help avoid issues like https://github.com/gwtproject/gwt/issues/10025.

A necessary component for #10011.

Light testing seems to suggest that 
```xml
            <dependency>
                <groupId>org.eclipse.jdt</groupId>
                <artifactId>org.eclipse.jdt.core</artifactId>
                <version>3.32.0</version>
            </dependency>
```
functions the same as our re-packaged jars, and repackaging may no longer be required at all.

vegegoku:
Use JDT jars from maven central rather than packaging our own

Fix #10026

### DIFF
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -29,7 +29,8 @@ jobs:
       - name: Checkout GWT tools into a sibling directory
         uses: actions/checkout@v4
         with:
-          repository: 'gwtproject/tools'
+          repository: 'vegegoku/tools'
+          ref: 'no-repackaging-jdt-test'
           path: 'tools'
       - name: Set up JDK ${{ matrix.java-version }}
         # GWT requires Java 11+ to build
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -29,8 +29,7 @@ jobs:
       - name: Checkout GWT tools into a sibling directory
         uses: actions/checkout@v4
         with:
-          repository: 'vegegoku/tools'
-          ref: 'no-repackaging-jdt-test'
+          repository: 'gwtproject/tools'
           path: 'tools'
       - name: Set up JDK ${{ matrix.java-version }}
         # GWT requires Java 11+ to build
--- a/maven/poms/gwt/gwt-dev/pom-template.xml
+++ b/maven/poms/gwt/gwt-dev/pom-template.xml
@@ -114,5 +114,9 @@
             <groupId>org.eclipse.jetty</groupId>
             <artifactId>apache-jsp</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.jdt</groupId>
+            <artifactId>org.eclipse.jdt.core</artifactId>
+        </dependency>
     </dependencies>
 </project>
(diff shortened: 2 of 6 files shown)


maven/poms/gwt/gwt-dev/pom-template.xml
```diff
@@ -114,5 +114,13 @@
             <groupId>org.eclipse.jetty</groupId>
             <artifactId>apache-jsp</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.eclipse.jdt</groupId>
+            <artifactId>org.eclipse.jdt.core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.eclipse.jdt</groupId>
+            <artifactId>ecj</artifactId>
```
niloc132, line 123: Why add ecj here? I could understand putting batch in here, but I think that's a transitive dependency of jdt-core - and the build.xml files etc above don't ever reference ecj, nor is it in the tools patch?

vegegoku: Actually batch is ecj, weird packaging from eclipse, but that is as you said is a transitive dependency and should not be included.

niloc132 (commented):
LGTM, just need to merge the tools PR first, so we can revert the CI change.

zbynek (dismissed):
Also fixes a problem where bad classpath configuration caused the transitively included JDT 3.19 to be used, resulting in
```
java.lang.ArrayIndexOutOfBoundsException: Index 922 out of bounds for length 870
        at org.eclipse.jdt.internal.compiler.parser.Parser.parse(Parser.java:13150)
        at org.eclipse.jdt.internal.compiler.parser.Parser.parse(Parser.java:13407)
        at org.eclipse.jdt.internal.compiler.parser.Parser.parse(Parser.java:13364)
        at com.google.gwt.dev.javac.JdtCompiler$ParserImpl.parse(JdtCompiler.java:210)
        at org.eclipse.jdt.internal.compiler.parser.Parser.dietParse(Parser.java:11750)
        at org.eclipse.jdt.internal.compiler.Compiler.internalBeginToCompile(Compiler.java:851)
        at org.eclipse.jdt.internal.compiler.Compiler.beginToCompile(Compiler.java:394)
        at org.eclipse.jdt.internal.compiler.Compiler.compile(Compiler.java:444)
        at org.eclipse.jdt.internal.compiler.Compiler.compile(Compiler.java:426)
 ``` 
 The dependency chain is
 ```
org.eclipse.jetty:apache-jsp:9.4.44.v20210927
    -> org.mortbay.jasper:apache-jsp:8.5.70
        -> org.eclipse.jdt:ecj:3.19.0
 ```

**Review by niloc132 — CHANGES_REQUESTED** [review]
Marking as "request changes", waiting for updates to the gwtproject/tools PR, then updating this to match.

File: .github/workflows/full-check.yml [hunk 2]
```diff
@@ -29,7 +29,8 @@ jobs:
       - name: Checkout GWT tools into a sibling directory
         uses: actions/checkout@v4
         with:
-          repository: 'gwtproject/tools'
+          repository: 'vegegoku/tools'
+          ref: 'no-repackaging-jdt-test'
```
niloc132, line 33: I think we just need to revert these temp workarounds and this should be ready for merge?

niloc132 (approved):

PR #10084, opened 2025-01-11
Remove Google poms

niloc132:
There will never be another Google release, so these poms (or the deploy script) no longer serve a purpose.


tbroyer (dismissed):
There's a reference to push-gwt.sh in maven/README.txt (that should probably be push-gwtproject.sh) but otherwise LGTM

niloc132:
> There's a reference to push-gwt.sh in maven/README.txt (that should probably be push-gwtproject.sh) but otherwise LGTM

Ah, thought we got all of those, missed the one in the prose (the code samples are right at least).

tbroyer (approved):

PR #10085, opened 2025-01-12
Remove JavaSourceParser

niloc132:
This class is only used by the typemodel for reading parameter names, but as written it can only ever return null - the parser is never populated with resources, so must always return null.


tbroyer (approved):

niloc132:
Passing nightly build, seeming to demonstrate that there's no compiler/generator/etc tests, and all samples build cleanly without this.

PR #10093, opened 2025-02-15
Mention source file name for JDT parsing exceptions

zbynek:
If JDT fails to parse a file, the thrown exception can be very cryptic. This at least adds the name of the source unit to the thrown exception.


niloc132:
Do you have an example stack trace that can cause this situation today?

niloc132 (approved):

Issue #10069, opened 2024-12-23
Tests: ClassPathEntryTest and ResourceAccumulatorTest fail on macOS


jnehlmeier:
##### Description

Both tests depend on JDK FileWatcher and macOS is known to have a bad implementation of it. It is relatively slow and resource intensive.

I am not yet sure if these tests are implemented to aggressively or if ClassPathEntry and ResourceAccumulator simply do not work correctly on macOS.

- ClassPathEntryTest
  - testResourceRenamed
  - testResourceDeleted
  - testResourceCreated
- ResourceAccumulatorTest
  - testListensInNewDirectories
  - testRenameParentDirectory
  - testRenameDirectory
  - testRenameFile
  - testSymlinkInfiniteLoop
  - testSymlinks
  - testMultipleListeners
  - testDeleteFile 

##### Steps to reproduce

Run dev tests via ANT on macOS.

jnehlmeier:
The following tests in `ClassPathEntryTest` are flaky as well. Sometimes they work, sometimes they don't. Seems like they don't work if they are executed last in the test run. `ResourceAccumulatorManager.getActiveListenerCount()` does not return the expected `0` in the failing case.

- testForResourceListenerLeaks_pathPrefixSetIsCollected
- testForResourceListenerLeaks_classPathEntryIsCollected

Issue #10067, opened 2024-12-23
Tests: DeadCodeEliminationTest.testFloatingPoint() depends on locale setting


jnehlmeier:
##### Description

`DeadCodeEliminationTest.testFloatingPoint()` fails if your default locale uses `,` instead of `.` in floating point numbers.

I don't like adding `-Duser.country=US -Duser.language=en` to `common.ant.xml -> gwt.junit` as I think tests should work without any special configuration, e.g. when running from the IDE directly without ANT.

Instead the test should use `DecimalFormatSymbols.getInstance().getDecimalSeparator()` to build the expected result when `String.format()` is used. 

##### Steps to reproduce

Run ant tests using `-Duser.country=DE -Duser.language=de` to simulate such a locale setting.

##### Known workarounds

Run ant tests using `-Duser.country=US -Duser.language=en`

niloc132:
Stack trace for this failure - make sure to edit the test jvm args and not ant's own args to reproduce:
```
Testcase: testFloatingPoint took 0,044 sec
	FAILED
return 1.1f; expected:<return 1[,]100000023841858;> but was:<return 1[.]100000023841858;>
junit.framework.ComparisonFailure: return 1.1f; expected:<return 1[,]100000023841858;> but was:<return 1[.]100000023841858;>
	at com.google.gwt.dev.jjs.impl.OptimizerTestBase$Result.intoString(OptimizerTestBase.java:140)
	at com.google.gwt.dev.jjs.impl.DeadCodeEliminationTest.testFloatingPoint(DeadCodeEliminationTest.java:468)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
```

The test is definitely at fault here, using Locale.ROOT or so resolves it.

PR #10097, opened 2025-02-19
Small patch, adding CharSequence::isEmpty

### LINKED ISSUE #10091: CharSequence isEmpty method is missing
Java 15 introduced `CharSequence#isEmpty()`: https://docs.oracle.com/en/java/javase/15/docs/api/java.base/java/lang/CharSequence.html#isEmpty()

When trying to use it in GWT like this:
```java
boolean empty = new StringBuilder().isEmpty();
```
The compiler throws the error:
```
The method isEmpty() is undefined for the type StringBuilder
```

I'm guessing it needs to be added to https://github.com/gwtproject/gwt/blob/main/user/super/com/google/gwt/emul/java/lang/CharSequence.java or something like that.

vjay82:
Fixes https://github.com/gwtproject/gwt/issues/10091

### DIFF
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -57,7 +57,7 @@ jobs:
             -Dtest.emma.htmlunit.disable=true
 
       - name: Report test results
-        uses: mikepenz/action-junit-report@v4.3.1
+        uses: mikepenz/action-junit-report@v5.4.0
         if: always()
         with:
           report_paths: 'gwt/build/out/**/test/**/reports/TEST-*.xml'
--- a/dev/core/src/com/google/gwt/dev/cfg/ConfigurationProperties.java
+++ b/dev/core/src/com/google/gwt/dev/cfg/ConfigurationProperties.java
@@ -142,6 +142,17 @@ public class ConfigurationProperties implements Serializable {
   }
 
   /**
+   * Returns all the values of a multi-valued configuration property, or null
+   * if not found.
+   *
+   * <p>A single-valued and unset configuration property will be returned as a list
+   * containing one null.
+   */
+  public List<String> getStringsOrNull(String key) {
+     return properties.get(key);
+  }
+
+  /**
    * Reads a configuration property as a comma-separated list of strings.
    * It may be a single-valued or multi-valued property. If multi-valued,
    * each value will be split on commas and all items concatenated.
--- a/dev/core/src/com/google/gwt/dev/javac/testing/impl/JavaResourceBase.java
+++ b/dev/core/src/com/google/gwt/dev/javac/testing/impl/JavaResourceBase.java
@@ -82,6 +82,7 @@ public class JavaResourceBase {
           "public interface CharSequence {",
           "  char charAt(int index);",
           "  int length();",
+		  "  boolean isEmpty();",
           "  String toString();",
           "}");
 
--- a/dev/core/src/com/google/gwt/dev/jjs/impl/ResolvePermutationDependentValues.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/impl/ResolvePermutationDependentValues.java
@@ -47,6 +47,7 @@ import com.google.gwt.thirdparty.guava.common.collect.Multimap;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 
 /**
@@ -74,13 +75,20 @@ public class ResolvePermutationDependentValues {
     }
 
     private JExpression propertyValueExpression(JPermutationDependentValue x) {
-      List<String> propertyValues = props.getConfigurationProperties().getStrings(x.getRequestedValue());
-
-      String propertyValue = propertyValues.isEmpty() ? null : Joiner.on(",").join(propertyValues);
-
-      if (propertyValue != null) {
+      List<String> propertyValues = props.getConfigurationProperties()
+          .getStringsOrNull(x.getRequestedValue());
+      if (propertyValues != null) {
         // It is a configuration property.
-        return program.getLiteral(x.getSourceInfo(), propertyValue);
+        // If no values are set, propertyValues is either empty (multivalued properties)
+        // or contains a single null (other properties).
+        if (propertyValues.stream().anyMatch(Objects::nonNull)) {
+          return program.getLiteral(x.getSourceInfo(),
+              Joiner.on(",").skipNulls().join(propertyValues));
+        }
+        if (x.getDefaultValueExpression() != null) {
+          return x.getDefaultValueExpression();
+        }
+        return program.getLiteralNull();
       }
 
       if (isSoftPermutationProperty(x.getRequestedValue())) {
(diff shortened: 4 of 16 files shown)


vjay82:
Ok, will have a look when some free time is at hand,

vjay82:
Hi, 

I implement the tests as instructed. 

Then I tried to let them run according to the instructions in the README.md.

Using Java 21:

```
set TZ=America/Los_Angeles
set ANT_OPTS=-Dfile.encoding=UTF-8
ant clean
ant user -Dtarget=test
...
[junit] Tests run: 5, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0,721 sec
[junit] Test com.google.gwt.dev.javac.PersistentUnitCacheTest FAILED
...
[junit] Tests run: 24, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 19,753 sec
[junit] Test com.google.gwt.dev.jjs.impl.DeadCodeEliminationTest FAILED
...
BUILD FAILED
```

My new` com.google.gwt.emultest.java15.lang.StringTest` was never called. (Wouldn't `CharSequenceTest` be a better name?)

I checked where `com.google.gwt.emultest.java11.lang.StringTest` might be called/linked and found `com.google.gwt.emultest.EmulJava11Suite`. 

I created a corresponding `EmulJava15Suite` but still no change. 

I also checked if `EmulJava11Suite` is refered to anywhere, which is not the case.

The java11 StringTest seems also not to be listed (probably not called?) in the output = ?

I set my test up to fail:

```
com.google.gwt.emultest.java15.lang.StringTest extends EmulTestBase {
	
  public void testIsEmpty() {
    // empty stub
    assertFalse(true);
  }

}
```
(Also broke `gwt\user\test-super\com\google\gwt\dev\jjs\super\com\google\gwt\dev\jjs\super\com\google\gwt\emultest\java15\lang\StringTest.java`)

The output of the ant-call was the same. 

I then tried to set up `...java11.lang.StringTest` to fail. 
Same output again.

Ich checked if `DeadCodeEliminationTest` (which is statet in the output) is referred to anywhere. It is not...

Finally, as I might have broken something, I reverted my changes back and ran the tests on the original repo:

```
set TZ=America/Los_Angeles
set ANT_OPTS=-Dfile.encoding=UTF-8
ant clean
ant user -Dtarget=test
...
[junit] Tests run: 5, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0,68 sec
[junit] Test com.google.gwt.dev.javac.PersistentUnitCacheTest FAILED
...
[junit] Tests run: 24, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 18,227 sec
[junit] Test com.google.gwt.dev.jjs.impl.DeadCodeEliminationTest FAILED
...
BUILD FAILED
```

Exactly the same thing.

I tried to run "all" tests without "user" prefix - to make sure every single one should be run. 
The command also fails, see [all tests.txt](https://github.com/user-attachments/files/19043314/all.tests.txt)

Any ideas?

niloc132:
Would you push the commit with the draft test so I can take a look there, give it a try?

The way I run tests is usually based on the table, cd'ing into the directory and running a task or two, like this:
```
cd user/
ant test.web.htmlunit '-Dgwt.junit.testcase.web.includes=**/CharSequenceTest.class'
```
that will run every test with that name in user/ with htmlunit, with a production build (rather than draft, or disabling class metadata) - for a simple emulation test like this, that should be sufficient.

With your own github fork, you can also push a branch name with a `-test` suffix, and enable github actions for your fork - the entire test suite will run. Name your branch something like charsequence-isempty-test and push that to your fork.

---
I'm very far from an ant expert, and as far as I can tell that `ant dev -Dtarget=test` is nonsense - if you specify another task in dev but not other subprojects (like buildtools), it will fail right away when buildtools can't use it. It seems to be running the `target` task everywhere...rather than just in the specified project.

As to why those tests are failing for you locally, I'm not sure, do you want to open a separate issue for that? I would guess there is a windows specific issue with how the tests are defined.

jnehlmeier:
`DeadCodeEliminationTest` is likely failing because of the Java default locale used on the system. There is a test that assumes US/en locale and it seems @vjay82 is located in Germany, like me. See: https://github.com/gwtproject/gwt/issues/10067

You have to run the tests using `-Duser.country=US -Duser.language=en`

jnehlmeier:
Also if you are using macOS some tests will simply fail, because Java on macOS has a bad implementation of `WatchService`. See: https://github.com/gwtproject/gwt/issues/10069

vjay82:
Hi, with 

`
cd user
ant test.web.htmlunit "-Dgwt.junit.testcase.web.includes=**/CharSequenceTest.class"
`
(it is " under Windows, otherwise straight into hell)

and a few other modifications I got it to run.

niloc132:
> and a few other modifications I got it to run. Only flaw still is, that the override CharSequenceTest is never used (breaking the test there has no effect). I didn't use your folder structure proposal because I found another one quite similar but not identical. Maybe you can have a look.


Thank you for keeping at this - there is a typo that prevents your test from working.

In the "regular" version of the CharSequenceTest, you placed it in `package com.google.gwt.emultest.java.lang;` However, your supersource version is in `package com.google.gwt.emultest.java15.lang;` - the difference in package (java vs java15) will mean that the supersource version will not be used when compiling to JS, so the test will never run.

vjay82:
I had to add the stub under com.google.gwt.emultest.java.lang as well.
The reason was, that the test runner complained the test not being referenced in EmulSuite.java.
I tried to include it in there but it resulted in a compilation error that the class file could not be found.
I looked at the other tests and discovered their files also exiting under com.google.gwt.emultest.java.lang . 
My conclusion was that these files probably get replaced when running under the appropriate Java version (however running your command line shows both test-files being executed.)

So you think moving the supersourced version to java.lang solves this? 
Then we can skip the java15 stub, no? 
(Wouldn't really matter as CharSequence supports the function independently anyway, right?)

vjay82:
Sorry, I didn't get what needs to be done now.

niloc132:
Your user/test-super/ implementation is in package `java15`. This is preferred to using the package `java`, which the stub in user/test/ is using. Move the stub to also use `java15`, and in turn, change the suite to also import the class from `java15`.

You've already updated the EmulSuite.gwt.xml to reference the java15 package as including sources, so that should be sufficient from what I can see here.

vjay82:
Hm, I'm pretty sure that didn't work. Thats why I copied the stub to java (without 15).
I will try again when time is available.

niloc132:
As above, we'd really prefer `java17` as the package (since we don't test with EOL JDKs). If you find it doesn't work with this package, please push what you have (to this or another branch) so we can help out, get this finished?

vjay82:
I moved the tests to the requested java17 - paths and broke the super implementation. 
Running the command still reports no errors. 
I tried to figure out where to put the test-super class by a bit of trial and error but couldn't find a location where the command would fail.

2025_03_29_13_26_24_CharSequenceTest java_Deleted_CharSequenceTest java_Deleted_gwt_Visual

zbynek:
https://github.com/gwtproject/gwt/pull/10106 has the super-sourced tests for Java 17 emulation.
Maybe I can include your changes in it so that we don't have to worry about conflicts?

vjay82:
@zbynek Yes, just go ahead 👍

zbynek:
Now part of #10106

PR #10100, opened 2025-02-26
Support JDT's implicit yield in switch statements

### LINKED ISSUE #10044: An single-assigment case-statement without {} within a switch generates a superfluous return, leading to function termination

niloc132:
JDT generates a yield for arrow cases if the arrow points at a expression, even though the expression cannot be returned. Synthesize a block wrapping that expression as a statement, and add a break to the end of it.

Fixes #10044
See #10059

### DIFF
--- a/user/test/com/google/gwt/dev/jjs/test/Java17Test.java
+++ b/user/test/com/google/gwt/dev/jjs/test/Java17Test.java
@@ -121,6 +121,9 @@ public class Java17Test extends GWTTestCase {
   public void testInlinedStringConstantsInCase() {
     assertFalse(isGwtSourceLevel17());
   }
+  public void testCaseArrowLabelsVoidExpression() {
+    assertFalse(isGwtSourceLevel17());
+  }
 
   private boolean isGwtSourceLevel17() {
     return JUnitShell.getCompilerOptions().getSourceLevel().compareTo(SourceLevel.JAVA17) >= 0;
--- a/dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/impl/GwtAstBuilder.java
@@ -547,7 +547,13 @@ public class GwtAstBuilder {
       try {
         SourceInfo info = makeSourceInfo(x);
         JExpression expression = pop(x.expression);
-        push(new JYieldStatement(info, expression));
+        if (x.switchExpression == null) {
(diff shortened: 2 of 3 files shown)


user/test/com/google/gwt/dev/jjs/test/Java17Test.java
```diff
@@ -121,6 +121,9 @@ public void testSwitchExprInlining() {
   public void testInlinedStringConstantsInCase() {
     assertFalse(isGwtSourceLevel17());
   }
+  public void testCaseArrowLabelsVoidExpression() {
```
zbynek, line 124: Missing newline above.

niloc132: Also missing in main (this is a backport for 2.12.2), so I'm somewhat unconcerned.

Related-ish, I will put up a PR soon that might get our checkstyle annotations back, again, in PRs...

vegegoku (approved):

zbynek (approved):

PR #10101, opened 2025-02-26
Fix Window.onClosing() not work in GWT 2.12

### LINKED ISSUE #10072: Window.onClosing() doesn't work in GWT 2.12
**GWT version: 2.12.0 and 2.12.1** 
**Browser (with version): Chrome** 
**Operating System: Windows 11** 

---

##### Description
com.google.gwt.user.client.Window.onClosing() method doesn't work after upgrading to GWT 2.12.x.
This is an issue introduced from https://github.com/gwtproject/gwt/pull/9984 in GWT 2.12.0.

When calling `Window.addWindowClosingHandler(ClosingHandler)`, it will set the new class member beforeCloseHandlersInitialized = true.
Unfortunately, onClosing() method is still checking closeHandlersInitialized. It should check the beforeCloseHandlersInitialized instead. See source code in https://github.com/gwtproject/gwt/blob/a48140fe3a57fb8975ccdd31acbcf9db0ebe1882/user/src/com/google/gwt/user/client/Window.java#L856
```java
  static String onClosing() {
    if (closeHandlersInitialized) {  // should be beforeCloseHandlersInitialized !!
      Window.ClosingEvent event = new Window.ClosingEvent();
      fireEvent(event);
      return event.getMessage();
    }
    return null;
  }
```

##### Steps to reproduce
Call `Window.addWindowClosingHandler(ClosingHandler);`.
When the window closes or redirects to other website, the Window#onClosing() or ClosingHandler#onWindowClosing method is not invoked.

##### Known workarounds
Have to temporarily call the deprecated method Window.addCloseHandler() to set the closeHandlersInitialized=true and then Window#onClosing() method will work. This is an ugly workaround.

niloc132:
When Window.onClosing(), if beforeCloseHandlersInitialized (not closeHandlersInitialized) is true, fire the ClosingEvent.

Fixes #10072 
See #10074

### DIFF
--- a/user/src/com/google/gwt/user/client/Window.java
+++ b/user/src/com/google/gwt/user/client/Window.java
@@ -853,7 +853,7 @@ public class Window {
   }
 
   static String onClosing() {
-    if (closeHandlersInitialized) {
+    if (beforeCloseHandlersInitialized) {
       Window.ClosingEvent event = new Window.ClosingEvent();
       fireEvent(event);
       return event.getMessage();


tbroyer:
Which issue or PR did you actually mean to reference in your PR description? 10074 maybe?

Otherwise LGTM

niloc132:
> Which issue or PR did you actually mean to reference in your PR description? 10074 maybe?

Yes, thank you - fixed.

tbroyer (approved):

vegegoku (approved):

PR #10102, opened 2025-02-26
Clean up file hash usage

### LINKED ISSUE #10090: Consolidate/cleanup hash functions
(Previous copy of this was eaten by the github issue management changeover, I'll probably continue to edit this and save it rather than wait until it is complete before submitting)

As part of #10029 the Util class is getting cleaned up - one of the functions of that class is to create "strong names" for various byte arrays throughout the codebase. This made me take a closer look at how hashing is used in the compiler and the server.

I'll be leaving the server and i18n alone for now to avoid risking changing any custom implementations or already generated constant keys. The gwt-rpc xsrf protection uses `Md5Utils`, and i18n uses `MD5KeyGenerator`. I am advocating removing some usages of `Md5Utils` and maybe moving/renaming that class, but for now will leave these usages alone.

There appear to be two hash algorithms used in GWT - md5 and sha1. None are used for true cryptographic purposes, and in all cases the developer/admin owns the content that would be hashed by them, or plaintext isn't available to a hypothetical attacker to find a collision. In all but one case, the JDK's own MessageDigest type is used, but one case uses Guava's hash functions.

Guava's documentation includes a summary of the provided hashing functions, generic enough to be useful here:

Image

SHA1 provides more bits which in theory will give better guarantees about collisions, but in practice the CompilerVersion use of sha1 will hash the gwt-dev.jar itself, then rehash that hash (with other inputs) into an md5. The unitCache files are also named with the sha1 of the compiler jar. The second usage of sha1 is hashing OperationKeys for requestfactory type tokens (both in APT and generators). The third is ModuleDef.java, hashing the filenames of all sources - but then only uses 32 of the 160 bits. As far as I can tell, in none of those cases are those extra bits necessary/helpful, there are unlikely to be collisions either way - I'm unclear why md5 wasn't considered sufficient - sha1 [hasn't been considered secure since 2005](https://en.wikipedia.org/wiki/SHA-1).

niloc132:
GWT currently uses a mix of md5 and sha1, though all use cases are non-cryptographic (and both hash functions are discouraged, and have been since before they were added to GWT). This patch proposes replacing them with Guava's murmur3 128 implementation, as it is slightly faster than md5 (and much faster than sha1), and can still be used to produce the same output format (16 upper-case hex digits).

As part of this refactor, some structured content now has size or field delimiters hashed to avoid collisions, and some streamed content is hashed as it is read to avoid holding the entire payload in memory at once. The Util method dealt with the size delimiters in some cases, but not uniformly, but also ignored exceptions - this patch also cleans up those errors to correctly throw with meaningful errors.

Fixes #10090


user/src/com/google/gwt/resources/rg/ExternalTextResourceGenerator.java
```diff
@@ -170,8 +172,8 @@ public void prepare(TreeLogger logger, ResourceContext context,
     offsets.put(method.getName(), hashes.get(toWrite));
   }
 
-  private String getMd5HashOfData() {
-    return Util.computeStrongName(Util.getBytes(data.toString()));
+  private String getMurmurHashOfData() {
```
niloc132, line 175: This method should be further renamed to be clear it is compute the hash again - or better, compute just once and memoize, as if `shouldUseJsonp()` is true we end up using it twice.

dev/core/src/com/google/gwt/dev/cfg/ModuleDef.java
```diff
@@ -455,31 +455,24 @@ public String getGwtXmlFilePath(String moduleName) {
...
-    filenames.addAll(gwtXmlPathByModuleName.values());
+  public long getInputFilenameHash() {
+    List<String> filenames = new ArrayList<>(gwtXmlPathByModuleName.values());
 
     for (Resource resource : getResourcesNewerThan(Integer.MIN_VALUE)) {
       filenames.add(resource.getLocation());
     }
 
-    // Take the first four bytes of the SHA-1 hash.
+    // Take the first eight bytes of the murmur3 hash.
```
niloc132, line 465: There's no reason that this method ever needed to return 4 bytes to begin with (especially of a sha1, which is 20 bytes wide). Even 8 bytes is a little silly of a 16 byte murmur, since the Recompiler.InputSummary could just as easily hold a string of the hash. The InputSummary only exists until the next recompile to quickly check if the inputs are the same or different.

I'm going to go ahead and change this all to use the full string. I don't understand how this code came about to begin with.

niloc132 (commented):
Quick self review, will resolve these and merge conflicts.

niloc132 (commented):
Another pass at self-review complete, I'm removing the "draft". As far as I can tell, the three mentioned changes have the risk of breaking the compiler if an individual user makes a change locally that causes a hash collision. The likelihood for this is very low even for a non-cryptographically secure hash like murmur3 - if we were distributing partly compiled files between developers (like git commits, etc) I might be more conservative with this change.

That said, I'm definitely open to feedback on reverting back to md5 for at least these three changes, and possibly having a single class that picks the hash function for the rest of the compiler, so that a project/team can decide for themselves to pay a little more for compilation in exchange for longer/better hashes.

Besides those three, the full build seems to still pass with collisions in all other calls to produce hashes - that is, I replaced the result in all other places with `"1234567890ABCDEF"`. This might point to insufficient tests - more research here would be a good idea long term.

Regardless of what hashing algorithm is used, I think there is value in this patch in making sure we don't make multiple copies of data just to hash it, but hash as we read/write.

zbynek (approved):

vegegoku (approved):

PR #10103, opened 2025-02-27
Change default SCL instance, and allow selecting from service loader

### LINKED ISSUE #10057: Change default dev mode server to static implementation

### LINKED ISSUE #10058: Consider service loader for picking up custom implementations
NOTE: This is specific to ServletContainerLauncher, see #10235 for RunStyle.

The `-runStyle` and `-server` flags of JUnitShell and DevMode respectively take an optional classname, which can be irritating to fully populate. The `-server` flag defaults to using a built in class without naming it, `com.google.gwt.dev.shell.JettyLauncher` (see https://github.com/gwtproject/gwt/issues/10057 to consider changing that default), while the `-runStyle` flag supports `Manual`, `Selenium` (hugely out of date), `HtmlUnit`, and `ExternalBrowser` (officially unsupported, but will attempt to run if referenced by name).

Both of these features are worthy to extend and make more useful, but can be irritating to enter a fully qualified class name, especially when canonical implementations are supported by simple name.

This ticket proposes to introduce an optional factory type for this sort of feature, which would allow a create(...) method to be defined on it, and would already have a `String getName()` method, something like

```java
public interface NamedFactory {
  String getName();
}
```
Subtypes of this would add their create() method, and implementations of the factory could then be picked up by a ServiceLoader call, and tested to ensure their names don't collide. A user-specified name then can be used in addition to the fully qualified class name (which would continue to work to avoid the factory where one doesn't exist or where there is a collision).

Example (written without the aid of an IDE, missing duplicate checking, exception handling, etc):
New class to add within GWT itself, to load by name as needed
```java
public interface RunStyleFactory extends NamedFactory {
  static RunStyle create(JUnitShell shell, String name) {
    for (RunStyleFactory f : ServiceLoader.load(RunStyleFactory.class)) {
      if (f.getName().equals(name)) {
        return f.create(shell);
      }
    }
    return Class.forName(name);
  }
  
  RunStyle create(JUnitShell shell);
}
```
Example factory provided in a library for WebDriver support (something like https://github.com/gwtproject/gwt-core/pull/14/files):
```java
public class WebDriverRunStyleFactory implements RunStyleFactory {
  @Override
  public String getName() {
    return "WebDriver";
  }
  @Override
  public WebDriverRunStyle create(JUnitShell shell) {
    return new WebDriverRunStyle(shell);
  }
}
```

In addition to making it easier to provide new implementations without a fully.qualified.ClassName on the command line, it will also make it easier to move standard implementations out of the gwt-dev.jar classpath - decoupling jetty versions (htmlunit uses jetty 9's websocket client, and dev mode presently defaults to jetty 9, see #10057).

Command line usage then can look like `-runStyle WebDriver:http://localhost:4444/?firefox`.

niloc132:
The default is now to ask a service loader for a ServletContainerLauncher if there is a single entry available, and failing that to use the static server - no servlets will be available.

ServletContainerLaunchers registered as a provider for the service loader must provide a name that matches a specific regex. Those names may now also be used when specifying a particular server name rather than the fully qualified class name.

The JettyLauncher will be kept for at least another version to ensure that projects can easily migrate to newer versions.

Fixes #10057
Fixes #10058


eliasbalasis:
I am aligned with this approach, particularly in relation to the custom launchers project at https://bitbucket.org/upperlimit-public/gwt-devmode-server of which I am the owner and I will make the relevant adjustments.

eliasbalasis:
Should we allow hyphens as well as underscores in the launcher name?

niloc132:
> Should we allow hyphens as well as underscores in the launcher name ? or perhaps only hyphens ? What do you think?

Sure - I kept the regex simple, just what things are allowed for 99% of class names. I mostly don't want to encourage named types, which could be referenced from command line, to include characters that need to be escaped when used from the command lie (` `, `*`, etc).

eliasbalasis:
Thanks @niloc132 , this makes sense, I will name my launchers accordingly.

dev/core/src/com/google/gwt/dev/DevMode.java
```diff
@@ -122,24 +125,48 @@ public boolean setFlag(boolean value) {
...
       if (options.isNoServer()) {
         return null;
       } else {
-        return new String[] {getTag(), DEFAULT_SCL};
+        if (registered.size() == 1) {
+          // Exactly one registered SCL, use it as the default, by fully qualified class name
+          return new String[] {
+              getTag(),
+              registered.values().iterator().next().getClass().getName()
+          };
```
vegegoku: Would it make sense here to log that we either found none or found multiple launchers and that is why we revert to the default one.

zbynek: Maybe just return empty string here and make sure `""` checks for service loader first in `setString`?

dev/core/src/com/google/gwt/dev/DevMode.java
zbynek, line 168: Maybe this should be more specific, e.g. "Specify a fully qualified class name or a service name of a different [...]".

Further details can be added to https://www.gwtproject.org/doc/latest/DevGuideCompilingAndDebugging

dev/core/src/com/google/gwt/dev/DevMode.java
niloc132: @zbynek the difference is that getDefaultArgs() is only called if the user _didn't_ pass `-server ...` of some kind - that's the only time we want the service loader to kick in, and only then if there is exactly one.

The "conflict" is that there is no way to pass `-server` without an argument. The closest I can get to looks like `-server ''` as a way to explicitly trigger "I want a server, but i refuse to tell you which one".

The server is enabled by default for DevMode, so a user has these options:
 * Don't specify anything, get the default
 * Use `-noserver` to turn it off
 * Specify `-server ...` explicitly to name the server they want (by class or by given name)
 * ... the weird other "i want to tell you that I want a server even though that's the default `-server ''`.

I think this is worth doing because of the existing weird code that allows for this, I just find it a little hard to follow as a reader, I'd rather the "default" loaded server be set in getDefaultArgs() than in setString?

---

@vegegoku No service loaded launchers is the default, so I don't think we want extra logging there? Once the server starts, you can see which one you got too, in the tabs at the top
image
I did add a logged message for multiple service loaded instances with none specified.

dev/core/src/com/google/gwt/dev/DevMode.java
niloc132: Added, PTAL. Also opened a draft doc PR to provide more details (and encourage more use of CodeServer) https://github.com/gwtproject/gwt-site/pull/418

dev/core/src/com/google/gwt/dev/util/arg/ArgHandlerWarDir.java
```diff
@@ -45,6 +45,11 @@ public String getTag() {
     return "-war";
   }
 
+  @Override
+  public String[] getTags() {
+    return new String[] {getTag(), "-launcherDir"};
```
niloc132, line 50: I can revert this if we don't think it makes sense, but it makes CodeServer and DevMode just a little more alike.

zbynek (dismissed):
Approving code changes, didn't test.

vegegoku (dismissed):

niloc132:
Test sample: https://github.com/niloc132/gwt-devmode-server-sample
Build this repo, and you can add it to any project just by putting the newly built launcher jar on the classpath to swap in Jetty 12.1 support with EE11. Supporting EE8 instead (e.g. javax.servlet) requires changing some code in the launcher and its dependencies (and so should be its own distributed jar) will probably be needed for existing projects that don't intend to update a newer EE in the near future.

niloc132:
Added a change to restore samples that require a working server by explicitly setting `-server` for the legacy JettyLauncher. This should be temporary, removed when https://github.com/gwtproject/gwt/issues/10232 is fixed.

user/src/com/google/gwt/user/tools/WebAppCreator.java
```diff
@@ -141,6 +142,29 @@ public boolean setString(String str) {
     }
   }
 
+  private final class ArgHandlerLegacyJetty extends ArgHandlerFlag {
+    @Override
+    public String getPurposeSnippet() {
+      return "DEPRECATED: Use deprecated JettyLauncher instead of an external server provided server impl";
```
zbynek: Shouldn't "external server provided server impl" be "external server impl provided as service" or maybe just "external server"?

niloc132: Agreed, I can word that better. I like "external server". Hopefully the flag won't live long enough or even get used enough to ever cause any confusion.

user/src/com/google/gwt/user/tools/WebAppCreator.java
```diff
@@ -505,6 +529,7 @@ private static String replaceFileName(Map<String, String> replacements, String n
   private boolean onlyEclipse;
   private File outDir;
   private boolean overwrite = false;
+  private String server = "";
```
zbynek, line 532: Maybe this should not be empty to make sure it's quoted correctly when Ant builds the command line.

niloc132: This does seem to work as is - if you don't set the new/deprecated flag, you just get "-server" "" in your xml as expected. The `-server` arg requires a value, even if it is the empty string (and the WebAppCreator "templating" doesn't support conditionals, just global find/replace).

zbynek: I thought this was the reason for the failed test

```
Not creating tests because -junit argument was not specified.

'unknown_parameter' does not appear to be a valid fully-qualified Java class name
....
Missing required argument 'moduleName'
```

Not worried about the XML being well-formed, but about how Ant interprets `""` ins XML.

niloc132: Hmm. Wonder why it worked locally... I'll go back and re-verify.

zbynek: As discussed privately the empty string is not the issue, but the naming of the variable is, because substitution is based on `String.replace`. The full build foor https://github.com/zbynek/gwt/commit/f18c9cc50a5464cdf46bcaf1d7e2dbba9aec3cc3 is green, feel free to merge/cherry-pick to this branch.

zbynek (approved):

PR #10106, opened 2025-03-12
Java 12-17API additions

### LINKED ISSUE #9872: New API emulation from Java version 12 - 17
The following are some  parts of the API changes between java 12 - 17 that we might want  to emulate:

 #### Java 12

- [ ] String.transform
- [ ] String.indent
- [ ] Collectors.teeing 

#### Java 16

- [ ] Stream.mapMulti
- [ ] Stream.toList

#### Java 17

- [ ] HexFormat

Related to https://github.com/gwtproject/gwt/issues/9869

### LINKED ISSUE #9991: Implement (or document) missing CharSequence methods
Java 8 added [CharSequence.codePoints()](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/CharSequence.html#codePoints()).

Java 11 added [CharSequence.compare()](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/CharSequence.html#compare(java.lang.CharSequence,java.lang.CharSequence)).

### LINKED ISSUE #10091: CharSequence isEmpty method is missing

zbynek:
Fixes #9872
Fixes #10091
Fixes #9991


user/test-super/com/google/gwt/dev/jjs/super/com/google/gwt/emultest/java12/lang/StringTest.java
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Tests for java.lang.String Java 12 API emulation.
+ */
+public class StringTest extends EmulTestBase {
+
+  public void testTransform() {
+    assertEquals(3, hideFromCompiler("foo").transform(String::length));
+  }
+
+  public void testIndent() {
```
niloc132, line 32: please also include a set of these tests that does not do a hideFromCompiler, so we can make sure that if the compiler is able to constant-fold these method calls, that it does so correctly.

niloc132 (commented):
Nice work on the draft - some quick feedback

user/super/com/google/gwt/emul/java/util/stream/Stream.java
```diff
@@ -246,6 +251,10 @@ public boolean tryAdvance(Consumer<? super T> action) {
     return StreamSupport.stream(spliterator, false);
   }
 
+  default List<T> toList() {
+    return Collections.unmodifiableList(this.collect(Collectors.toList()));
```
natros, line 255: The official documentation states that "The implementation in this interface returns a List produced as if by the following"

```java
Collections.unmodifiableList(new ArrayList<>(Arrays.asList(this.toArray())))
```
I'm not saying that your version is wrong. I just like to know if it is better or more efficient.
https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/util/stream/Stream.html#toList()

niloc132: I also was going to comment on either this or using our existing Collectors.toUnmodifiableList() (note that I'm wrong: that will not accept nulls, but Stream.toList() is permitted to hold nulls... but has the other same semantics about immutability).

To your question though, StreamImpl.toArray already calls collect(toList()):
https://github.com/gwtproject/gwt/blob/a5e649dec66f65f4e1cd0ecd7bffd3e706243750/user/super/com/google/gwt/emul/java/util/stream/StreamImpl.java#L567-L570

Effectively, we would be then building a list, to build an array, to wrap in a list, to copy to another list. This implementation is better, simply creating and appending to a list, then wrapping to make it unmodifiable.

zbynek:
Full build https://github.com/zbynek/gwt/actions/runs/14282864925

user/test-super/com/google/gwt/dev/jjs/super/com/google/gwt/emultest/java12/lang/StringTest.java
zbynek: I did not implement any tests without `hideFromCompiler` yet since I wasn't sure which methods need them. It seems that tests either hide all inputs from compiler (like the ones from https://github.com/gwtproject/gwt/pull/9975/files#diff-79013db0b72ddfb9522327d93e6c5804f44c160025590e971ddfc7fd4eb7c045R26 ) or hide nothing (https://github.com/zbynek/gwt/blob/f0c1486a46018aec9de93458136cfc161f569e81/user/test/com/google/gwt/emultest/java/lang/StringTest.java#L542 ). Also I'm not sure if it wouldn't be better to test the constant-folding optimization and verify that these are actually folded rather than make assertions where we don't know if we're testing folding or emulation.

user/super/com/google/gwt/emul/java/util/stream/LongStream.java
```diff
@@ -364,5 +375,13 @@ public boolean tryAdvance(LongConsumer action) {
     return StreamSupport.longStream(spliterator, false);
   }
 
+  default LongStream mapMulti(LongStream.LongMapMultiConsumer mapper) {
+    return flatMap(element -> {
+      List<Long> buffer = new ArrayList<>();
```
niloc132: What do you think about using LongStream.Builder instead of this arraylist?

For Double it matters a little less, since a double and a Double are represented the same in GWT, but Integer is an object wrapper on a js number, and some Longs then are an object wrapper on an object wrapper on 3 js numbers, so it does start to add up.

For the LongConsumer lambda, we even save the cost of creating the method reference (presently in GWT, that means create an entirely new class with a constructor, ctm, etc), but LongStream.Builder already directly implements LongConsumer, so we'd get that "for free". Then, we'd just call build() instead of streaming the arraylist and mapping back to long (another lambda, more expensive than it looks).

The Object version would still save on the cost of the lambda (assuming anything else in the compiled app ever used Stream.Builder) - but I think I'd vote make this enhancement for all the implementations?

user/test/com/google/gwt/emultest/java17/util/MapEntryTest.java
```diff
@@ -0,0 +1,34 @@
...
+package com.google.gwt.emultest.java17.util;
+
+import com.google.gwt.dev.util.arg.SourceLevel;
+import com.google.gwt.emultest.java.util.EmulTestBase;
+import com.google.gwt.junit.DoNotRunWith;
+import com.google.gwt.junit.JUnitShell;
+import com.google.gwt.junit.Platform;
+
+@DoNotRunWith(Platform.Devel)
+public class MapEntryTest extends EmulTestBase {
```
niloc132, line 28: I dont think this got added to the suite?

user/test/com/google/gwt/emultest/java17/lang/StringTest.java
```diff
@@ -0,0 +1,45 @@
...
+import com.google.gwt.emultest.java.util.EmulTestBase;
+import com.google.gwt.junit.DoNotRunWith;
+import com.google.gwt.junit.JUnitShell;
+import com.google.gwt.junit.Platform;
+
+/**
+ * Tests for java.lang.String Java 12 - 17 API emulation.
+ */
+@DoNotRunWith(Platform.Devel)
+public class StringTest extends EmulTestBase {
```
niloc132, line 28: missing testStripIndent

user/test-super/com/google/gwt/emultest/super/com/google/gwt/emultest/java17/util/stream/DoubleStreamTest.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public void testMapMulti() {
+    DoubleStream.DoubleMapMultiConsumer doubling = (num, callback) -> {
+      callback.accept(num);
+      callback.accept(num + num);
+    };
+    assertEquals(new double[0],
+        hideFromCompiler(DoubleStream.of()).mapMulti(doubling).toArray());
+    assertEquals(new double[]{1.0, 2.0, 3.0, 6.0},
+        hideFromCompiler(DoubleStream.of(1.0, 3.0)).mapMulti(doubling).toArray());
```
niloc132: I don't think we need to hide DoubleStreams/etc from the compiler, as they aren't js literals, and can't be compiled into some other literal?

gwt.iml
```diff
@@ -0,0 +1,97 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
niloc132: delete this file (from the branch at least)?

user/super/com/google/gwt/emul/java/lang/String.java
```diff
@@ -488,6 +491,7 @@ public String intern() {
     return checkNotNull(this);
   }
 
+  @Override
   public boolean isEmpty() {
     return length() == 0;
   }
```
niloc132: Actually I think we can just remove this method, since its the same as its supertype? Same as codePoints.

zbynek:
https://github.com/zbynek/gwt/actions/runs/16452377759 is 🟢 , @niloc132 thanks for the feedback

user/super/com/google/gwt/emul/java/util/BitSet.java
```diff
@@ -21,6 +21,10 @@
 import javaemul.internal.ArrayHelper;
 import javaemul.internal.LongUtils;
 
+import java.util.function.IntConsumer;
+import java.util.stream.IntStream;
```
niloc132, line 22: checkstyle import complaints above ^

user/super/com/google/gwt/emul/java/util/stream/Collectors.java
```diff
@@ -27,13 +27,15 @@
...
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.StringJoiner;
 import java.util.function.BiConsumer;
 import java.util.function.BinaryOperator;
+import java.util.function.BiFunction;
 import java.util.function.Function;
 import java.util.function.Predicate;
```
niloc132, line 40: More checkstyle import tweaks ^

user/super/com/google/gwt/emul/java/util/BitSet.java
```diff
@@ -37,6 +41,36 @@ public class BitSet {
...
+    }
+
+    @Override
+    public long estimateSize() {
+      return size();
+    }
+
+    @Override
+    public int characteristics() {
+      return Spliterator.SIZED | Spliterator.ORDERED | Spliterator.DISTINCT | Spliterator.SORTED;
```
niloc132: I believe that at least in GWT, if it is SIZED, it is also always SUBSIZED.

user/super/com/google/gwt/emul/java/util/BitSet.java
```diff
@@ -604,6 +638,11 @@ public int nextSetBit(int fromIndex) {
     return bitIndex(index) + Integer.numberOfTrailingZeros(word);
   }
 
+  public IntStream stream() {
+    Spliterator.OfInt spliterator = new BitSetSpliterator();
+    return StreamSupport.intStream(spliterator, false);
+  }
+
```
niloc132, line 645: Just a note, I don't think this is actionable, but maybe there's a way for this to be handled. From the javadoc:
> The stream binds to this bit set when the terminal stream operation commences (specifically, the spliterator for the stream is [late-binding](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Spliterator.html#binding)). If the bit set is modified during that operation then the result is undefined.

However, the late-binding link goes on to say
> A late-binding Spliterator binds to the source of elements at the point of first traversal, first split, or first query for estimated size, rather than at the time the Spliterator is created. A Spliterator that is not late-binding binds to the source of elements at the point of construction or first invocation of any method. Modifications made to the source prior to binding are reflected when the Spliterator is traversed. After binding a Spliterator should, on a best-effort basis, throw [ConcurrentModificationException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/ConcurrentModificationException.html) if structural interference is detected. Spliterators that do this are called fail-fast. The bulk traversal method ([forEachRemaining()](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Spliterator.html#forEachRemaining(java.util.function.Consumer))) of a Spliterator may optimize traversal and check for structural interference after all elements have been traversed, rather than checking per-element and failing immediately.

In the JVM, if you create a bitset, set one value, then get `stream().spliterator()`, it will report size of one. Then, if you set another bit or clear the existing, the spliterator will continue to report size of one. In contrast, your impl will return 1, then 2 (or zero).

Using tryAdvance with these mutations (in the JVM) will not throw CME as described, but will only return the one item. Your implementation will end up returning the second (or just stopping after the first).

I don't think this is bad, given "the result is undefined", just should be called out to discuss.

user/test-super/com/google/gwt/emultest/super/com/google/gwt/emultest/java17/lang/StringTest.java
```diff
@@ -0,0 +1,64 @@
...
+  public void testTranslateEscapes() {
+    assertEquals("\b \r\n\t\f",
+        hideFromCompiler("\\b\\s\\r\\n\\t\\f").translateEscapes());
+    assertEquals("\u0001\u0011\u00d1",
+        hideFromCompiler("\\1\\21\\321").translateEscapes());
+    assertEquals("\u00ff\u001f8\u00200",
+        hideFromCompiler("\\377\\378\\400").translateEscapes());
+    assertEquals("ab",
+        hideFromCompiler("a\\\nb").translateEscapes());
+  }
```
niloc132, line 86: I think your implementation correctly handles `"\\\r"` and `"\\\n"` by returning `""`, but I think it fails on `"\\\r\n"`, which should return `""`, but instead returns `"\n"`. 

For `"\\\n\r"` it appears to correctly return `"\r"` - `\r\n` technically falls into the `<line-terminator>` category, despite being two separate characters https://docs.oracle.com/javase/specs/jls/se17/html/jls-3.html#jls-LineTerminator

user/super/com/google/gwt/emul/java/lang/String.java
```diff
@@ -839,11 +967,11 @@ public boolean tryAdvance(Consumer<? super String> action) {
       }
       return nextIndex < length();
     }
+  }
 
-    private int cappedIndexOf(char c) {
-      int index = indexOf(c, nextIndex);
-      return index == -1 ? length() : index;
-    }
+  private int cappedIndexOf(char c, int nextIndex) {
```
niloc132: Why change this signature, and reference a local instead of a field?

gwt.iml
zbynek: oops, thanks.

user/test-super/com/google/gwt/dev/jjs/super/com/google/gwt/emultest/java12/lang/StringTest.java
niloc132: Using hideFromCompiler ensures that the emulated sources provided can be called and function the way that the test expects. Passing a constant and allowing constant folding means that the operation will be performed by the compiler itself, using the JRE's own implementation, which does two things for us: we know that constant folding doesn't somehow break with this call, and that the JRE implementation doesn't behave in a way that fails the test.

This isn't a guarantee - the constant folding mechanism is pretty specific (I've been meaning to file a bug about that) and won't always kick in, but it does let us get a little more coverage.

user/super/com/google/gwt/emul/java/lang/String.java
zbynek: good point, not needed, reverting (I wanted to use it from `String` directly, but in the end I managed without it).

user/super/com/google/gwt/emul/java/util/BitSet.java
zbynek: Thanks for pointing this out. I'd like to avoid adding concurrent modification detection, since it makes code more complex and JVM is not doing that anyway. If you have any other suggestion how to bring the behavior closer to JVM's, please let me know.

niloc132: I think you have room in the spec for your implementation to function as it is written today, so I'm not worried about it, just want to be sure we had this note for later if it ever came up.

user/super/com/google/gwt/emul/java/lang/String.java
```diff
@@ -488,10 +491,6 @@ public String intern() {
     return checkNotNull(this);
   }
 
-  public boolean isEmpty() {
-    return length() == 0;
```
niloc132, line 492: far-off future nit: this is going to end up being inlined as something like
`<expr>.length==0`, and promoting this to the supertype (which, i think i suggested) will mean we keep getting that. But, it would be smaller to just implement as `<expr>==""` in the cases where we know we have a java.lang.String, since java String is final+immutable, and js string can be `==`/`===` compared. I think we could also live with `!<expr>.length`, but that's still longer regardless of == or === usage.

user/super/com/google/gwt/emul/java/lang/String.java
```diff
@@ -811,6 +814,131 @@ private int getTrailingWhitespaceLength() {
     return length;
   }
 
+  public String indent(int spaces) {
+    if (spaces  == 0) {
```
niloc132: checkstyle ignores supersource, for better or worse.
```suggestion
    if (spaces == 0) {
```

EDIT: actually this is wrong, javadoc specifies that line endings are rewritten

tools/api-checker/config/gwt212_213userapi.conf
```diff
@@ -156,6 +156,7 @@ java.lang.Deprecated MISSING
 java.lang.FunctionalInterface MISSING
 java.lang.Override MISSING
 java.lang.SafeVarargs MISSING
+java.lang.String::isEmpty() MISSING
```
niloc132, line 160:
```suggestion
# Moved to CharSequence as a default method
java.lang.String::isEmpty() MISSING
```

dev/core/src/com/google/gwt/dev/javac/testing/impl/JavaResourceBase.java
```diff
@@ -82,6 +82,7 @@ public class JavaResourceBase {
           "public interface CharSequence {",
           "  char charAt(int index);",
           "  int length();",
+          "  default boolean isEmpty() { return length() == 0; }",
```
niloc132, line 85: should be safe to delete from line 348 now too

user/super/com/google/gwt/emul/java/lang/String.java
```diff
@@ -811,6 +814,131 @@ private int getTrailingWhitespaceLength() {
...
+    } else {
+      indentedLines = lines().map(
+          line -> line.substring(Math.min(-spaces, line.getLeadingWhitespaceLength())));
+    }
+    return indentedLines.collect(Collectors.joining("\n", "", "\n"));
+  }
+
+  public String stripIndent() {
+    if (isEmpty()) {
+      return "";
```
niloc132, line 834: technically this could also be `return this;`, since the "isEmpty" check is "are you the empty string". Might result in making the code bigger though, but probably will be smaller - String is final so all methods are made static and `this` is an argument (so will be rewritten to argument `a` or something) - one char instead of two.

I'm not 100% confident in that analysis, so maybe don't change this just from my guesswork, just trying to think through what JS we'll get.

user/test-super/com/google/gwt/emultest/super/com/google/gwt/emultest/java17/lang/StringTest.java
```diff
@@ -0,0 +1,68 @@
...
+  public void testIndent() {
+    assertEquals("  x\n", hideFromCompiler("x").indent(2));
+    assertEquals("x\n", hideFromCompiler("  x").indent(-2));
+    assertEquals("  x\n  y\n", hideFromCompiler("x\ny").indent(2));
+    assertEquals("  x\n  y\n", hideFromCompiler("x\r\ny").indent(2));
+    assertEquals("  x\n  y\n", hideFromCompiler("x\ry").indent(2));
+    assertEquals("x\ny\n", hideFromCompiler("  x\n  y").indent(-2));
+    assertEquals("x\ny\n", hideFromCompiler("  x\r\n  y").indent(-2));
+    assertEquals("x\ny\n", hideFromCompiler("  x\r  y").indent(-2));
+  }
```
niloc132, line 54: From https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#indent(int), we should have a few more tests:
 * validate that `\t` is removed as if it was a single space
 * validate that indent(0) changes line endings to `\n` - this will fail
 * generally validate that `\r\n` is rewritten to `\n` (edit: this is covered for positive and negative already I see)
 * validate that negative indent with fewer spaces just removes what spaces are available - and that indent with too many spaces leaves extra spaces.
 * validate that the empty string is handled correctly, and other strings with no line break - I _think_ this will fail
 * validate trailing/leading linebreak in the string

user/super/com/google/gwt/emul/java/lang/String.java
```diff
@@ -811,6 +814,131 @@ private int getTrailingWhitespaceLength() {
...
+      return this;
+    }
+    Stream<String> indentedLines;
+    if (spaces > 0) {
+      String spaceString = " ".repeat(spaces);
+      indentedLines = lines().map(line -> spaceString + line);
+    } else {
+      indentedLines = lines().map(
+          line -> line.substring(Math.min(-spaces, line.getLeadingWhitespaceLength())));
+    }
```
niloc132, line 828: For a little more code, this could be a little faster in the `spaces == 0` case, just assign `indentedLines = lines()` with no calls to getLeadingWhitespaceLength(), bounds checks, or min() tests for each line. I think we would rather the code you have now though, as that seems a little silly to do, but just pointing it out.
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#indent(int)

It would be good to have tests that validate this too, added a note below for this.

user/test-super/com/google/gwt/emultest/super/com/google/gwt/emultest/java17/lang/StringTest.java
```diff
@@ -0,0 +1,68 @@
...
+    assertEquals("x", hideFromCompiler("x").stripIndent());
+    assertEquals("x", hideFromCompiler("  x").stripIndent());
+    assertEquals("x\n", hideFromCompiler("x\r\n").stripIndent());
+    assertEquals("x\ny", hideFromCompiler("  x\n  y").stripIndent());
+    assertEquals("x\ny", hideFromCompiler("  x\r\n  y").stripIndent());
+    assertEquals(" x\ny", hideFromCompiler("  x\n y").stripIndent());
+    assertEquals("x\n y", hideFromCompiler(" x\n  y").stripIndent());
+    assertEquals("x\n\ny", hideFromCompiler("  x\n      \n  y").stripIndent());
+    assertEquals(" x\ny", hideFromCompiler("\t x\r\n y").stripIndent());
+  }
```
niloc132, line 71: I think this should have a handful of more tests too:
 * `"  x\n. "` to confirm that the second line is entirely removed
 * `"  x\n\n  y"` to confirm blank lines do not have their indent counted
 * ...except the last line, where a blank line's indent is meaningful. Combining the above two notes, something like `"    x\n  "` should yield `"  x\n"` (and `"  x\n    "` should have the same result)

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/lang/String.html#stripIndent()

user/super/com/google/gwt/emul/java/lang/String.java
zbynek: So should we go with `Js.isFalsy(length())` here?
If we care about `foo.isEmpty()` throwing for `foo == null`, `!foo.length` is IMHO the best we can do in terms of character count.
If we don't, `!foo` is the shortest (so `Js.isFalsy(toString())` could be the default implementation).

user/super/com/google/gwt/emul/java/lang/String.java
zbynek: I merged it to the `spaces > 0` case, adding empty string should be easier to optimize out than counting the leading space and getting substrings.

user/super/com/google/gwt/emul/java/lang/String.java
niloc132: Yeah I don't think we do anything at all right now - I'm mostly thinking about what could happen to improve all the little checks where the compiler has already decided that a given value can't be null (e.g. the developer wrote `str != null || str.isEmpty()`, or the compiler has already proven that nulls will never be passed as an argument, etc).

But definitely not for this patch, it is no worse than it was when you started afaict.

user/super/com/google/gwt/emul/java/lang/String.java
zbynek: Thanks for the suggestion. I think for now I'll mark this as resolved and if someone wants to check which option produces smaller output, they can do so in a follow-up PR.

user/test-super/com/google/gwt/emultest/super/com/google/gwt/emultest/java17/util/stream/CollectorsTest.java
```diff
@@ -0,0 +1,37 @@
...
+package com.google.gwt.emultest.java17.util.stream;
+
+import com.google.gwt.emultest.java.util.EmulTestBase;
+
+import java.util.stream.Collector;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ * Tests for java.lang.String Java 12 API emulation.
```
niloc132:
```suggestion
 * Tests for java.util.stream.Collectors Java 12 API emulation.
```

user/test-super/com/google/gwt/emultest/super/com/google/gwt/emultest/java17/lang/CharSequenceTest.java
```diff
@@ -0,0 +1,28 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.gwt.emultest.java17.lang;
+
+import com.google.gwt.emultest.java.util.EmulTestBase;
+
+public class CharSequenceTest extends EmulTestBase {
```
niloc132, line 23:
```suggestion
/**
 * Tests for Java 17 java.lang.CharSequence emulation.
 */
public class CharSequenceTest extends EmulTestBase {
```

user/test-super/com/google/gwt/emultest/super/com/google/gwt/emultest/java17/util/MapEntryTest.java
```diff
@@ -0,0 +1,34 @@
...
+ * the License.
+ */
+package com.google.gwt.emultest.java17.util;
+
+import com.google.gwt.emultest.java.util.EmulTestBase;
+
+import java.util.AbstractMap;
+import java.util.Map;
+
+public class MapEntryTest extends EmulTestBase {
```
niloc132, line 26: Checkstyle doesnt run (...yet?) on emulation, but we'll have to add these at some point, so I stubbed them in (but some versions might need a tweak).
```suggestion
/**
 * Tests for Java 17 java.util.Map.Entry emulation.
 */
public class MapEntryTest extends EmulTestBase {
```

user/test-super/com/google/gwt/emultest/super/com/google/gwt/emultest/java17/util/stream/DoubleStreamTest.java
```diff
@@ -0,0 +1,36 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.gwt.emultest.java17.util.stream;
+
+import com.google.gwt.emultest.java.util.EmulTestBase;
+
+import java.util.stream.DoubleStream;
+
+public class DoubleStreamTest extends EmulTestBase {
```
niloc132, line 25:
```suggestion
/**
 * Tests for Java 17 java.util.stream.DoubleStream emulation.
 */
public class DoubleStreamTest extends EmulTestBase {
```

user/test-super/com/google/gwt/emultest/super/com/google/gwt/emultest/java17/util/stream/IntStreamTest.java
```diff
@@ -0,0 +1,37 @@
...
+ * the License.
+ */
+package com.google.gwt.emultest.java17.util.stream;
+
+import com.google.gwt.emultest.java.util.EmulTestBase;
+
+import java.util.stream.IntStream;
+import java.util.stream.Stream;
+
+public class IntStreamTest extends EmulTestBase {
```
niloc132, line 26:
```suggestion
/**
 * Tests for Java 17 java.util.stream.IntStream emulation.
 */
public class IntStreamTest extends EmulTestBase {
```

user/test-super/com/google/gwt/emultest/super/com/google/gwt/emultest/java17/util/stream/LongStreamTest.java
```diff
@@ -0,0 +1,36 @@
...
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.gwt.emultest.java17.util.stream;
+
+import com.google.gwt.emultest.java.util.EmulTestBase;
+
+import java.util.stream.LongStream;
+
+public class LongStreamTest extends EmulTestBase {
```
niloc132, line 25:
```suggestion
/**
 * Tests for Java 17 java.util.stream.LongStream emulation.
 */
public class LongStreamTest extends EmulTestBase {
```

user/test-super/com/google/gwt/emultest/super/com/google/gwt/emultest/java17/util/stream/StreamTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+import java.util.Arrays;
+import java.util.function.BiConsumer;
+import java.util.function.Consumer;
+import java.util.function.DoubleConsumer;
+import java.util.function.IntConsumer;
+import java.util.function.LongConsumer;
+import java.util.stream.Stream;
+
+public class StreamTest extends EmulTestBase {
```
niloc132, line 31:
```suggestion
/**
 * Tests for Java 17 java.util.stream.Stream emulation.
 */
public class StreamTest extends EmulTestBase {
```

user/test/com/google/gwt/emultest/java/lang/CharSequenceTest.java
```diff
@@ -0,0 +1,45 @@
...
+ * the License.
+ */
+package com.google.gwt.emultest.java.lang;
+
+import com.google.gwt.emultest.java.util.EmulTestBase;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class CharSequenceTest extends EmulTestBase {
```
niloc132, line 26:
```suggestion
/**
 * Tests for Java 17 java.lang.CharSequence emulation.
 */
public class CharSequenceTest extends EmulTestBase {
```

user/test/com/google/gwt/emultest/java17/lang/CharSequenceTest.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package com.google.gwt.emultest.java17.lang;
+
+import com.google.gwt.dev.util.arg.SourceLevel;
+import com.google.gwt.emultest.java.util.EmulTestBase;
+import com.google.gwt.junit.DoNotRunWith;
+import com.google.gwt.junit.JUnitShell;
+import com.google.gwt.junit.Platform;
+
+@DoNotRunWith(Platform.Devel)
```
niloc132, line 27:
```suggestion
/**
 * Tests for Java 17 java.lang.CharSequence emulation.
 */
@DoNotRunWith(Platform.Devel)
```

user/test/com/google/gwt/emultest/java17/util/MapEntryTest.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package com.google.gwt.emultest.java17.util;
+
+import com.google.gwt.dev.util.arg.SourceLevel;
+import com.google.gwt.emultest.java.util.EmulTestBase;
+import com.google.gwt.junit.DoNotRunWith;
+import com.google.gwt.junit.JUnitShell;
+import com.google.gwt.junit.Platform;
+
+@DoNotRunWith(Platform.Devel)
```
niloc132, line 27:
```suggestion
/**
 * Tests for Java 17 java.util.Map.Entry emulation.
 */
@DoNotRunWith(Platform.Devel)
```

user/test/com/google/gwt/emultest/java17/util/stream/DoubleStreamTest.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package com.google.gwt.emultest.java17.util.stream;
+
+import com.google.gwt.dev.util.arg.SourceLevel;
+import com.google.gwt.emultest.java.util.EmulTestBase;
+import com.google.gwt.junit.DoNotRunWith;
+import com.google.gwt.junit.JUnitShell;
+import com.google.gwt.junit.Platform;
+
+@DoNotRunWith(Platform.Devel)
```
niloc132, line 27:
```suggestion
/**
 * Tests for java.util.stream.Collectors Java 17 emulation.
 */
@DoNotRunWith(Platform.Devel)
```

user/test/com/google/gwt/emultest/java17/util/stream/IntStreamTest.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package com.google.gwt.emultest.java17.util.stream;
+
+import com.google.gwt.dev.util.arg.SourceLevel;
+import com.google.gwt.emultest.java.util.EmulTestBase;
+import com.google.gwt.junit.DoNotRunWith;
+import com.google.gwt.junit.JUnitShell;
+import com.google.gwt.junit.Platform;
+
+@DoNotRunWith(Platform.Devel)
```
niloc132, line 27:
```suggestion
/**
 * Tests for java.util.stream.IntStream Java 17 emulation.
 */
@DoNotRunWith(Platform.Devel)
```

user/test/com/google/gwt/emultest/java17/util/stream/LongStreamTest.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+package com.google.gwt.emultest.java17.util.stream;
+
+import com.google.gwt.dev.util.arg.SourceLevel;
+import com.google.gwt.emultest.java.util.EmulTestBase;
+import com.google.gwt.junit.DoNotRunWith;
+import com.google.gwt.junit.JUnitShell;
+import com.google.gwt.junit.Platform;
+
+@DoNotRunWith(Platform.Devel)
```
niloc132, line 27:
```suggestion
/**
 * Tests for java.util.stream.LongStream Java 17 emulation.
 */
@DoNotRunWith(Platform.Devel)
```

user/test/com/google/gwt/emultest/java17/util/stream/DoubleStreamTest.java
```diff
@@ -0,0 +1,37 @@
...
+package com.google.gwt.emultest.java17.util.stream;
+
+import com.google.gwt.dev.util.arg.SourceLevel;
+import com.google.gwt.emultest.java.util.EmulTestBase;
+import com.google.gwt.junit.DoNotRunWith;
+import com.google.gwt.junit.JUnitShell;
+import com.google.gwt.junit.Platform;
+
+/**
+ * Tests for java.util.stream.Collectors Java 17 emulation.
```
niloc132:
```suggestion
 * Tests for java.util.stream.DoubleStream Java 17 emulation.
```

this was my fault, sorry

niloc132 (approved):

PR #10108, opened 2025-03-27
Improving accessability - added aria presentation role for layout tables

Lonzak:
According to accessibility rules layout tables exist merely to position content visually and should not be used in HTML5. 
Screen readers may interpret them as data tables (i.e., announcing column and row numbers) etc). To change that in GWT would be a bigger task. So as a minor improvement those tables should be marked as layout tables. This can be done by assigning the role="presentation" to ensure it is not identified as a table to screen reader users.

### DIFF
--- a/user/src/com/google/gwt/user/client/ui/CellPanel.java
+++ b/user/src/com/google/gwt/user/client/ui/CellPanel.java
@@ -72,6 +72,8 @@ public abstract class CellPanel extends ComplexPanel {
 
   public CellPanel() {
     table = DOM.createTable();
+    //aria role to indicate it is a plain layout table
+    table.setAttribute("role", "presentation");
     body = DOM.createTBody();
     DOM.appendChild(table, body);
     setElement(table);
--- a/user/src/com/google/gwt/user/client/ui/CellPanel.java
+++ b/user/src/com/google/gwt/user/client/ui/CellPanel.java
@@ -72,7 +72,7 @@ public abstract class CellPanel extends ComplexPanel {
 
   public CellPanel() {
     table = DOM.createTable();
-    //aria role to indicate it is a plain layout table
+    // aria role to indicate it is a plain layout table
     table.setAttribute("role", "presentation");
     body = DOM.createTBody();
     DOM.appendChild(table, body);
--- a/user/src/com/google/gwt/user/client/ui/CellPanel.java
+++ b/user/src/com/google/gwt/user/client/ui/CellPanel.java
@@ -15,6 +15,7 @@
  */
 package com.google.gwt.user.client.ui;
 
+import com.google.gwt.aria.client.Roles;
 import com.google.gwt.dom.client.Element;
 import com.google.gwt.user.client.DOM;
 import com.google.gwt.user.client.ui.HasHorizontalAlignment.HorizontalAlignmentConstant;
@@ -73,7 +74,7 @@ public abstract class CellPanel extends ComplexPanel {
   public CellPanel() {
     table = DOM.createTable();
     // aria role to indicate it is a plain layout table
-    table.setAttribute("role", "presentation");
+    Roles.getPresentationRole().set(table);
     body = DOM.createTBody();
     DOM.appendChild(table, body);
     setElement(table);
--- a/user/src/com/google/gwt/user/client/ui/HTMLTable.java
+++ b/user/src/com/google/gwt/user/client/ui/HTMLTable.java
@@ -15,6 +15,7 @@
  */
 package com.google.gwt.user.client.ui;
 
+import com.google.gwt.aria.client.Roles;
 import com.google.gwt.core.client.JsArray;
 import com.google.gwt.core.shared.GWT;
 import com.google.gwt.dom.client.Document;
@@ -788,8 +789,8 @@ public abstract class HTMLTable extends Panel implements SourcesTableEvents,
    */
   public HTMLTable() {
     tableElem = DOM.createTable();
-    //aria role to indicate it is a plain layout table
-    tableElem.setAttribute("role", "presentation");
+    // aria role to indicate it is a plain layout table
+    Roles.getPresentationRole().set(tableElem);
     bodyElem = DOM.createTBody();
     DOM.appendChild(tableElem, bodyElem);
     setElement(tableElem);
--- a/user/src/com/google/gwt/user/client/ui/HTMLTable.java
+++ b/user/src/com/google/gwt/user/client/ui/HTMLTable.java
@@ -788,6 +788,8 @@ public abstract class HTMLTable extends Panel implements SourcesTableEvents,
    */
   public HTMLTable() {
     tableElem = DOM.createTable();
+    //aria role to indicate it is a plain layout table
+    tableElem.setAttribute("role", "presentation");
     bodyElem = DOM.createTBody();
     DOM.appendChild(tableElem, bodyElem);
     setElement(tableElem);
--- a/user/src/com/google/gwt/user/client/ui/HTMLTable.java
+++ b/user/src/com/google/gwt/user/client/ui/HTMLTable.java
@@ -789,8 +789,6 @@ public abstract class HTMLTable extends Panel implements SourcesTableEvents,
    */
   public HTMLTable() {
     tableElem = DOM.createTable();
-    // aria role to indicate it is a plain layout table
-    Roles.getPresentationRole().set(tableElem);
     bodyElem = DOM.createTBody();
     DOM.appendChild(tableElem, bodyElem);
     setElement(tableElem);
--- a/user/src/com/google/gwt/user/client/ui/HTMLTable.java
+++ b/user/src/com/google/gwt/user/client/ui/HTMLTable.java
@@ -15,7 +15,6 @@
  */
 package com.google.gwt.user.client.ui;
 
-import com.google.gwt.aria.client.Roles;
 import com.google.gwt.core.client.JsArray;
 import com.google.gwt.core.shared.GWT;
 import com.google.gwt.dom.client.Document;
--- a/user/test/com/google/gwt/user/client/ui/FlexTableTest.java
+++ b/user/test/com/google/gwt/user/client/ui/FlexTableTest.java
@@ -150,5 +150,11 @@ public class FlexTableTest extends HTMLTableTestBase {
     cellFormatter.setHeight(3, 1, "300px");
     cellFormatter.setColSpan(3, 1, 2);
   }
+  
+  public void testPresentationRole() {
+    FlexTable t = new FlexTable();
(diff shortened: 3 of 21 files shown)


user/src/com/google/gwt/user/client/ui/CellPanel.java
```diff
@@ -72,6 +72,8 @@ public abstract class CellPanel extends ComplexPanel {
 
   public CellPanel() {
     table = DOM.createTable();
+    //aria role to indicate it is a plain layout table
```
niloc132, line 75: checkstyle is going to require a space after the `//` and before the comment text
```suggestion
    // aria role to indicate it is a plain layout table
```

user/src/com/google/gwt/user/client/ui/CellPanel.java
```diff
@@ -72,6 +72,8 @@ public abstract class CellPanel extends ComplexPanel {
...
+    table.setAttribute("role", "presentation");
```
niloc132, line 76: Instead of setting attributes directly, consider using the gwt aria package like other classes do to assign their roles, see TabBar, CellTree, etc.
```suggestion
    Roles.getPresentationRole().set(table);
```

Leaving tests as you have them seems to make sense, we want to verify that the dom is as expected. It could also make sense to query the dom in each of these cases for _all_ tables within the current parent, and validate that each has a role set (at a quick glance, in all cases you've modified we should have the "presentation" role assigned, I don't see nested elements which should actually be presented as tables).

niloc132 (commented):
Thanks for the patch, a few thoughts:
 * consider spelling out UA? In the context of browsers and JS, it usually means "user agent", and I think you probably mean something else like usability/accessibility? Substituting "user agent"(e.g. "the browser") doesn't make sense in your title or description, and I don't see a obvious definition searching for the term in the context of [dom](https://www.google.com/search?hl=en&q=dom%20ua) and [aria](https://www.google.com/search?hl=en&q=aria%20ua).

niloc132:
> UA stands for Universal Accessibility but as always abbreviations mean something different in different contexts. Just accessibility is fine. Strange with the blank after comment // I did run 'ant compile.tests checkstyle' as it is described in the readme.md and everything was fine. But no problem - I fixed it.
> 
> I also changed the code to use 'Roles' class which is much better. Thanks for the pointer.

I think the "space after `//`" shows up as a warning, not an error, so it won't kill your build, but it will get logged if you go hunting for it. The CI run will log it at least (annotations are currently broken..), but still need a human to point it out.

Thanks!

user/src/com/google/gwt/user/client/ui/HTMLTable.java
```diff
@@ -788,8 +789,8 @@ protected void setAttr(int row, String attrName, String value) {
    */
   public HTMLTable() {
     tableElem = DOM.createTable();
-    //aria role to indicate it is a plain layout table
-    tableElem.setAttribute("role", "presentation");
+    // aria role to indicate it is a plain layout table
+    Roles.getPresentationRole().set(tableElem);
```
zbynek, line 793: In the other cases it's clear that the table is just for presentation, here it's not obvious, people might actually use this to display tabular data. Since users can easily add the role when needed, I would suggest leaving `HTMLTable` and its subclasses like `Grid` as is.

Lonzak: Interesting - for me Grid and Flextable was only for layouting (... and in most cases it is probably used for this) if a data table is needed one should use CellTable or DataGrid. But sure there is no restriction and e.g.  FlexTable could be manually extended with <th>...

jnehlmeier: This should be removed. 

CellTable and DataGrid are advanced widgets that had been introduced as a faster alternative to Grid/FlexTable, especially for slow browsers like IE 6 (or if the data table is really large). For smaller/normal data tables using Grid / FlexTable is the way to go. Of course you can always misuse them to do pure layout but then you have to set the aria role yourself.

jnehlmeier (changes requested):

user/src/com/google/gwt/user/client/ui/HTMLTable.java
Lonzak: Ok removed the role setting for HTMLTable

zbynek (dismissed):

user/src/com/google/gwt/user/client/ui/HTMLTable.java
```diff
@@ -15,6 +15,7 @@
  */
 package com.google.gwt.user.client.ui;
 
+import com.google.gwt.aria.client.Roles;
```
jnehlmeier, line 18: Remove the unused import

user/test/com/google/gwt/user/client/ui/FlexTableTest.java
```diff
@@ -150,5 +150,4 @@ public void secondarySetHeightTest() {
     cellFormatter.setHeight(3, 1, "300px");
     cellFormatter.setColSpan(3, 1, 2);
   }
```
jnehlmeier, line 152: Add the newline again. Ideally HTMLTable and FlexTableTest should have no changes in this PR.

jnehlmeier (changes requested):
Just two small things, otherwise it looks good.

jnehlmeier (dismissed):

zbynek (dismissed):

niloc132 (changes requested):
Looks like this is failing tests (verified in github actions by pushing to `aria-test` on my fork):
```
    [junit] Running com.google.gwt.user.UiSuite
    [junit] Tests run: 1008, Failures: 1, Errors: 38, Skipped: 0, Time elapsed: 143.649 sec
    [junit] Test com.google.gwt.user.UiSuite FAILED
```

https://github.com/niloc132/gwt/actions/runs/14194633693
One of the stack traces:
```
java.lang.AssertionError: Element cannot be null.
	at java.lang.Throwable.Throwable(Throwable.java:73)
	at java.lang.Error.Error(Error.java:30)
	at java.lang.AssertionError.AssertionError(AssertionError.java:51)
	at com.google.gwt.aria.client.RoleImpl.set(RoleImpl.java:219)
	at com.google.gwt.user.client.ui.TreeItem$TreeItemImpl.initializeClonableElements(TreeItem.java:92)
	at com.google.gwt.user.client.ui.TreeItem$TreeItemImpl.TreeItem$TreeItemImpl(TreeItem.java:65)
	at com.google.gwt.user.client.ui.TreeItem.$clinit(TreeItem.java:255)
	at com.google.gwt.user.client.ui.TreeItem.TreeItem(TreeItem.java:307)
	at com.google.gwt.user.client.ui.Tree.init(Tree.java:1088)
	at com.google.gwt.user.client.ui.Tree.Tree(Tree.java:204)
	at com.google.gwt.uibinder.test.client.UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl$Widgets.build_myTree(UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl.java:2162)
	at com.google.gwt.uibinder.test.client.UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl$Widgets.get_myTree(UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl.java:2158)
	at com.google.gwt.uibinder.test.client.UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl$Widgets.build_f_HTMLPanel1(UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl.java:1350)
	at com.google.gwt.uibinder.test.client.UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl$Widgets.get_f_HTMLPanel1(UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl.java:1209)
	at com.google.gwt.uibinder.test.client.UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl$Widgets.build_root(UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl.java:1078)
	at com.google.gwt.uibinder.test.client.UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl$Widgets.get_root(UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl.java:1070)
	at com.google.gwt.uibinder.test.client.UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl.createAndBindUi(UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl.java:185)
	at com.google.gwt.uibinder.test.client.UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl.createAndBindUi(UiBinderJsInteropTest_MyWidgetBasedUi_BinderImpl.java:182)
	at com.google.gwt.uibinder.test.client.UiBinderJsInteropTest$MyWidgetBasedUi.init(UiBinderJsInteropTest.java:40)
	at com.google.gwt.uibinder.test.client.WidgetBasedUi.WidgetBasedUi(WidgetBasedUi.java:277)
	at com.google.gwt.uibinder.test.client.UiBinderJsInteropTest$MyWidgetBasedUi.UiBinderJsInteropTest$MyWidgetBasedUi(UiBinderJsInteropTest.java:30)
	at com.google.gwt.uibinder.test.client.UiBinderJsInteropTest.testJsElementTypes(UiBinderJsInteropTest.java:45)
```

From looking at TreeItem.java, you are calling `Roles.getPresentationRole().set(BASE_BARE_ELEM)` before `BASE_BARE_ELEM` was first assigned in `initializeClonableElements()`.

I did not check other stack traces.

niloc132 (changes requested):
Tests still seem to be failing
https://github.com/niloc132/gwt/actions/runs/14759920702

It appears that the tests aren't testing what they should be - for example, menubar.init() is creating a `<table>` and appending to the outer `<div>` which already (correctly?) has the `menubar` role applied? Instead, you want to be finding the table tag _inside_ the div, and testing that.

I'd suggest running the tests locally, or turn on actions on your fork so that you get these results yourself (since you're pushing to main).

The quickest test command locally will be to `cd user` and run `ant test.draft.htmlunit '-Dgwt.junit.testcase.includes=**/UiSuite.class'`, to just run the suite that handles all of the tests you've changed.

zbynek:
@Lonzak please merge https://github.com/Lonzak/gwt/pull/1 to fix the tests

Lonzak:
sorry for the delay ...working on the junit part (running the test takes 4-6hours) so had to post pone this

zbynek (approved):

Lonzak:
Ok so everything should be ok now...?

niloc132 (approved):
Approved, once we get nightlies back on track let's land this.

Passing nightly build: https://github.com/niloc132/gwt/actions/runs/16323934154

PR #10113, opened 2025-04-06
Improve Character emulation

### LINKED ISSUE #9705: Character.toString(...) results in ERROR (compile)
**GWT version:** 2.9.0
**Browser (with version):** Firefox
**Operating System:** Linux Ubuntu Mate

---

##### Description
Compiling:
```java
@Override
	public CompletableFuture<SubscriberListResult> loadSubscribers(final String from, final String until) {
		CallbackFuture<SubscriberListResult> cf = new CallbackFuture<>();
		final Token param = state.token();
		REST.withCallback(cf.callback()).call(rest).getSubscribers(
				(from == null || from == "") ? Character.toString(Character.MIN_CODE_POINT) : from,
				(until == null || until == "") ? Character.toString(Character.MAX_CODE_POINT) : until, param);
		return cf.future();
	}
```

results in:

```java
GET /recompile/gwt
   Job com.newsrx.ButterAdmin_1_12
      starting job: com.newsrx.ButterAdmin_1_12
      binding: locale=default
      binding: user.agent=safari
      Compiling module com.newsrx.ButterAdmin
         Ignored 1 unit with compilation errors in first pass.
Compile with -strict or with -logLevel set to TRACE or DEBUG to see all errors.
         [ERROR] Errors in 'file:/home/michael/git/ButterAdmin/ButterAdmin-client/src/main/java/com/newsrx/client/app/model/ApiImpl.java'
            [ERROR] Line 482: The method toString(Object) from the type Object is not visible
            [ERROR] Line 483: The method toString(Object) from the type Object is not visible
         Tracing compile failure path for type 'com.newsrx.client.app.model.ApiImpl'
            [ERROR] Errors in 'file:/home/michael/git/ButterAdmin/ButterAdmin-client/src/main/java/com/newsrx/client/app/model/ApiImpl.java'
               [ERROR] Line 482: The method toString(Object) from the type Object is not visible
               [ERROR] Line 483: The method toString(Object) from the type Object is not visible
```

zbynek:
Fixes #9705 
Fixes #1989 (not 100% of methods are covered, but some are just impossible without bundling unicode database into compiled code)

Implements several APIs using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape that should be well supported in all browsers.

The implementation is mostly similar to https://groups.google.com/g/google-web-toolkit-contributors/c/73-aScAShs4/m/gZAhlUXiBAAJ , but the fallbacks for old Edge versions are not included.

### DIFF
--- a/user/super/com/google/gwt/emul/java/lang/CaseMapper.java
+++ b/user/super/com/google/gwt/emul/java/lang/CaseMapper.java
@@ -23,7 +23,23 @@ final class CaseMapper {
   }
 
   public static char charToUpperCase(char c) {
-    return String.valueOf(c).toUpperCase().charAt(0);
+    String upper = String.valueOf(c).toUpperCase();
+    return hasExtraCodePoints(upper) ? c : upper.charAt(0);
+  }
+
+  public static int intToLowerCase(int codePoint) {
+    return String.NativeString.fromCodePoint(codePoint).toLowerCase().codePointAt(0);
+  }
+
+  public static int intToUpperCase(int codePoint) {
+    String upper = String.NativeString.fromCodePoint(codePoint).toUpperCase();
+    return hasExtraCodePoints(upper) ? codePoint : upper.codePointAt(0);
+  }
+
+  // If String.toUpperCase produces more than 1 codepoint, Character.toUpperCase should
+  // act either as identity or title-case conversion (not supported in GWT).
+  private static boolean hasExtraCodePoints(String str) {
+    return str.asNativeString().codePointAt(1) > 0;
   }
 
   private CaseMapper() {}
--- a/user/super/com/google/gwt/emul/java/lang/Character.java
+++ b/user/super/com/google/gwt/emul/java/lang/Character.java
@@ -301,10 +301,19 @@ public final class Character implements Comparable<Character>, Serializable {
   private static NativeRegExp isLeterOrDigitRegex;
 
   public static boolean isLetterOrDigit(char c) {
+    return isLetterOrDigit(String.valueOf(c));
+  }
+
+  public static boolean isLetterOrDigit(int codePoint) {
+    return isValidCodePoint(codePoint)
+        && isLetterOrDigit(String.NativeString.fromCodePoint(codePoint));
+  }
+
+  private static boolean isLetterOrDigit(String str) {
     if (isLeterOrDigitRegex == null) {
       isLeterOrDigitRegex = new NativeRegExp("[\\p{Nd}\\p{L}]", "u");
     }
-    return isLeterOrDigitRegex.test(String.valueOf(c));
+    return isLeterOrDigitRegex.test(str);
   }
 
   private static NativeRegExp lowerCaseRegex;
--- a/user/super/com/google/gwt/emul/java/lang/Character.java
+++ b/user/super/com/google/gwt/emul/java/lang/Character.java
@@ -18,14 +18,25 @@ package java.lang;
 import static javaemul.internal.InternalPreconditions.checkCriticalArgument;
 
 import java.io.Serializable;
+import java.util.Arrays;
+
 import javaemul.internal.NativeRegExp;
 import javaemul.internal.annotations.HasNoSideEffects;
 
 /**
  * Wraps a native <code>char</code> as an object.
  *
- * Some methods are not possible to implement without a Unicode database,
- * which would blow up the code size.
+ * <p>Some methods are not possible to implement without a Unicode database,
+ * which would blow up the code size.</p>
+ *
+ * <p>Methods such as isLetter, isDigit, ... use the JS native API for Unicode.
+ * Their output is only consistent with JVMs that have the same Unicode support
+ * as the target browser. As of 2025, most browsers provide Unicode 16.0 support
+ * which is on par with OpenJDK 24.</p>
+ *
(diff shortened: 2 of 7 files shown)


user/super/com/google/gwt/emul/java/lang/Character.java
```diff
@@ -194,24 +177,32 @@ public static int compare(char x, char y) {
...
-
-    // The offset by 10 is to re-base the alpha values
-    if (c >= 'a' && c < (radix + 'a' - 10)) {
-      return c - 'a' + 10;
-    }
-
-    if (c >= 'A' && c < (radix + 'A' - 10)) {
-      return c - 'A' + 10;
+    int digit;
+    if (Character.isDigit(codepoint)) {
```
niloc132, line 188:
```suggestion
    if (isDigit(codepoint)) {
```

user/super/com/google/gwt/emul/java/lang/Character.java
```diff
@@ -234,57 +225,145 @@ public static boolean isBmpCodePoint(int codePoint) {
...
 
-  /*
-   * TODO: correct Unicode handling.
-   */
   public static boolean isDigit(char c) {
+    return isDigit(String.valueOf(c));
+  }
+
+  // Known differences between Java 17 and Chrome 135
+  // 11f50 .. 11f59, 16ac0 .. 16ac9, 1e4f0 .. 1e4f9, 1fbf0 .. 1fbf9
```
niloc132, line 253: Should we leave this method out if it is still wrong? Which side do we consider to be "wrong" here, Java or Chrome? Would it make sense to have a test (possibly ignored) that shows this, so it is easier to reevaluate later?

Also, consider using hex in digit() above, so that it is the same convention as here, or change this to be decimal?

user/test/com/google/gwt/emultest/java/lang/CharacterTest.java
```diff
@@ -15,25 +15,33 @@
...
 
 import java.util.Arrays;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
 
 /**
  * Tests for java.lang.Character.
  */
+@DoNotRunWith(Platform.HtmlUnitBug)
```
niloc132, line 30: Can the specific failing tests be annotated with this, each with a comment about why it fails, rather than skip existing tests that pass?

user/super/com/google/gwt/emul/java/lang/CaseMapper.java
```diff
@@ -23,7 +23,23 @@ public static char charToLowerCase(char c) {
...
+    return String.NativeString.fromCodePoint(codePoint).toLowerCase().codePointAt(0);
+  }
+
+  public static int intToUpperCase(int codePoint) {
+    String upper = String.NativeString.fromCodePoint(codePoint).toUpperCase();
+    return hasExtraCodePoints(upper) ? codePoint : upper.codePointAt(0);
+  }
+
+  // If String.toUpperCase produces more than 1 codepoint, Character.toUpperCase should
+  // act either as identity or title-case conversion (not supported in GWT).
```
niloc132, line 40: Maybe add a failing test and mark it ignored so we can see about restoring it when it works?

user/test/com/google/gwt/emultest/java/lang/CharacterTest.java
zbynek: I's pretty much all the tests that call any Character.is* method. They fail because the regexp doesn't even parse in HtmlUnit 2-4. If we want any meaningful test coverage we need https://github.com/gwtproject/gwt/pull/10115 + a custom build of HtmlUnit that includes https://github.com/mozilla/rhino/pull/1848

niloc132:
> Marking as draft until HtmlUnit can be updated to 4.x.

How about just a @DoNotRunWith that we can reevaluate later, and confirm that it runs in manual mode?

zbynek:
> How about just a @DoNotRunWith that we can reevaluate later, and confirm that it runs in manual mode?

Works for me, but it also means that any `GwtTestcase` out there using `HtmlUnit` as a runner that happens to call `Character.isDigit` will need to be ignored or switched to Selenium. Are we OK with that?

niloc132:
I get it, thanks, that is unreasonable to force at this time. Let's sit on this then as you suggested. 

It will be a little tricky to do right - forcing new htmlunit also means dropping legacy dev mode tests or fixing the impls you have in your patch.

If we can't fix the new impl to drop legacy dev mode, we have to:
 * drop legacy dev mode once and for all (maybe its time?)
 * support old htmlunit (keep both old and new)
 * avoid updating htmlunit for a while longer (and thus avoid this patch until we stop avoiding it)

I really want to keep the "mavenize the build" release as light on features as possible so that projects can reliably upgrade just by changing dependencies and not expect to encounter any other changes along the way - my initial hope had been for 2.13 to ship soon with all of the code changes, followed a month or so later by 2.14. Then, 2.15 might do various other things like upgrade all kinds of dependencies, continue dropping old deprecated features, etc - teams that successfully move to 2.14 then might have issues with that next step, but at least won't be confusing those changes with the modularization efforts.

To that end, breaking "gwt-runstyle-htmlunit2" out of the main "this artifact contains the compiler" jar will make it at least possible for projects to supply an alternate implementation (of their own or another ecosystem/official build). Then, the existence of legacy dev mode is its own problem to solve.

PR #10115, opened 2025-04-13
Update HtmlUnit to 4.11.1, update necessary dependencies

zbynek:
Fixes #9958 

Related changes need to be merged in `tools` repo first, the workflow definitions here need to be reverted.

(I was hoping this would help with regex parsing for, #10113 but it's not sufficient.)


dev/core/src/com/google/gwt/dev/shell/HostedModePluginObject.java
```diff
@@ -193,8 +191,8 @@ public boolean connect(String url, String sessionKey, String address,
     // Object[] with a success boolean and a value, and HtmlUnit will guard against this.
     // The simplest way to do that here is to mark java.lang.Object as the java equivalent
     // of some JS type - the name of the type doesn't matter.
-    webClient.setActiveXObjectMap(Collections.singletonMap(
-            "GwtLegacyDevModeExceptionOrReturnValue", "java.lang.Object"));
+    // webClient.setActiveXObjectMap(Collections.singletonMap(
+    //        "GwtLegacyDevModeExceptionOrReturnValue", "java.lang.Object"));
```
niloc132, line 195: If we're removing this, we should probably remove it outright, and the comment too.

That said, I think legacy dev mode tests work still with HtmlUnit, so this is probably required to keep? Unless we are actually outright removing that feature, in which case there is a lot more to remove too?

niloc132 (commented):
Thanks for working on this - I had tentatively planned on splitting this off until 2.14 (mavenize gwt itself first) and/or split into an external project (via https://github.com/gwtproject/gwt/issues/10058, so that htmlunit could be entirely external, not bring so many other dependencies to the gwt-dev classpath...

What do you think about us fixing 10058 and then offering a separate gwt-runstyle-htmlunit jar that provides -runStyle=htmlunit4, which then drops legacy dev mode support etc, changes the system tz, etc? I'd also love to expose the webClient.options somehow so that downstream projects could enable future polyfills as desired...

niloc132:
Agreed, I don't want to continue to maintain the old one, just to make sure it is there for legacy dev mode until we're committed to removing it. Post-mavenization (or Gradle, if there were suddenly a push to have good Gradle support), I would want to have `gwt-dev.jar` to continue to exist in roughly its current form for legacy ("still productive and profitable, but without significant development resources") applications to continue to be maintained. Those either prepared to _add_ `gwt-runstyle-htmlunit4.jar`, or switch outright to the specific jars they actually want (pretend these are `gwt-compiler.jar`, `gwt-test.jar`, `gwt-jre.jar`, etc... plus their desired `gwt-runstyle-*.jar`s) then could have an easy way to track the latest htmlunit versions as they are available. Or, switch to the new jars, and leave out htmlunit entirely if desired, and get none of its classpath (which as of today, still includes jetty 9).

I'm not quite sure what you're asking about rhino - if there was an easy update that could be done for htmlunit to get es5 class, es6 classes etc, they would have already done it. Or are you asking if GWT's custom rhino could be replaced with htmlunit's implementation? If it is the latter, I think the answer is no, as JSNI doesn't actually support all of es3 - the `with` keyword has been altered to fail:
https://github.com/gwtproject/gwt/blob/15609b6e008ab0b945617d9d4d32e6859d56b894/dev/core/src/com/google/gwt/dev/js/rhino/Parser.java#L691-L696

JSNI also supports _more_ than es3, the JSNI method reference syntax:
https://github.com/gwtproject/gwt/blob/15609b6e008ab0b945617d9d4d32e6859d56b894/dev/core/src/com/google/gwt/dev/js/rhino/TokenStream.java#L1355-L1576

zbynek:
> Or are you asking if GWT's custom rhino could be replaced with htmlunit's implementation? If it is the latter, I think the answer is no

That's what I meant, thanks. The reason I mentioned ES6 is that Rhino 1.8 (and its HtmlUnit fork) enable [many ES6 features](https://mozilla.github.io/rhino/compat/engines.html) by default.

niloc132:
Oh that's a nice view - possibly worth a look when it gets closer (and there's a lot less red even in the last column), but I think we should be discouraging writing new JSNI, and instead encouraging jsinterop where possible? 

Note that GWT doesn't use rhino to emit JS, so we won't get any extra generated JS features out of this, only parsing JSNI expressions.

zbynek:
I was wondering if this update can be done without removing dev mode tests. I managed to get them to run (see last commit), but there are many strange failures under Java 17:

```
Caused by: java.lang.ExceptionInInitializerError: Exception java.lang.IllegalAccessError:
 class com.google.gwt.dom.client.Style$Unit$1 tried to access private method
  'void com.google.gwt.dom.client.Style$Unit.<init>(java.lang.String, int)'
   (com.google.gwt.dom.client.Style$Unit$1 and com.google.gwt.dom.client.Style$Unit are in unnamed module
   of loader com.google.gwt.dev.shell.CompilingClassLoader @382b53c2), (Type com.google.gwt.dom.client.Style$Unit$1
   (loader: com.google.gwt.dev.shell.CompilingClassLoader @382b53c2) is not a nest member of
   type com.google.gwt.dom.client.Style (loader: com.google.gwt.dev.shell.CompilingClassLoader @382b53c2):
   current type is not listed as a nest member, Type com.google.gwt.dom.client.Style$Unit
   (loader: com.google.gwt.dev.shell.CompilingClassLoader @382b53c2) is not a nest member of
    type com.google.gwt.dom.client.Style (loader: com.google.gwt.dev.shell.CompilingClassLoader @382b53c2):
    current type is not listed as a nest member) [in thread "Code server for com.google.gwt.dom.DOMTest.JUnit
```

zbynek:
@niloc132 I checked with `main` and I'm getting very similar errors with Java 17, and I expect the same happens with Java 11. Thoughts on merging this as is (possibly waiting for https://github.com/mozilla/rhino/pull/1848 ?)

zbynek:
This is also needed for running on Java 26 
#10336

```
Caused by: java.lang.NoClassDefFoundError: java/applet/AppletStub
	at com.gargoylesoftware.htmlunit.html.DefaultElementFactory.createElementNS(DefaultElementFactory.java:551)
	at com.gargoylesoftware.htmlunit.html.DefaultElementFactory.createElementNS(DefaultElementFactory.java:51)
```

rbri:
@niloc132 HtmlUnit is now at 5.1.0 (JDK 17). anything i can do to help with moving on to a recent version? Any issue that blocks you from updating?

Issue #10125, opened 2025-05-01
hiddenDiv not removed when using UiBinder


ChristopherBruno78:
If you create a component using uiBinder, the a hidden div remains in the DOM.

Example:

Button.ui.xml
```
<ui:UiBinder xmlns:ui='urn:ui:com.google.gwt.uibinder'>
    <button class="m-Button default">
        <div>
            <i class="icon" ui:field="iconLayer" />
            <label ui:field="labelLayer" />
        </div>
    </button>
</ui:UiBinder>
```

Button.java

```
package com.example.ui; 

public class Button
  extends FocusWidget {

   interface ButtonUiBinder extends UiBinder<ButtonElement, Button> {}

  private static final ButtonUiBinder uiBinder = GWT.create(
    ButtonUiBinder.class
  );

@UiField 
LabelElement labelLayer;

@UiField
Element iconLayer;

  public Button() {
     setElement(uiBinder.createAndBindUi(this));
  }

}
```

And then create a Composite to use the button:

```
<ui:UiBinder xmlns:ui='urn:ui:com.google.gwt.uibinder'
             xmlns:g='urn:import:com.google.gwt.user.client.ui'
             xmlns:a='urn:import:com.example.ui' 
>
 
    <g:FlowPanel >
        <a:Button />
    </g:FlowPanel>

</ui:UiBinder>
```

```
public class MyComposite extends Composite {

 interface MyCompositeUiBinder extends UiBinder<FlowPanel, MyComposite> {}

  private static final MyCompositeUiBinder uiBinder = GWT.create(
    MyCompositeUiBinder.class
  );

  public MyComposite() {
       initWidget(uiBinder.createAndBindUi(this));
  }

}

```
When MyComposite is added to the RootPanel, you see a phantom hidden div in the DOM that does nothing. Be nice if that was not there.

ChristopherBruno78:
@zbynek Thanks I tried your suggestion but the hidden div is still there. Attached is a screenshot of what I am referring to. 

<img width="1086" alt="Image" src="https://github.com/user-attachments/assets/58d3471b-4df6-4a21-a960-a50e7f105955" />

ChristopherBruno78:
Was able to fix it and created a PR. There maybe a better way to accomplish it, however, please review.

jhickman:
[...]

PR #10126, opened 2025-05-02
Removes hidden div, issue #10125

ChristopherBruno78:
Removes the hidden created in UIBinder that is not removed in certain cases. #10125

### DIFF
--- a/user/src/com/google/gwt/user/client/ui/HTMLPanel.java
+++ b/user/src/com/google/gwt/user/client/ui/HTMLPanel.java
@@ -355,5 +355,4 @@ public class HTMLPanel extends ComplexPanel {
     hiddenDiv.removeFromParent();
     return child;
   }
-
 }
--- a/user/src/com/google/gwt/user/client/ui/HTMLPanel.java
+++ b/user/src/com/google/gwt/user/client/ui/HTMLPanel.java
@@ -332,8 +332,8 @@ public class HTMLPanel extends ComplexPanel {
     if (hiddenDiv == null) {
       hiddenDiv = Document.get().createDivElement();
       UIObject.setVisible(hiddenDiv, false);
-      RootPanel.getBodyElement().appendChild(hiddenDiv);
     }
+    RootPanel.getBodyElement().appendChild(hiddenDiv);
 
     // Hang on to the panel's original parent and sibling elements so that it
     // can be replaced.
@@ -352,7 +352,8 @@ public class HTMLPanel extends ComplexPanel {
     } else {
       hiddenDiv.removeChild(getElement());
     }
-
+    hiddenDiv.removeFromParent();
     return child;
   }
+
 }
--- a/user/src/com/google/gwt/uibinder/client/UiBinderUtil.java
+++ b/user/src/com/google/gwt/uibinder/client/UiBinderUtil.java
@@ -55,7 +55,7 @@ public class UiBinderUtil {
       } else {
         orphan(element);
       }
-      removeHiddenDiv();
+      orphan(hiddenDiv);
     }
   }
 
@@ -88,7 +88,7 @@ public class UiBinderUtil {
     hiddenDiv.setInnerHTML(html);
     Element newbie = hiddenDiv.getFirstChildElement();
     orphan(newbie);
-    removeHiddenDiv();
+    orphan(hiddenDiv);
     return newbie;
   }
 
@@ -97,15 +97,8 @@ public class UiBinderUtil {
     if (hiddenDiv == null) {
       hiddenDiv = Document.get().createDivElement();
       UIObject.setVisible(hiddenDiv, false);
-      RootPanel.getBodyElement().appendChild(hiddenDiv);
     }
-  }
-
-  private static void removeHiddenDiv() {
-    if (hiddenDiv != null) {
-      hiddenDiv.removeFromParent();
-      hiddenDiv = null;
-    }
+    RootPanel.getBodyElement().appendChild(hiddenDiv);
   }
 
   private static void orphan(Node node) {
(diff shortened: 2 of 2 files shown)


jhickman (changes requested):
Thanks for the PR.  I did look at this and got my head back into it after a few years.

In the UiBinderUtil, it should be noted that HTMLPanel does this same thing as this code was copied from there.   Anywhere using the HTMLPanel will have the same kind of div appear; and it's also a static field in HTMLPanel.   So, if both HTMLPanel AND UiBinderUtil is used, then 2 hidden DIVs will be left in the DOM.

The div is used for temporary placement while native functions are being called/used (such as getElementById() or setInnerHtml).

Rather than setting the hiddenDiv to null and causing it to be re-created (which will be done A LOT since this exists in AbstractFieldWriter, maybe just orphan it and reattach when needed?

ChristopherBruno78:
@jhickman Thanks for the review. I made your recommend changes.

user/src/com/google/gwt/user/client/ui/HTMLPanel.java
```diff
@@ -332,8 +332,8 @@ private Element attachToDomAndGetElement(String id) {
     if (hiddenDiv == null) {
       hiddenDiv = Document.get().createDivElement();
       UIObject.setVisible(hiddenDiv, false);
-      RootPanel.getBodyElement().appendChild(hiddenDiv);
     }
+    RootPanel.getBodyElement().appendChild(hiddenDiv);
```
jhickman, line 336: tricky situation here.  If the div is already in the DOM, this will move it to the end of the `<body>`, making it the last child.  This **does detach and reattach** it under the hood.  Although it doesn't recreate the element, this move is unnecessary.

jnehlmeier:
I generally don't like that change. 

The `hiddenDiv` is a static element and used for all `HTMLPanel` instances in the UI (same for `UiBinderUtil` with its own `hiddenDiv`) and this change will add/remove the `div` element as often as `HTMLPanel` instances are used to build the UI plus `UiBinderUtil` calls that occur if the UI is built using UiBinder. So it can result in quite a few more (useless) DOM operations for more complex UIs.

Creating the `div` lazily and keeping it attached is an optimization to reduce the amount of DOM operations while building the UI. I don't see the point of removing this optimization. If it is just for aesthetics GWT could mark the element as GWT internal and keep the logic as is.

user/src/com/google/gwt/user/client/ui/HTMLPanel.java
ChristopherBruno78: In this case the hidden div is used to insert the element to use the native getElementById. It seems like it doesn't matter where it is in the <body>, and although the hidden div may be used again, in this instance after the element is retrieved, the need for the hiddenDiv is done and should be removed. I think the hidden div doesn't need to be a static variable, it can recreated on the fly, used and removed. In terms of performance, this is very minimal and it's not worth keeping the div around just for performance.

niloc132:
Yeah, this is going to make every app out there _very slightly_ bigger and, depending on how much of this kind of dom building they do, slightly slower.

Since there can only be a max of one, and it will get created/removed a lot, a singleton definitely made sense historically here - are you sure that dom elements are effectively free to re-create, rather than just making one and shuffling it around?

Two other approaches I'd suggest if we want to do this:
 * let it continue to be automatic, but guard it with a config property that defaults to false, so no one else is penalized for this (there is never more than one extra dom element in the entire app, and it is invisible, and no one else has been bothered for over a decade)
 * Make a new public static method on UiBinderUtil to grant any app that cares control to remove the last instance of the invisible div.
   * Alternatively, use deferred binding here (and remake the static methods to instance ones) so a specific application can rewrite how this works via a `replace-with` element in their app. You could even put a scheduleFinally call or the like to set up cleanup at the end of the whole event loop exactly once, rather than doubling the number of times the document is manipulated for this div.

But in short, I don't know why every app should have to pay for one user not liking a single, invisible, dom element.

ChristopherBruno78:
I think the performance concerns mattered more in the age of IE, but I think at least inserting a DOM comment would be helpful. I guess the original idea GWT was to abstract the DOM and focus on widgets, but nowadays GWT is a great alternative to building complex components and UIs, but the developer still should understand the DOM structure that is being built.  A comment for anything inserted not directly apart of a component would demystify it and be helpful.

niloc132:
I don't necessarily disagree, but there are myriad such notes that could be added in this way. Every attribute, event listener... once you go down this road, why even support minifying and optimizing code?

By and large, the generated code (and dom structures) aren't intended to be read by humans, but by machines.

user/src/com/google/gwt/user/client/ui/HTMLPanel.java
ChristopherBruno78: I noticed I do not see the hiddenDiv with HTMLPanel for some reason, although looking at the code it seems like it should be there.

ChristopherBruno78:
> By and large, the generated code (and dom structures) aren't intended to be read by humans, but by machines.

If you are building your own components, what is in the DOM is going read by the developer. It's that this div seems to come from nowhere and is not documented that makes it problematic. Behavior associated with a component is assumed to be documented in the source.

niloc132:
I would not be opposed to adding comments to the source - either UiBinderUtil itself, or the generated Java that comes out of the UiBinderWriter could have additional comments, help someone directly reading the source, and have no impact on the generated output (even in PRETTY or DETAILED, there are never comments in gwt-generated JS, and I don't think any widget is capable of emitting comments in the dom).

I still don't see anything "problematic" in any way about this div. "Surprising" perhaps, but I think that judgement can be made about many things that GWT emits as generated output.

PR #10132, opened 2025-05-13
Do not emulate annotations

### LINKED ISSUE #10131: Support attributes introduced in Java 9 for @Deprecated annotation
Java 9 introduced additional attributes `String since` and `boolean forRemoval` to the `Deprecated` annotation.

However, using these attributes in our code leads to a GWT compilation error:

    @Deprecated( since = "2017-11-13" ) // -> [ERROR] Line 383: The attribute since is undefined for the annotation type Deprecated
    public void foo() {
      [...]
    }

Setting `-sourceLevel 17` does not prevent this.

It would be nice if this could be supported at some point. At the very least when support for Java 8 is dropped at some point.

zbynek:
Fixes #10131 

As pointed out the annotation doesn't have to be emulated at all since #10020 was fixed.


build.xml
```diff
@@ -31,7 +31,7 @@
   </macrodef>
 
   <property name="gwt.apicheck.config"
-            location="tools/api-checker/config/gwt211_212userapi.conf"/>
+            location="tools/api-checker/config/gwt211_213userapi.conf"/>
```
niloc132, line 34: I'll make a new api jar instead for 2.12 - or you can? the instructions are in https://github.com/gwtproject/tools/tree/main/api-checker-reference

zbynek:
Passing build https://github.com/zbynek/gwt/actions/runs/15096128855

niloc132 (commented):
This looks right, but why stop here - we've got a few other emulated annotation types that can be removed too:
javax.annotation.Generated
javax.annotation.processing.Generated

org.junit.runner.RunWith
org.junit.runners.SuiteClasses

Do not remove javax.validation.Pattern, as it has inner types like Flag that can't be emulated any other way (though also might not technically be needed for emulation, but might possible be used downstream).

user/super/com/google/gwt/emul/java/lang/annotation/AnnotationTypeMismatchException.java
```diff
@@ -1,25 +0,0 @@
-/*
```
zbynek, line 1: This one is not an annotation, but 
1) there is no use-case for throwing this from browser code
2) [both methods declared by this class](https://docs.oracle.com/en/java/javase/17/docs/api//java.base/java/lang/annotation/AnnotationTypeMismatchException.html) were missing from emulation

Probably better to remove than add the missing methods.

niloc132: That does seem like a weird one to be emulated - looks like it could have been a part of a push to emulate anything that can be emulated way back: https://github.com/gwtproject/gwt/commit/21eef3b0834e1963ba6d519ffb0c9a889c2dd5b9

That would make it at least possible to handle such exceptions in gwt-rpc clients.

However in this case, since the Method can't be made into a field, without a custom field serializer, this can't work.

Normally I'd be inclined to say we shouldn't take away existing emulation that someone _might_ be using... but I think this one is pretty sketchy to begin with.

niloc132 (approved):

PR #10139, opened 2025-07-03
Fix console log uses debug

larzeni:
This patch updates ConsoleLogHandler to use console.debug() instead of console.log() for log levels FINE, FINER, and FINEST, ensuring that these messages are only shown in the "Verbose" tab in browser developer tools. This aligns with standard browser behavior for log levels.

### DIFF
--- a/user/src/com/google/gwt/logging/client/ConsoleLogHandler.java
+++ b/user/src/com/google/gwt/logging/client/ConsoleLogHandler.java
@@ -57,7 +57,7 @@ public class ConsoleLogHandler extends Handler {
     } else if (val >= Level.INFO.intValue()) {
       info(msg);
     } else {
-      log(msg);
+      logVerbose(msg);
     }
   }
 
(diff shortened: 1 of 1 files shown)


larzeni:
I beg your pardon, this is my first attempt to publish a patch

user/src/com/google/gwt/logging/client/ConsoleLogHandler.java
```diff
@@ -77,7 +77,9 @@ private native void info(String message) /*-{
     window.console.info(message);
   }-*/;
 
-  private native void log(String message) /*-{
-    window.console.log(message);
+  private native void logVerbose(String message) /*-{
```
jnehlmeier: Please rename the method to `debug` to match the naming scheme of all the other logging methods. Also the comment can be removed as it is already pretty clear what is going on.

jnehlmeier (changes requested):

jnehlmeier (approved):

PR #10141, opened 2025-07-16
Update nightly snapshot deployment URL

niloc132:
Release staging URL is set, however we probably want to use the publish API with a zip bundle instead, at least until the project uses Maven/Gradle to build and publish.

Secrets in GitHub Actions have already been changed to work with the new deployment.

### DIFF
--- a/maven/push-gwtproject.sh
+++ b/maven/push-gwtproject.sh
@@ -6,10 +6,10 @@
 # GWT_MAVEN_REPO_ID = a server id in your .m2/settings.xml with remote repo username and password
 #
 # Sonatype staging repo (promotes to Maven Central)
-#   GWT_MAVEN_REPO_URL=https://oss.sonatype.org/service/local/staging/deploy/maven2/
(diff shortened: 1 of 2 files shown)


maven/push-gwtproject.sh
```diff
@@ -6,10 +6,10 @@
 # GWT_MAVEN_REPO_ID = a server id in your .m2/settings.xml with remote repo username and password
 #
 # Sonatype staging repo (promotes to Maven Central)
-#   GWT_MAVEN_REPO_URL=https://oss.sonatype.org/service/local/staging/deploy/maven2/
+#   GWT_MAVEN_REPO_URL=https://ossrh-staging-api.central.sonatype.com/service/local/staging/deploy/maven2/
 #
 # Sonatype Google SNAPSHOTs repo (can only deploy SNAPSHOTs here, and they are immediately public)
```
tbroyer: nit: maybe remove "Google" from the comment here

tbroyer (dismissed):

tbroyer (approved):

PR #10142, opened 2025-07-16
Restore PR annotations

niloc132:
Our reviewdog annotations stopped appearing some time ago - they've been fixed once, but broken again. This attempts to repair them again, so PRs with minor style issues don't fail, but do have warnings visible for reviewers to easily comment.

### DIFF
--- a/.github/workflows/quick-check.yml
+++ b/.github/workflows/quick-check.yml
@@ -7,6 +7,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
       matrix:
         java-version: ['11', '17', '21', '22']
     steps:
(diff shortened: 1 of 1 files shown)


niloc132:
Example of annotations for a deliberately broken class... with an accidental error picked up too. The import was not deliberately broken.

https://github.com/niloc132/gwt/pull/4/files#diff-f89e15f3f18e7f71fcffcdcac9c6c06dc83e17cfedc6e30ef7e14b42f1eb7b69R25

image
image

zbynek:
I was checking something like this earlier and noticed that we're running CheckStyle in all the branches (so either one may fail), because of the matrix strategy that fails the build in all branches and thus the branch responsible for running reviewdog does not have access to the reports. Maybe we should also try to skip checkstyle in all branches that don't publish the result?

Attemt to recreate the issue: https://github.com/niloc132/gwt/pull/5

niloc132:
You're right @zbynek  - i'll make the change to ensure that for these small runs we don't let one failure cancel the others.

niloc132:
@zbynek I've disabled fail-fast for the quick checks, so that all JVM versions run to completion.

zbynek (approved):

PR #10143, opened 2025-07-16
Deprecate Util for removal

### LINKED ISSUE #10030: Deprecate `com.google.gwt.dev.util.Util` for removal
This has a TODO indicating that at least stream functions should be replaced with guava. Other IO functions should also move to Guava or java.nio, and xml functions have no need to exist at all. This reduces the class to just a few calls, and can be moved to somewhere more appropriate in the codebase, and the class deleted.

There are other TODOs in calls to methods in this class that can be cleaned up by this work as well.

Replacing usage of this class will help with a few classes of issues
* Masking IO errors by throwing NPE instead
* Buffering output that need not be buffered (such as reading into a bytearray, or writing a bytearray/string to a file in one step).
* Buffering output that might benefit from being buffered (writing cache files to disk)

Fully completing this work suggests raising the java release level to 11.

See also https://github.com/gwtproject/gwt/issues/10011.

niloc132:
This class is intended to be "less public" than Utility (deprecated in #10083), but is just as accessible to developers as Util is. Like Utility, many methods could be replaced by either newer JRE method or Guava. Some implementations were too general, or masked errors by returning null, unnecessarily buffered in-memory reads/writes, these have been updated as part of this patch.

Hashing changes were merged in #10102.

The patch slightly increases the size of the project, but will be by far a net decrease after Util.java itself is removed.

The class itself has been deprecated for removal, allowing downstream Generators/Linkers/etc time to update.  

Fixes #10030


dev/core/src/com/google/gwt/dev/util/HtmlTextOutput.java
```diff
@@ -16,11 +16,39 @@
...
+  private static final Pattern ESCAPE_PATTERN = Pattern.compile("[&<>'\"]");
+
+  /**
+   * Escapes any characters that need to be escaped for HTML output - &, <>, ', and ".
+   */
+  private static String escapeHtml(String s) {
+    if (s == null) {
+      return null;
+    }
+    return ESCAPE_PATTERN.matcher(s).replaceAll(match -> {
```
vegegoku, line 34: This is duplicate with the escape function in the ModuleSpace class

vegegoku (approved):

dev/core/src/com/google/gwt/core/ext/linker/impl/SelectionScriptLinker.java
```diff
@@ -15,6 +15,8 @@
  */
 package com.google.gwt.core.ext.linker.impl;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
```
zbynek, line 18: The existing codebase imports the whole class instead of statically importing this one field, not sure if that matters.

zbynek (approved):

dev/core/src/com/google/gwt/dev/util/HtmlTextOutput.java
niloc132: Deliberate duplicate, so that ModuleSpace can be part of "this is only for legacy dev mode", and HtmlTextOutput is in turn only for "soyc html report writing". Ideally both will go away, or switch to some standardized library (gwt's own safehtml for instance...), but for now I was okay with just pasting it twice.

Even for all that code added, if we count that Util will be delete soon, this set of changes is a net negative for lines of code in the project.

dev/core/src/com/google/gwt/core/ext/linker/impl/SelectionScriptLinker.java
niloc132: We don't disallow static imports - and I didn't like how many extra line breaks were required just to do this cleanup. That said, keeping it readable and consistent is important, if you'd like me to make a chance (either direction) here, I'm happy to do it.

PR #10146, opened 2025-07-23
TreeSet/TreeMap should implement Cloneable and offer clone()

### LINKED ISSUE #10145: TreeSet and TreeMap don't support clone()
Most other Java collection classes correctly implement the Cloneable marker interface and have a clone() method, but at least these two classes are missing this.

See also https://github.com/gwtproject/gwt/issues/5067.

niloc132:
Fixes #10145

### DIFF
--- a/user/super/com/google/gwt/emul/java/util/TreeMap.java
+++ b/user/super/com/google/gwt/emul/java/util/TreeMap.java
@@ -28,7 +28,7 @@ import java.io.Serializable;
  * @param <K> key type
  * @param <V> value type
  */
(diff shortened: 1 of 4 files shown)


zbynek (approved):
Looks good. There is no speed optimization (it should be possible to clone the map without sorting everything), but that's already noted in https://github.com/gwtproject/gwt/blob/0396c65993315febd7e5369a0189dc8dbf5c73e4/user/super/com/google/gwt/emul/java/util/TreeMap.java#L496

niloc132:
RIght, this doesn't fix that issue at all. I have a draft that tried to fix how we load the data, but it isn't good enough to be worth even putting up a draft yet.

PR #10150, opened 2025-08-01
Add back javax.annotation.processing.Generated

zbynek:
Fixes https://github.com/gwtproject/gwt/issues/10149 by partially reverting https://github.com/gwtproject/gwt/pull/10132


niloc132:
Suggesting that we close this, and instead go with fixing the root issue via https://github.com/gwtproject/gwt/pull/10153

PR #10154, opened 2025-08-23
Don't run full build on schedule, run it for "ready" label

zbynek:
This implements @niloc132 's idea to run full build when a specific label is present. Also removes the scheduled build as it's enough to build each commit to `main` once (over the last year scheduled builds did not detect any flaky issues with GWT itself, they only fail sometimes because of some strange Ant issue).

Demonstration PR: https://github.com/zbynek/gwt/pull/4

### DIFF
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -4,10 +4,6 @@
 # build, though maven snapshots are not yet deployed.
 name: Full build
 on:
-  schedule:
-    # This time is selected to be early in the morning in NA/SA/EU/ME. This
-    # only runs on the default branch.
-    - cron:  '0 6 * * *'
   push:
     # Build on pushes to main and release branches. For forks, the -test may
     # be helpful to run tests when preparing a merge.
@@ -17,8 +13,11 @@ on:
       - '*-test'
   workflow_dispatch:
     # Allow running manually
+  pull_request:
+    types: [ labeled, synchronized ]  
 jobs:
   build:
+    if: ${{ !github.event.pull_request || contains(github.event.pull_request.labels.*.name, 'ready') }}
     runs-on: ubuntu-latest
     strategy:
       matrix:
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -14,7 +14,7 @@ on:
   workflow_dispatch:
     # Allow running manually
   pull_request:
-    types: [ labeled, synchronized ]  
+    types: [ labeled, synchronize ]
 jobs:
   build:
     if: ${{ !github.event.pull_request || contains(github.event.pull_request.labels.*.name, 'ready') }}
@@ -84,7 +84,7 @@ jobs:
       - name: Set up sonatype credentials
         # Using the same java version as above, set up a settings.xml file
         uses: actions/setup-java@v4
-        if: ${{ github.event_name == 'schedule' && github.repository_owner == 'gwtproject' && matrix.java-version == '21' }}
+        if: ${{ github.event_name == 'push' && github.ref_name == 'main' && github.repository_owner == 'gwtproject' && matrix.java-version == '21' }}
         with:
           java-version: ${{ matrix.java-version }}
           distribution: 'temurin'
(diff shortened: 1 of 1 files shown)


.github/workflows/full-check.yml
```diff
@@ -17,8 +13,11 @@ on:
       - '*-test'
   workflow_dispatch:
     # Allow running manually
+  pull_request:
+    types: [ labeled, synchronized ]
```
niloc132: If I understand this correctly, it roughly says "If a PR has a label change or has code pushed to it, run the full check if the `ready` label is present".

The ability to add a label is limited to a certain group of contributors - I'll assume for now that all of those users are trusted to be able to do anything they want (such as publish artifacts with the keys that build has access to). This does at least limit an untrusted attacker from being able to create a malicious PR.

But it does not prevent an attacker from creating a seemingly-helpful PR, then wait for the `ready` label to be added, and then push a malicious commit - the build will helpfully run right away.

I'm not sure how to defend against that, aside from something like "if the user isn't trusted, revoke the label as soon as an early build step" coupled with a "this build can never run concurrently on two commits of the same PR".

.github/workflows/full-check.yml
```diff
@@ -4,10 +4,6 @@
 # build, though maven snapshots are not yet deployed.
 name: Full build
 on:
-  schedule:
```
niloc132, line 7: Note that with this removed, we will no longer publish to sonatype snapshots - the steps below use this check:
```
if: ${{ github.event_name == 'schedule' && github.repository_owner == 'gwtproject' && matrix.java-version == '21' }}
```
This check is far from perfect - but as `schedule` only runs on the default branch, we don't have to worry about pushes to `release/*` overwriting latest HEAD-SNAPSHOT. To finish this patch and let a `push` event replace the `schedule` event for publishing, we'll have to make sure the GWT_VERSION env var makes sense too.

.github/workflows/full-check.yml
zbynek: >  "If a PR has a label change or has code pushed to it, run the full check if the ready label is present".

Exactly.

> But it does not prevent an attacker from creating a seemingly-helpful PR, then wait for the ready label to be added, and then push a malicious commit - the build will helpfully run right away.

Is this in any way different from the quick PR build that runs on each commit of each PR (after initial approval)?

.github/workflows/full-check.yml
zbynek: New version pushed that should run this only when pushed to `main`. As I understand there are currently no Sonatype snapshots for `release/*` branches so maybe we can keep it like that for now?

.github/workflows/full-check.yml
niloc132: I don't think it is just initial approval, but after each change I (or someone else) must click the approve-workflow-run button?

image

It looks like we have this configured today to only apply until the user gets their first PR landed, then later PRs are assumed-good. My gut reaction is that this isn't safe enough, but maybe I need to be a little better educated here...

zbynek: So the risks here are
* someone uses GWT's credentials to upload something harmful to Sonatype -- likely not an issue since
> Anyone with collaborator access to this repository can use these secrets and variables for actions. They are not passed to workflows that are triggered by a pull request from a fork.
-- https://github.com/gwtproject/gwt/settings/variables/actions

* someone spams the CI and makes GWT run out of CI quota -- I seriously doubt anyone would spend time getting a useful PR merged just to do that. I don't remember that happening in orgs that have more PR traffic than gwt.
* am I missing something else?

niloc132 (commented):
Sorry for my delay on this - given no other negative feedback on this, I think we should do it.

A few questions:
 * Do we need a step that removes the label after another push? That way, an innocuous change wouldn't get pushed, get labeled, then while building get a new push that contains something malicious? It could nearly also work by making the friste step in the build remove the label, but it does leave a race possible... it looks like this is what the protobuf project does though https://github.com/protocolbuffers/protobuf/blob/main/.github/workflows/test_runner.yml. Maybe we could just leave it with `labeled` and not `synchronized` and that would be a non-issue (plus a "auto-remove label" step)?
 * Let's add a `ready` label to the project, with a description of what it means ("this PR has been reviewed by a maintainer and is ready for a CI run" etc?)

zbynek:
>  an innocuous change wouldn't get pushed, get labeled, then while building get a new push that contains something malicious?

I still don't understand your concern about malicious code.
1) the same concerns apply to `quick-check`, you can modify code that executes there too (by adding an ant step that runs custom code...)
2) I don't know what could that code do that's GWT-specific (other than depleting the CI time quota). The author can just run the code in their own repo's CI.

> Let's add a ready label to the project, 

Done.

niloc132:
I think you're right, I'm just paranoid about this.

niloc132 (approved):

PR #10165, opened 2025-10-22
Allow JsUtils.uncheckedCast to be inlined away

### LINKED ISSUE #10055: $clinit_Boolean is not always optimized out when empty
**GWT version:** 2.12.1, latest
**Operating System:** any

---

##### Description

While most empty clinits seem to be correctly optimized out, sometimes they unnecessarily remain in the compiled output. 

As an example, if you build https://github.com/gwtproject/gwt-site-webapp/ but change turn draft mode off, `$clinit_Object` is correctly removed, but `$clinit_Boolean` is not, despite being empty:
```javascript
function $clinit_Boolean(){
  $clinit_Boolean = emptyMethod;
}
```

A Java lambda of type `elemental2.core.ReadonlyArray.FindPredicateFn`
```
(el, index, parent) -> !"none".equals(((HTMLElement) el).style.display)
```
is translated as
```
function GWTProjectEntryPoint$lambda$9$Type(){
}

defineClass(81, $wnd.Function, {}, GWTProjectEntryPoint$lambda$9$Type);
_.onInvoke_2 = function onInvoke_5(arg0, arg1, arg2){
  return $clinit_GWTProjectEntryPoint() , $clinit_Boolean() , $equals('none', castToNative(arg0, $wnd.HTMLElement).style.display)?false:true;
}
;
```
which looks like boxing of boolean to Boolean via inlining Boolean.valueOf(boolean) gone a bit wrong:
https://github.com/gwtproject/gwt/blob/c2229e7557450cd2842269f5fe9cb8586e90b50c/user/super/com/google/gwt/emul/java/lang/Boolean.java#L63-L65

```java
  private boolean shouldEnhanceLink(Element link) {
    return
        // Enhance only local links
        isSameOriginRexp.test(link.getAttribute("href"))
        // Do not load links that are marked as full page reload
        && !Boolean.parseBoolean(link.getAttribute("data-full-load"));
  }
```
is inlined into two different places, bringing `Boolean.parseBoolean` with it:

```javascript
function $lambda$16(this$static, evt_0){
  //...
  if ($test(isSameOriginRexp, target.getAttribute('href')) && ($clinit_Boolean() , !$equalsIgnoreCase('true', target.getAttribute('data-full-load'))) && $handleAsClick(castToNative(evt_0, $wnd.MouseEvent))) {
    //...
  }
}
```
```
function $lambda$19(el_0){
  //...
  if ($test(isSameOriginRexp, el_0.getAttribute('href')) && ($clinit_Boolean() , !$equalsIgnoreCase('true', el_0.getAttribute('data-full-load')))) {
    //...
  }
}
```
https://github.com/gwtproject/gwt/blob/c2229e7557450cd2842269f5fe9cb8586e90b50c/user/super/com/google/gwt/emul/java/lang/Boolean.java#L55-L57

##### Steps to reproduce
Check out [gwt-site-webapp @ 27070768ae49c048535358a03c70a5b9d84ddf98](https://github.com/gwtproject/gwt-site-webapp/tree/27070768ae49c048535358a03c70a5b9d84ddf98)
```shell
git clone git@github.com:gwtproject/gwt-site-webapp.git
cd gwt-site-webapp
git checkout 27070768ae49c048535358a03c70a5b9d84ddf98
```
Change draft compile to false
```patch
diff --git a/pom.xml b/pom.xml
index 747c5b2..f8fc02e 100644
--- a/pom.xml
+++ b/pom.xml
@@ -81,7 +81,7 @@
                     <compilerArgs>
                         <arg>-saveSource</arg>
                     </compilerArgs>
-                    <draftCompile>true</draftCompile>
+                    <draftCompile>false</draftCompile>
                     <style>PRETTY</style>
                     <failOnError>true</failOnError>
                     <launcherDir>${project.build.directory}/www</launcherDir>
```
Build the project, with `mvn clean verify`

Expected: If `$clinit_Boolean` is present, it is non-empty.
Actual: `$clinit_Boolean` is present, and only reassigns itself to emptyMethod

niloc132:
This will have minimal impacts in cases where JS inlining and static eval passes are able to make improvements, but will improve cases where Java/GWT passes can benefit from this method being removed, such as removing clinits.

The new implementation is effectively what jsinterop-base's Js.uncheckedCast() does, so is unlikely to cause any surprises.

Fixes #10055

### DIFF
--- a/dev/core/src/com/google/gwt/dev/jjs/impl/MethodInliner.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/impl/MethodInliner.java
@@ -134,6 +134,22 @@ public class MethodInliner {
         return;
       }
 
+      JMethod.Specialization specialization = getCurrentMethod().getSpecialization();
+      // If we have a specialization, don't inline that away - specializations must be called
+      // so they aren't pruned or type tightened into uselessness.
+      if (specialization != null) {
+        if (specialization.getTargetMethod() == method) {
+          return;
+        }
+        // We might have had a static impl that in turn will have a specialization - ensure we
+        // also don't inline that specialization away. Note that this check looks for it "backwards"
+        // by checking if the inlinable method has an instance method, since current method might
+        // have once had a static method, but it was already removed.
+        if (specialization.getTargetMethod() == program.instanceMethodForStaticImpl(method)) {
+          return;
+        }
+      }
+
       if (tryInlineMethodCall(x, ctx) == InlineResult.BLACKLIST) {
         // Do not try to inline this method again
         cannotInline.add(method);
--- a/user/super/com/google/gwt/emul/java/util/EnumSet.java
+++ b/user/super/com/google/gwt/emul/java/util/EnumSet.java
@@ -21,6 +21,7 @@ import static javaemul.internal.InternalPreconditions.checkNotNull;
 import static javaemul.internal.InternalPreconditions.checkState;
 
 import javaemul.internal.ArrayHelper;
+import javaemul.internal.JsUtils;
 import javaemul.internal.annotations.SpecializeMethod;
 
 /**
@@ -136,7 +137,7 @@ public abstract class EnumSet<E extends Enum<E>> extends AbstractSet<E> {
     @SpecializeMethod(params = Enum.class, target = "containsEnum")
     @Override
     public boolean contains(Object o) {
-      return (o instanceof Enum) && containsEnum((Enum) o);
+      return (o instanceof Enum) && containsEnum(JsUtils.uncheckedCast(o));
     }
 
     private boolean containsEnum(Enum e) {
@@ -151,7 +152,7 @@ public abstract class EnumSet<E extends Enum<E>> extends AbstractSet<E> {
     @SpecializeMethod(params = Enum.class, target = "removeEnum")
     @Override
     public boolean remove(Object o) {
-      return (o instanceof Enum) && removeEnum((Enum) o);
+      return (o instanceof Enum) && removeEnum(JsUtils.uncheckedCast(o));
     }
 
     private boolean removeEnum(Enum e) {
--- a/user/super/com/google/gwt/emul/javaemul/internal/JsUtils.java
+++ b/user/super/com/google/gwt/emul/javaemul/internal/JsUtils.java
@@ -82,9 +82,9 @@ public final class JsUtils {
   }-*/;
 
   @UncheckedCast
-  public static native <T> T uncheckedCast(@DoNotAutobox Object o) /*-{
-    return o;
-  }-*/;
+  public static <T> T uncheckedCast(@DoNotAutobox Object o) {
+    return (T) o;
+  }
 
   @UncheckedCast
   public static native <T> T getProperty(Object map, String key) /*-{
--- a/user/super/com/google/gwt/emul/javaemul/internal/annotations/SpecializeMethod.java
+++ b/user/super/com/google/gwt/emul/javaemul/internal/annotations/SpecializeMethod.java
@@ -22,6 +22,10 @@ import java.lang.annotation.Target;
  * An annotation to mark a given method as being specialized. If the specified
  * parameters and return context match of a JMethodCall, then the call
  * is retargeted at the specialized version.
+ * <p/>
+ * The annotated method must call the target method directly, and the compiler
+ * must not inline the target method to prevent it from being pruned while it
+ * still might be used.
  */
 @Target(ElementType.METHOD)
 @CompilerHint


niloc132:
Size changes in the included samples built by default are indeed meager - but it does depend on the app, and what other optimizations can take advantage of this improvement.
 * 89 bytes saved in Hello, where permutations are about 36KB
 * 618 bytes saved in DynaTable, where permutations are 113k-115KB
 * 687 bytes saved in JSON, where permutations are 80-81KB (almost 1%!)
 * 630 bytes saved in Mail, where permutations are 198-199KB
 * 779 bytes saved in Showcase, where permutations are 232-260KB

niloc132:
There are no tests for this, as I haven't yet worked out a good way to test it - perhaps a "full compile" test that lends itself well to being optimized away. I'm not sure this change really warrants a test, at least not of the complexity this might take to set up, but I will briefly give it a shot.

niloc132:
Test failure seems to be around a hashmap.put("Content-Type", "text/x-gwt-rpc") call that rpc makes to the requestbuilder - for some reason that is being optimized to have the key be null rather than the expected "Content-Type", so the header is missing when the call gets to the server as part of the test wiring.

This suggests that some overzealous optimization might be taking place, but I haven't yet pinned down what it is. I don't _think_ I made a mistake in this patch, unless somehow I needed a "this might not be null" hint? If so, why isn't jsinterop-base impacted by that?

My best guess so far is that it only applies to `@SpecializeMethod`-annotated methods. It looks like when using `@UncheckedCast` on a non-JSNI method, the TypeTightener gets more aggressive (right or wrong) about tightening types to null. Then a compiler pass or two later, the MethodCallSpecializer rewrites the map.put(key, value) call to map.putStringValue(key, value), but `key` has been tightened to always be null, so the data is lost.

niloc132:
More debugging - my hunch continues to be that SpecializeMethod is niche enough that it isn't as hardened as it should be, and JSNI methods have defied optimization so that method specialization hasn't broken before. Looking at a diff of an AST dump before and after this patch, we can see a number of ways that code gets cleaned up a little quicker with the jsinterop-base impl:
Simpler code:
```patch
@@ -50125,4 +50120,4 @@
       len = batchEnd - batchStart;
-      final Object[] spliceArgs = JsUtils.uncheckedCast(src).slice(batchStart, batchEnd);
-      JsUtils.uncheckedCast(spliceArgs).splice(0, 0, new Object[] {JsUtils.uncheckedCast((double) destOfs), JsUtils.uncheckedCast((double) (false ? len : 0))});
+      final Object[] spliceArgs = src.slice(batchStart, batchEnd);
+      spliceArgs.splice(0, 0, new Object[] {(double) destOfs, (double) (false ? len : 0)});
       ArrayHelper.getSpliceFunction().apply(destArray, spliceArgs);
```
Earlier inlining:
```patch
@@ -48992,7 +48987,7 @@
       case "string": 
-      return String.$hashCode(JsUtils.uncheckedCast(o));
+      return String.$hashCode(o);
       case "number": 
-      return Double.$hashCode(JsUtils.uncheckedCast(o));
+      return (int) JsUtils.unsafeCastToDouble((InternalPreconditions.$clinit(), InternalPreconditions.checkCriticalNotNull(o), o));
       case "boolean": 
-      return Boolean.$hashCode(JsUtils.uncheckedCast(o));
+      return JsUtils.unsafeCastToBoolean((InternalPreconditions.$clinit(), InternalPreconditions.checkCriticalNotNull(o), o)) ? 1231 : 1237;
       default: 
```

Background for our bug: `@SpecializeMethod` is an internal annotation that lets the compiler pick a more specific method based on what types are actually used. It behaves a bit like Java's own dynamic dispatch, but the compiler gets to change override not just based on the actual types, but the types that it works out by optimizing. Methods can also be selected that have different names, rather than being required to have the same name. 

In our case, AbstractHashMap has a few of these, the relevant one is for `put`:
https://github.com/gwtproject/gwt/blob/554447e9c741e961263c6fdcc86d24b9932dbc23/user/super/com/google/gwt/emul/java/util/AbstractHashMap.java#L211-L216
This not only has a typecheck to pick an implementation to use, but the compiler will actually rewrite calls to this second method if the first arg is able to be proven to be a string:
https://github.com/gwtproject/gwt/blob/554447e9c741e961263c6fdcc86d24b9932dbc23/user/super/com/google/gwt/emul/java/util/AbstractHashMap.java#L285-L292

Relevant AST dump diff for the bug this exposed - first, we see that content is being inlined quicker. This means that there are no longer any references to `$putStringValue` (the static version of `putStringValue` - `@SpecializeMethod` automatically applies to static implementations too).
```patch
@@ -50890,3 +50885,3 @@
   public Object put(Object key, Object value){
-    return key instanceof String ? AbstractHashMap.$putStringValue(this, JsUtils.uncheckedCast(key), value) : InternalHashCodeMap.$put(this.hashCodeMap, key, value);
+    return key instanceof String ? key == null ? InternalHashCodeMap.$put(this.hashCodeMap, null, value) : InternalStringMap.$put(this.stringMap, key, value) : InternalHashCodeMap.$put(this.hashCodeMap, key, value);
   }
```
The ternary is a little silly here (if key isnt a string, then it can't be null, so the first branch can literally never be taken). The rest all makes sense - but since nothing else calls $putStringValue any more, its params are "tightened" to nulls:

```patch
@@ -66242,4 +66204,4 @@
 
-  private static final Object $putStringValue(AbstractHashMap this$static, String key, Object value){
-    return key == null ? InternalHashCodeMap.$put(this$static.hashCodeMap, null, value) : InternalStringMap.$put(this$static.stringMap, key, value);
+  private static final Object $putStringValue(AbstractHashMap this$static, null key, null value){
+    return InternalHashCodeMap.$put(this$static.hashCodeMap, null, value);
   }
```
(Also note that because it did "successfully" tighten to null, more of the method goes away. This might make the size changes suspect in my comment earlier).

The "right" answer here is probably to just remove the unused method, but I guess this serves as an intermediate step?

From the order of operations that the compiler does when optimizing, note that MethodCallSpecializer is before MethodInlining specific to prevent this case:
https://github.com/gwtproject/gwt/blob/554447e9c741e961263c6fdcc86d24b9932dbc23/dev/core/src/com/google/gwt/dev/jjs/JavaToJavaScriptCompiler.java#L1498-L1506
Finalizer and MakeCallsStatic run first, which will give our static methods a matching static target once they are inlined later. For some reason, we're missing specializing from `put` to `putStringValue` on the first pass (both before and after this change), but on the second pass we can do it. 

Both before and after end up with the same optimization for the original map.put() call, the last line of this:
```java
   public static final void $setHeader(RequestBuilder this$static, String header, String value){
     StringValidator.throwIfEmptyOrNull("header", header);
     StringValidator.throwIfEmptyOrNull("value", value);
     this$static.headers == null && (this$static.headers = new HashMap());
     AbstractHashMap.$putStringValue(this$static.headers, header, value);
   }
```
But since in our "after" code, the key is ignored (assumed to always be null), the map doesn't contain the data where we expect.

The "just hack it" part of me wants to just rearrange the compiler passes, so that we tighten types after specialization, or only inline after a few rounds of the other compilation steps to prevent this (as inlining both exposes the uncheckedCast issue and removes the callsites for the specialized impl). The existing interrelationships between compiler passes are poorly documented as it stands though, as this bug seems to reinforce. 

(With that said... reordering does fix this one test at least.)

niloc132:
The fundamental issue is that we're changing the specialization before we get a chance to actually use it. This can safely be prevented by _already using it_. The base method starts off by using it, but then optimizations end up rewriting it away (specifically inlining, then type tightening to make it incompatible).

This is suggested in a comment for a test, but not outright asserted elsewhere that I can find:
https://github.com/gwtproject/gwt/blob/7c36c8922b0cda7880ff31cfc0135b910258b8fb/user/test/com/google/gwt/dev/jjs/optimized/SpecializationTest.java#L44-L48

The fix then seems to be to forbid inlining specializations into their own base method. There are two facets to this - we need to handle not only "inline putStringValue into put), but also "inline $putStringValue into put". This latter case can happen when the compiler has produced staticified versions of both put and putStringValue (where $put contains a call then to $putStringValue), then inlined $put into put. Inlining $putStringValue into $put is handled by the first case.

This change does seem to suggest that after specializations are unmarked (when the optimization loop is finished), it makes sense to run the inliner again before pruning, I'll explore that next (adding it to the obvious location fails the compile).

I've also added explicit notes in `@SpecializeMethod` to indicate that decorated methods must call their specialized impl - this was already done in all places where it is used. I've also tweaked a few usages of `@SpecializeMethod` to use uncheckedCast to avoid an unnecessary type check (Java 16 instanceof patterns are the "right" fix for this, but they generate a bit more code, and can't be always turned on at this time).

Gains are down with the updated patch, about 100 bytes for everything except Hello, which is still 89 bytes.

vjay82:
Just as a short feedback. In our project (compiled with -style OBFUSCATED, -optimize 9 and -XdisableCastChecking) the output even grows a little. 
My suspicion is that for example "function u(a,b){return a.indexOf(b);}" becomes more efficient over time if used a lot (length of "u" instead of "indexOf"). Probably still a good thing to shift more workload from V8 to compression.
Aside from that, all tests passed in our project.

niloc132:
Thanks! Inlining is tricky to get right, and GWT doesn't take into account the obfuscated length of the method names vs inlined JS native calls, though perhaps it should. 

The big advantage of inlining of course is that it can often allow other optimizations later by specializing for the method it was inlined into - though in this case, it never will. 

Note that -XdisableCastChecking is deprecated since 2.8.0, and jre.checks.checkLevel should be used instead (values are "NORMAL", "OPTIMIZED", and "MINIMAL"), see InternalPreconditions for more info https://github.com/gwtproject/gwt/blob/7c36c8922b0cda7880ff31cfc0135b910258b8fb/user/super/com/google/gwt/emul/javaemul/internal/InternalPreconditions.java#L23-L73

Getting distracted from the fix itself, you mentioned compression - are you compressing your JS (GWT can do it at compile time if you don't mind gzip, and your server likely can do brotli), and are you doing other things that will decrease output size like compiler.stackMode=strip (allows deduplicating some functions where their contents are the same)?

vjay82:
I also think, this is a cleaner approach.

Distraction: Thank you for the hint. Yes, setting that too, just forgot to mention it, as it is hidden away in the module.xml (and not in the build script itself). We are using native stack traces and utilizing ZStd, brotli or gzip. Also we remove some strings from lambda clinit (which you are probably addressing more efficiently very soon when looking at the issues-tab) in a post-compile-step. 

There are other low-hanging fruits in the output of the iframe linker (lots of escaped line-breaks) but removing those messes up stack traces of course so not doing that. I asked myself, why things have to go through eval in the first place (nowadays).

Also thanks for the continued development.

niloc132:
I'm doing a little more exploring in that area now - if you have other observations that seem appropriate to files as bugs, I'd welcome that.

I went digging into the newline thing - there are some legacy reasons why certain output decisions had to be made (limit identifiers in a `var` for an old safari bug, etc), but it looks like the newline thing doesn't exist for a particular reason. I found https://github.com/gwtproject/gwt/issues/986 discussing this and the matching commit e5fbf1c1adf3509f792030f54090187e94ed75c4 that purports to fix it, yet always adds newlines after function statements (where the `}` should be sufficient always). With that said, 442cf3399b8b6a2537280ba4d336bf8ece54aeed discusses how sourcemap size might be impacted by such a change, so we might want to let this be configurable.

Clinit removal works ... except for super dev mode, so it needs more work. This other works is related only in that I got irritated at seeing yet another empty clinit decl and double-clinit call and wanted to chase them down, see if other improvements could pop out too.

Regarding indexOf, there are two small improvements I have from looking at the apps I have readily available to me:
 * JS String.includes is a thing in all browsers now, which at least saves bytes in the `s.indexOf(h)!=-1` case
 * Java String.indexOf(char) with a literal char optimizes pretty badly, since GWT will have to first turn the literal char into an int, the int into a string via JS `String.fromCharCode` - if we statically know it is a `char` to begin with, we could just compile it to a String. For example, java `s.indexOf('.')` is the same as js `s.indexOf('.')`) such as in https://github.com/gwtproject/gwt/blob/7c36c8922b0cda7880ff31cfc0135b910258b8fb/user/super/com/google/gwt/emul/java/util/logging/LogManager.java#L72 In order to make this actually work though, `String.valueOf(char)` cannot be inlined until much later (if ever), since otherwise the inlining can end up happening before the static eval from char to string happens. On the plus side, this more or less prevents another occurrence of your indexOf issue above.

Okay, probably enough distraction here. File issues you'd like to see get worked on, and we can collaborate as time/effort permits to make them happen.

vegegoku (approved):

PR #10166, opened 2025-10-22
Enhances I18NCreator for creating Message interfaces from properties

tsschmidt:
Currently I18NCreator creates method signatures that only use String arguments regardless of any MessageFormat that may be present for an argument. This patch attempts to create more accurate interface signatures based on the MessageFormat and also includes some GWT specific formats such as {0,list},{0,localdatetime,..} and Plurals.

This patch also proposes to add the convention {0,safehtml} to mark that an argument to be of type SafeHtml.  If an Argument is found to be of type SafeHtml the return type of the method will be set to SafeHtml.

Fixes #9049

### DIFF
--- a/user/src/com/google/gwt/i18n/rebind/AbstractLocalizableInterfaceCreator.java
+++ b/user/src/com/google/gwt/i18n/rebind/AbstractLocalizableInterfaceCreator.java
@@ -175,16 +175,6 @@ public abstract class AbstractLocalizableInterfaceCreator {
   }
 
   /**
-   * Create a String method declaration from a Dictionary/value pair.
-   * 
-   * @param key Dictionary
-   * @param defaultValue default value
-   */
-  public void genSimpleMethodDecl(String key, String defaultValue) {
-    genMethodDecl("String", defaultValue, key);
-  }
-
-  /**
    * Create method args based upon the default value.
    * 
    * @param defaultValue
@@ -250,11 +240,15 @@ public abstract class AbstractLocalizableInterfaceCreator {
               + resourceFile
               + "' cannot be used to generate message classes, as it has no key/value pairs defined.");
     }
-    for (String key : keys) {
-      String value = p.getProperty(key);
-      genSimpleMethodDecl(key, value);
-    }
+    generateMethods(p, keys);
     composer.commit(new PrintWriterTreeLogger());
+  }
+
+  void generateMethods(LocalizedProperties properties, String[] keys) {
+    for (String key : keys) {
+      String value = properties.getProperty(key);
+      genMethodDecl(key, value);
+    }
   }
 
   private void addFormatters() {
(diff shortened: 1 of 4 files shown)


user/src/com/google/gwt/i18n/rebind/AbstractLocalizableInterfaceCreator.java
```diff
@@ -281,12 +275,12 @@ private void genMethodDecl(String type, String defaultValue, String key) {
...
     composer.println("@Key(" + makeJavaString(key) + ")");
     String methodName = formatKey(key);
-    composer.print(type + " " + methodName);
+    composer.print("String " + methodName);
     composer.print("(");
     genMethodArgs(defaultValue);
     composer.print(");\n");
   }
-
+
```
niloc132: Extra spaces still here I'm afraid
```suggestion

```

user/src/com/google/gwt/i18n/rebind/MessagesInterfaceCreator.java
```diff
@@ -100,4 +136,96 @@ protected String javaDocComment(String path) {
...
+      }
+    }
+    return "String";
+  }
+
+  private void genPluralsAnnotation(Map<String, String> plurals) {
+    composer.print("@AlternateMessage({");
+    String[] keys = plurals.keySet().toArray(new String[] {});
+    if (keys.length > 1) {
+      composer.println("");
```
niloc132: I think this is just
```suggestion
      composer.println();
```
Same for below

user/src/com/google/gwt/i18n/rebind/MessagesInterfaceCreator.java
```diff
@@ -100,4 +136,96 @@ protected String javaDocComment(String path) {
...
+      composer.indent();
+    }
+    for (int i = 0; i < keys.length; i++) {
+      String key = keys[i];
+      if (i > 0) {
+        composer.println(",");
+      }
+      composer.print("\"" + key.substring(key.indexOf('[') + 1, key.length() - 1) + "\", ");
+      composer.print("\"" + plurals.get(key) + "\"");
```
niloc132: Probably should use makeJavaString here - in theory the keys never have strings in them, but I _think_ it is technically allowed for properties files to have spaces in their keys if escaped?

niloc132 (commented):
This appears to be in the style of other existing Creator types, so I can't fault it there, but I would generally like the test to be a bit more than 'it can run and doesn't fail in obvious ways' - at a glance, it doesn't even check if the generated sources compile?

A few things to consider below - thank you for getting this patch going again, I'd be inclined to merge it without further _additional_ delay. If you'd like a more rigorous review, I'd be happy to.

tsschmidt:
Updated with recommendations from comments above.  I am not inclined to take on rewriting Creator tests.  Let me know if anything else needs updating to current style or implementation guidelines.

user/src/com/google/gwt/i18n/rebind/MessagesInterfaceCreator.java
```diff
@@ -100,4 +136,99 @@ protected String javaDocComment(String path) {
...
+  
+  private void determineNumberType(String subFormat) {
+    switch (subFormat) {
+      case "integer":
+        composer.print("Integer");
+        break;
+      case "currency":
+      case "percent":
+      default:
+        if(subFormat.contains(".")) {
```
niloc132: checkstyle is asking for a space here
```suggestion
        if (subFormat.contains(".")) {
```

user/src/com/google/gwt/i18n/rebind/AbstractLocalizableInterfaceCreator.java
```diff
@@ -250,11 +240,15 @@ void generateFromPropertiesFile() throws IOException {
               + resourceFile
               + "' cannot be used to generate message classes, as it has no key/value pairs defined.");
```
niloc132, line 241: this one is outside your diff, so you can ignore it

niloc132 (approved):
Tests are passing - approving, will leave it up a day or so for other comment.

PR #10169, opened 2025-10-27
Add permissions for publishing JUnit report

zbynek:
Based on suggestion in https://github.com/mikepenz/action-junit-report?tab=readme-ov-file#pr-run-permissions

Recent PRs marked as `ready` run the JUnit step, but are missing the report because of permissions. This PR should fix that.

### DIFF
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -15,6 +15,9 @@ on:
     # Allow running manually
   pull_request:
     types: [ labeled, synchronize ]
+permissions:
+  # allow publishing JUnit report for PRs
+  checks: write 
 jobs:
   build:
     if: ${{ !github.event.pull_request || contains(github.event.pull_request.labels.*.name, 'ready') }}


niloc132:
Note that for reviewdog, we follow the other path, and use annotations instead. The linked doc indicates that this action also supports creating annotations for failed tests:
```yaml
        with:
          annotate_only: true # forked repo cannot write to checks so just do annotations
```

zbynek:
Annotations are great if the annotated files are part of the PR (always the case with checkstyle), but I'm not sure it's possible to use them in case of failing unit tests that didn't change (which is the case for most? test failures).

niloc132:
That's a great point - and it looks like stack traces are lost in annotations too: https://github.com/mikepenz/action-junit-report/issues/1278

niloc132 (approved):

niloc132:
Looks like these still aren't working:
https://github.com/gwtproject/gwt/actions/runs/19096338630?pr=10176
image
image

Issue #9909, opened 2024-01-29
Explore options to update `long` emulation


niloc132:
GWT's existing long emulation was [last updated in 2015](https://github.com/gwtproject/gwt/commit/e31f9a4afaf), with a few assumptions that might no longer be true. Among other things, it retains the earlier assumption that 24bit ints are faster to operate on than larger values. Unfortunately, there appear to be no public benchmarks (the old gwt-benchmark repo only has add/multiply benchmarks, and add is all "Small" values), so we might need to work up some acceptable examples to do this work.

Contrast with J2CL's current implementation of `long`, which uses [closure-library's own long.js](https://github.com/google/closure-library/blob/7818ff7dc0b53555a7fb3c3427e6761e88bde3a2/closure/goog/math/long.js#L19-L39) (intended to reflect java long behavior as well). This implementation uses two 32bit values, occasionally splitting to 16bit values for operations like add/multiply/etc (though not bitwise operations), whereas GWT operates on 13 bit values, and appears to end up doing more work.

Also consider [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) - with the demise of IE11, this should be well supported. This may require introducing changes to the JS AST or JSNI parser.

niloc132:
From https://github.com/gwtproject/gwt/pull/10176, also consider handling Java 9's Math.multiplyHigh, Math.fma, and possibly Math.unsignedMultiplyHigh if other Java 18-21 support has been added.

niloc132:
A [recent issue to the J2CL repo](https://github.com/google/j2cl/issues/302#issue-5503546134) could point out some beneficial directions to consider, quoted here:

> I am the author and lead maintainer of Scala.js. For years, we've been optimizing our implementation of longs on JavaScript. Our efforts recently culminated in a paper with our best algorithms:
> 
>     S. Doeraene and T. Schlatter,
>     "64-bit Integer Division for the JavaScript Platform,"
>     33rd IEEE Symposium on Computer Arithmetic (ARITH), Fulda, Germany, 2026.
>     https://arith2026.org/papers/64-bit%20Integer%20Division%20for%20the%20JavaScript%20Platform.pdf
> 
> Our implementation can be found at
> https://github.com/scala-js/scala-js/blob/v1.22.0/linker-private-library/src/main/scala/org/scalajs/linker/runtime/RuntimeLong.scala
> as well as, for conversion to string with arbitrary radices,
[...]

PR #10176, opened 2025-11-05
Math emulation improvements

### LINKED ISSUE #9663: Add support for Math.IEEEremainder
**GWT version:** 
**Browser (with version):** 
**Operating System:** 

---

##### Description

Add support for: https://docs.oracle.com/javase/7/docs/api/java/lang/Math.html#IEEEremainder(double,%20double)

### LINKED ISSUE #9711: log1p(x) and expm1(x) innacurate for small x
At the moment we have naïve versions of expm1() and log1p()
https://github.com/gwtproject/gwt/blob/master/user/super/com/google/gwt/emul/java/lang/Math.java

```
public static double expm1(double d) {
    return d == 0 ? d : exp(d) - 1;
  }
```

```
 public static double log1p(double x) {
    return x == 0 ? x : log(x + 1);
  }
```

Where possible (ie basically everywhere except IE11) the native browser versions should be used
https://caniuse.com/?search=expm1
https://caniuse.com/?search=lop1p

See here for why that's important (TLDR: for small x)
https://scicomp.stackexchange.com/questions/20629/when-should-log1p-and-expm1-be-used
https://www.johndcook.com/blog/cpp_expm1/

zbynek:
Fixes #9711 by using native emulation of `log1p`, `expm1` and `hypot`.
Fixes #9663 by adding the implementation of `IEEEremainder`
Adds `absExact`,  `multiplyFull`, `ulp`, and `getExponent` + missing variations of `multiplyExact`, `floorDiv`, `floorMod` for completeness.

Only missing methods for JVM 17 compatibility are `multiplyHigh` and `fma` which might be better handled in #9909


user/test-super/com/google/gwt/emultest/super/com/google/gwt/emultest/java17/lang/MathTest.java
```diff
@@ -0,0 +1,46 @@
...
+ * the License.
+ */
+package com.google.gwt.emultest.java17.lang;
+
+import com.google.gwt.emultest.java.util.EmulTestBase;
+
+/**
+ * Tests for java.lang.String Java 12 API emulation.
+ */
+public class StringTest extends EmulTestBase {
```
niloc132:
```suggestion
public class MathTest extends EmulTestBase {
```

user/test-super/com/google/gwt/emultest/super/com/google/gwt/emultest/java17/lang/MathTest.java
```diff
@@ -0,0 +1,46 @@
...
+  public void testAbsExactWithFolding() {
+    assertEquals(Integer.MAX_VALUE, Math.absExact(Integer.MAX_VALUE));
+    assertEquals(0, Math.absExact(0));
+    assertEquals(1, Math.absExact(-1));
+    assertThrowsArithmetic(() -> Math.absExact(Integer.MIN_VALUE));
+    assertEquals(Long.MAX_VALUE, Math.absExact(Long.MAX_VALUE));
+    assertEquals(0, Math.absExact(0L));
+    assertEquals(1, Math.absExact(-1L));
+    assertThrowsArithmetic(() -> Math.absExact(Long.MIN_VALUE));
+  }
```
niloc132, line 45:
```suggestion
  }

  private <T> T hideFromCompiler(T value) {
    if (Math.random() < -1) {
      // Can never happen, but fools the compiler enough not to optimize this call.
      fail();
    }
    return value;
  }

  private void assertThrowsArithmetic(Runnable check) {
    try {
      check.run();
      fail("Should have failed");
    } catch (ArithmeticException ex) {
      // good
    }
  }
```

user/test-super/com/google/gwt/emultest/super/com/google/gwt/emultest/java17/lang/MathTest.java
```diff
@@ -0,0 +1,46 @@
...
+}
```
niloc132:
```suggestion
}

```

user/super/com/google/gwt/emul/java/lang/Math.java
```diff
@@ -31,7 +32,6 @@ public final class Math {
   //   public static float ulp (float x)
   //   public static int getExponent (double d)
   //   public static int getExponent (float f)
-  //   public static double IEEEremainder(double f1, double f2)
```
niloc132, line 34: I think you can remove the getExponent calls too

user/super/com/google/gwt/emul/javaemul/internal/JsUtils.java
```diff
@@ -123,10 +123,14 @@ public static float intBitsToFloat(int value) {
     return JsUtils.<float[]>uncheckedCast(new Float32Array(buf))[0];
   }
 
-  public static long doubleToRawLongBits(double value) {
+  public static int[] doubleToRawIntBits(double value) {
     ArrayBuffer buf = new ArrayBuffer(8);
     JsUtils.<double[]>uncheckedCast(new Float64Array(buf))[0] = value;
-    int[] intBits = JsUtils.<int[]>uncheckedCast(new Uint32Array(buf));
+    return JsUtils.<int[]>uncheckedCast(new Uint32Array(buf));
```
niloc132, line 129: Why not just change the return type to Uint32Array?

...Why didn't the original code do this?

Also, any signature change is going to break Double.java:
https://github.com/gwtproject/gwt/blob/cfb4e79167db6f8704d6a9f653c57351b76fe036/user/super/com/google/gwt/emul/java/lang/Double.java#L64-L76

zbynek: > Why not just change the return type to Uint32Array

Are you suggesting to use the `UInt32Array` defined in `JsUtils` (+ implementing `get`) or changing it to the existing `Uint32ArrayNative` which has `get` already implemented (but creates a dependency between packages)?

> any signature change is going to break Double.java

I didn't change any signature here, `doubleToRawLongBits` is still returning `long`.

niloc132:
While reviewing code elsewhere, I also noticed that DoubleTest.testDoubleConstants() has two checks that can be reenabled if we support getExponent() now

user/super/com/google/gwt/emul/javaemul/internal/JsUtils.java
niloc132: Thanks, both make sense - I wasn't keeping track of which UInt32Array was in use here.

dev/core/src/com/google/gwt/dev/js/JsToStringGenerationVisitor.java
```diff
@@ -84,6 +84,7 @@
 /**
  * Produces text output from a JavaScript AST.
  */
+@SuppressWarnings("checkstyle.MethodName")
```
niloc132: Maybe we land these other checkstyle changes in a separate, earlier PR?

The one thing thats nice about the comments for a 1000line class like this is that only a third of the file needs special treatment, though I'm sure we don't want to add dozens of annotations. 

...do we really save that much effort by just using `_` instead of a prefix like `print` or `p`? This seems silly.

This is definitely load-bearing code though, this is used to generate our actual output JS, so some caution is warranted.

dev/core/src/com/google/gwt/dev/cfg/ModuleDefSchema.java
```diff
@@ -301,8 +300,8 @@ protected Schema __collapse_property_begin(PropertyName name,
 
     @SuppressWarnings("unused") // called reflectively
     protected Schema __define_configuration_property_begin(PropertyName name,
-        String is_multi_valued) throws UnableToCompleteException {
-      boolean isMultiValued = toPrimitiveBoolean(is_multi_valued);
+        String isMultiValuedStr) throws UnableToCompleteException {
+      boolean isMultiValued = toPrimitiveBoolean(isMultiValuedStr);
```
niloc132: note that reflective magic is used here to build from module files, and i _thought_ that params were picked up reflectively too so naming matters? I'm happy to see that there are no errors though - there should be at least one test using the multi value flag.

user/super/com/google/gwt/emul/java/lang/Math.java
```diff
@@ -31,7 +32,6 @@ public final class Math {
   //   public static float ulp (float x)
   //   public static int getExponent (double d)
   //   public static int getExponent (float f)
```
niloc132:
```suggestion
```

user/test/com/google/gwt/emultest/java8/lang/MathTest.java
```diff
@@ -294,6 +334,172 @@ public void testToIntExact() {
...
+    assertTrue(Double.isNaN(Math.IEEEremainder(hideFromCompiler(Double.NaN), 2.0)));
+    assertTrue(Double.isNaN(Math.IEEEremainder(hideFromCompiler(5.0), Double.NaN)));
+    assertTrue(Double.isNaN(Math.IEEEremainder(hideFromCompiler(Double.NaN), Double.NaN)));
+  }
+
+  public void testHypot() {
+    assertEquals(5.0, Math.hypot(3.0, 4.0), EPS);
+    assertEquals(Double.POSITIVE_INFINITY, Math.hypot(1, Double.NEGATIVE_INFINITY));
+    assertEquals(Double.POSITIVE_INFINITY, Math.hypot(Double.POSITIVE_INFINITY, 1));
+    assertEquals(Double.POSITIVE_INFINITY, Math.hypot(Double.NaN, Double.NEGATIVE_INFINITY));
```
niloc132: warning, these are boxing to Double without the third arg (either 0 or EPS in other lines)

user/test/com/google/gwt/emultest/java8/lang/MathTest.java
```diff
@@ -294,6 +334,244 @@ public void testToIntExact() {
...
+    assertEquals(5.0, Math.hypot(hideFromCompiler(3.0), 4.0), EPS);
+    assertEquals(Double.POSITIVE_INFINITY, Math.hypot(hideFromCompiler(-1),
+        Double.NEGATIVE_INFINITY), EPS);
+    assertEquals(Double.POSITIVE_INFINITY, Math.hypot(hideFromCompiler(Double.POSITIVE_INFINITY),
+        1), EPS);
+    assertEquals(Double.POSITIVE_INFINITY, Math.hypot(hideFromCompiler(Double.NaN),
+        Double.NEGATIVE_INFINITY), EPS);
+  }
+
+  public void testGetExponentWithFolding() {
```
niloc132, line 510: aside, constant folding doesn't work on non-String methods at this time https://github.com/gwtproject/gwt/issues/10147

niloc132 (approved):

Issue #9947, opened 2024-03-29
Spurious incremental CompilerTest failure


niloc132:
https://github.com/gwtproject/gwt/actions/runs/8483890223/attempts/1

I believe I've seen this at least once before - something is going wrong either in the compiler or in the test code, causing `com.google.gwt.dev.CompilerTest` to fail "randomly". Every test method that validates incremental changes ends up failing with a stack trace something like this:

```
  <testcase classname="com.google.gwt.dev.CompilerTest" name="testChangeJsType" time="2.888">
    <failure message="expected:&lt;[com.foo.Bar, com.foo.Foo]&gt; but was:&lt;[java.util.Map, com.foo.Bar, java.util.Set, com.foo.Foo, java.util.List]&gt;" type="junit.framework.AssertionFailedError">junit.framework.AssertionFailedError: expected:&lt;[com.foo.Bar, com.foo.Foo]&gt; but was:&lt;[java.util.Map, com.foo.Bar, java.util.Set, com.foo.Foo, java.util.List]&gt;
	at com.google.gwt.dev.CompilerTest.compileToJs(CompilerTest.java:2739)
	at com.google.gwt.dev.CompilerTest.compileToJs(CompilerTest.java:2677)
	at com.google.gwt.dev.CompilerTest.checkRecompiledModifiedApp(CompilerTest.java:2637)
	at com.google.gwt.dev.CompilerTest.testChangeJsType(CompilerTest.java:1252)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
</failure>
  </testcase>
```

Every failure looks like this - as if Set, Map, and List had somehow also changed through simulated user actions. 

Every failure included those three classes (I didn't see any others, but didn't exhaustively search), and every failure was specific to recompilation. Not all recompilation tests failed in this way though, for example `testIncrementalRecompile_bridgeMethodOverrideChain` and `testIncrementalRecompile_jsPropertyConsistencyCheck` did not fail in this way. 

Failing CompilerTest.xml attached (but renamed with .txt suffix).
[...]

Issue #9731, opened 2021-06-10
dead code not eliminated when two Boolean.parseBoolean in same function


arenevier:
Hi,

In some circumstances, unreachable code is not removed properly.

I am adding a minimal reproducible testcase:

```
package com.google.gwt.sample.hello.client;
    
import com.google.gwt.core.client.EntryPoint;
import com.google.gwt.user.client.Window;

abstract class BaseDelegate {
  abstract public void display();
} 

class DelegateFoo extends BaseDelegate {
  @Override
  public void display() {
    Window.alert("Foo");
  }
} 
  
class DelegateNoFoo extends BaseDelegate {
  @Override
  public void display() {
    Window.alert("No Foo");
  }
}
  
public class Hello implements EntryPoint {
  static boolean isFeatureFooEnabled() {
    if (Boolean.parseBoolean(System.getProperty("myapp.property1"))) {
      return true;
    } 
    if (Boolean.parseBoolean(System.getProperty("myapp.property2"))) {
      return true;
    }
    return false;
  }

  public void onModuleLoad() {
    final BaseDelegate delegate = isFeatureFooEnabled() ? new DelegateFoo() : new DelegateNoFoo();
    delegate.display();
  }
} 
```

If **myapp.property1** and **myapp.property2** are both set to **false**, the above code will compile to
```
  delegate = ($clinit_Boolean() , $clinit_Boolean() , false)?new DelegateFoo:new DelegateNoFoo;
  delegate.display();
```

So it means that even if DelegateNoFoo is never used, its code will not be removed. It's a constraint example, but in our application, those delegates are actually pretty big and depend on many other classes. It means that we end up building and shipping a lot of unused code.

If there is only one Boolean.parseBoolean, the code compiles (correctly) to 
```delegate = ($clinit_Boolean() , new DelegateNoFoo);```

Also, if we modify isFeatureFooEnabled to 
```
  static boolean isFeatureFooEnabled() {
    if (Boolean.parseBoolean(System.getProperty("myapp.property1"))) {
      return true;
    }
    return Boolean.parseBoolean(System.getProperty("myapp.property2"));
  }
```

It will compile to ```delegate = ($clinit_Boolean() , $clinit_Boolean() , new DelegateNoFoo);```

But this is fragile and won't scale if **isFeatureFooEnabled()** becomes more complex

niloc132:
The linked issue #9922 should comprehensively solve this for the case of boolean, but there are still other oddities that it would be good to improve. 

[...]

PR #10177, opened 2025-11-06
Enhance AstDumper for debugging and testing

niloc132:
Enhances the AstDumper utility class with the aim of bringing back some of the functionality previously removed in 64f9b00adb7e29f9247b95bd0706823de1472e6b, where individual methods could be traced from source Java all the way to JS output. This is a small step in that direction that nevertheless provides more insight into what the compiler is doing, through a series of changes:
 * Improve printed Java output - classes should list their name, enums and records should be labeled as such, etc. AST traversal is made more visitor-like to make filtering easier.
 * JS AST should also be printed, rather than ending with the Java optimization loops, allowing insight into JS passes.
 * A filter can be applied to the ast at any stage, resulting in only printing code from the specified files or lines, to more easily follow specific changes through a program.

More improvement remains - multiple permutations may end up mixed in the same file, different compiler passes are not individually listed or labeled, arguments are re-parsed every time they are used, but impact to the API is minimal for now.

### DIFF
--- a/dev/core/src/com/google/gwt/dev/jjs/impl/AstDumper.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/impl/AstDumper.java
@@ -16,13 +16,71 @@
 
 package com.google.gwt.dev.jjs.impl;
 
+import com.google.gwt.dev.jjs.HasSourceInfo;
+import com.google.gwt.dev.jjs.ast.Context;
+import com.google.gwt.dev.jjs.ast.JClassType;
+import com.google.gwt.dev.jjs.ast.JInterfaceType;
+import com.google.gwt.dev.jjs.ast.JMethod;
+import com.google.gwt.dev.jjs.ast.JNode;
 import com.google.gwt.dev.jjs.ast.JProgram;
+import com.google.gwt.dev.jjs.ast.JVisitor;
+import com.google.gwt.dev.js.JsSourceGenerationVisitor;
+import com.google.gwt.dev.js.ast.JsArrayAccess;
+import com.google.gwt.dev.js.ast.JsArrayLiteral;
+import com.google.gwt.dev.js.ast.JsBinaryOperation;
+import com.google.gwt.dev.js.ast.JsBlock;
+import com.google.gwt.dev.js.ast.JsBooleanLiteral;
+import com.google.gwt.dev.js.ast.JsBreak;
+import com.google.gwt.dev.js.ast.JsCase;
+import com.google.gwt.dev.js.ast.JsCatch;
+import com.google.gwt.dev.js.ast.JsConditional;
+import com.google.gwt.dev.js.ast.JsContext;
+import com.google.gwt.dev.js.ast.JsContinue;
+import com.google.gwt.dev.js.ast.JsDebugger;
+import com.google.gwt.dev.js.ast.JsDefault;
+import com.google.gwt.dev.js.ast.JsDoWhile;
+import com.google.gwt.dev.js.ast.JsEmpty;
+import com.google.gwt.dev.js.ast.JsExprStmt;
+import com.google.gwt.dev.js.ast.JsFor;
+import com.google.gwt.dev.js.ast.JsForIn;
+import com.google.gwt.dev.js.ast.JsFunction;
+import com.google.gwt.dev.js.ast.JsIf;
+import com.google.gwt.dev.js.ast.JsInvocation;
+import com.google.gwt.dev.js.ast.JsLabel;
+import com.google.gwt.dev.js.ast.JsNameOf;
+import com.google.gwt.dev.js.ast.JsNameRef;
+import com.google.gwt.dev.js.ast.JsNew;
+import com.google.gwt.dev.js.ast.JsNode;
+import com.google.gwt.dev.js.ast.JsNullLiteral;
+import com.google.gwt.dev.js.ast.JsNumberLiteral;
+import com.google.gwt.dev.js.ast.JsNumericEntry;
+import com.google.gwt.dev.js.ast.JsObjectLiteral;
+import com.google.gwt.dev.js.ast.JsParameter;
+import com.google.gwt.dev.js.ast.JsPositionMarker;
+import com.google.gwt.dev.js.ast.JsPostfixOperation;
+import com.google.gwt.dev.js.ast.JsPrefixOperation;
+import com.google.gwt.dev.js.ast.JsProgram;
+import com.google.gwt.dev.js.ast.JsProgramFragment;
+import com.google.gwt.dev.js.ast.JsPropertyInitializer;
+import com.google.gwt.dev.js.ast.JsRegExp;
+import com.google.gwt.dev.js.ast.JsReturn;
+import com.google.gwt.dev.js.ast.JsStringLiteral;
+import com.google.gwt.dev.js.ast.JsSwitch;
+import com.google.gwt.dev.js.ast.JsThisRef;
+import com.google.gwt.dev.js.ast.JsThrow;
+import com.google.gwt.dev.js.ast.JsTry;
+import com.google.gwt.dev.js.ast.JsVars;
+import com.google.gwt.dev.js.ast.JsVisitor;
+import com.google.gwt.dev.js.ast.JsWhile;
 import com.google.gwt.dev.util.AbstractTextOutput;
 import com.google.gwt.dev.util.TextOutput;
 
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.PrintWriter;
+import java.util.Arrays;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 /**
  * A simple utility to dump a JProgram to a temp file, which can be called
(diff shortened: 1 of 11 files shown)


niloc132:
Example of generating an AST dump with JS output included, while working on a fix for #9731 - running a build of the Hello sample before and after the change, and running `diff -U10 before.dump after.dump` results in this sample output (demonstrating two ways that the code is improving):
```patch
 function Button(handler){
   $clinit_FocusWidget();
   ButtonBase.call(this, $createPushButtonElement($doc));
   ($clinit_DOM() , this.element).className = 'gwt-Button' || '';
-  ($clinit_DOM() , this.element).innerHTML = 'Click me' || '';
-  $addDomHandler(this, handler, ($clinit_ClickEvent() , $clinit_ClickEvent() , TYPE));
+  this.element.innerHTML = 'Click me' || '';
+  $addDomHandler(this, handler, ($clinit_ClickEvent() , TYPE));
 }
```

niloc132:
CompilerTest doesn't fail locally, investigating...

niloc132:
Failures appear to be an example of https://github.com/gwtproject/gwt/issues/9947, rerunning.

zbynek (dismissed):

dev/core/src/com/google/gwt/dev/jjs/impl/ToStringGenerationVisitor.java
```diff
@@ -401,11 +415,12 @@ public boolean visit(JDebuggerStatement x, Context ctx) {
 
   @Override
   public boolean visit(JDeclarationStatement x, Context ctx) {
-    if (!suppressType) {
```
zbynek, line 404: This makes suppressType unread.

dev/core/src/com/google/gwt/dev/jjs/impl/AstDumper.java
```diff
@@ -80,25 +106,243 @@ public static void maybeDumpAST(JProgram jprogram, String versionString) {
...
+   * Filters the AST by top-level node to determine what the delegate can see, by checking if any
+   * code is present from the requested files. Top level nodes are usually functions are class
+   * declaration code.
+   */
+  private static class JsFilteredAstVisitor extends JsSuperVisitor {
+    private final JsVisitor delegate;
+    private final Set<FilterRange> sourceFiles;
+
+    private boolean shouldVisitCurrentTopLevelStatement = false;
+
```
niloc132, line 306:
```suggestion
```

dev/core/src/com/google/gwt/dev/jjs/impl/ToStringGenerationVisitor.java
niloc132: You're right... makes me suspicious that I "broke" this, at least in the context of `for` loops. I wish this had tests (since apparently it once upon a time produced compilable code, but that was many years ago), but I don't think it is critical to add them just for a debugging tool just yet...

I'll verify what our behavior is now here. My hunch is that it doesn't matter, and that we have just an assignment JBinaryExpression instead, but we'll see.

zbynek: I don't know what was the intention behind this flag and if it's still relevant (and if it is needed for `for` loops, is it also needed for try-with-resources?), but since this is just for debugging it should be fine to remove the flag.

niloc132: I don't think try-with-resources does the same thing? Each entry there needs a `;` between them and a type, right?

That is, this is a compile error
```java
    try (InputStream is1 = Files.newInputStream(null), is2 = Files.newInputStream(null)) {
      
    }
```
and should instead be ; delimited, with type names always present:
```java
    try (InputStream is1 = Files.newInputStream(null);InputStream is2 = Files.newInputStream(null)) {
      
    }
```
But for `for` loops you can use commas:
```
    for (int i = 0, j = 100; i < 10; i++) {
      
    }
```
GWT indeed represents the "initialization" statement as a list of statements, which is "not valid java", but it does the same thing for any other list of declarations. Those at least will be valid Java when they get here, but multi-variable for loops will not be.

niloc132:
Here's an example of the output when writing emulation unit testing - if you add ` -Dgwt.jjs.dumpAst=/tmp/ast.dump -Dgwt.jjs.dumpAst.filter=com/google/gwt/emultest/java/lang/StringTest.java:130-164` to the jvmargs, you will see this method in its various stages of compilation:
https://github.com/gwtproject/gwt/blob/a0bbaf01b9c92a7a91c7eec8128e003ea4f877f0/user/test/com/google/gwt/emultest/java/lang/StringTest.java#L130-L164

If you are doing draft tests (e.g. test.draft.htmlunit), you won't see as much, while prod tests (e.g. test.web.htmlunit) will show more steps to get to the final JS. This isn't a particularly interesting method to look at, so I'm not pasting the whole results ([see my gist for the whole thing](https://gist.github.com/niloc132/4bc97bc176bb901b2a2f673123831d50)), but one more interesting thing to call out is that by just jumping to the bottom, you can see the unobfuscated (but still optimized) JS method:
```javascript
_.testConcat = function testConcat(){
  var abc, c, def, empty, s;
  abc = valueOf(stampJavaTypeInfo(getClassLiteralForArray(C_classLit, 1), {8:1, 3:1}, 7, 15, [97, 98, 99]));
  def = valueOf(stampJavaTypeInfo(getClassLiteralForArray(C_classLit, 1), {8:1, 3:1}, 7, 15, [100, 101, 102]));
  empty = valueOf(stampJavaTypeInfo(getClassLiteralForArray(C_classLit, 1), {8:1, 3:1}, 7, 15, []));
  assertEquals('abcdef', abc + def);
  assertEquals('abcdef', (checkCriticalNotNull(abc) , abc + (checkCriticalNotNull(def) , def)));
  assertEquals('', (checkCriticalNotNull(empty) , empty + (checkCriticalNotNull(empty) , empty)));
  c = (checkCriticalStringElementIndex(0, def.length) , def.charCodeAt(0));
  s = abc;
  assertEquals('abcd', abc + 'd');
  assertEquals('abcd', abc + String.fromCharCode(c));
  assertEquals('abcd', s + 'd');
  assertEquals('abcd', s + String.fromCharCode(c));
  s += String.fromCharCode(c);
  assertEquals('abcd', s);
  try {
    ($wnd.Math.random() < -1 && fail() , null).$_nullMethod();
    fail();
  }
   catch ($e0) {
    $e0 = toJava($e0);
    if (!instanceOf($e0, 13))
      if (!instanceOf($e0, 15))
        throw toJs($e0);
  }
  try {
    $concat(($wnd.Math.random() < -1 && fail() , ''), ($wnd.Math.random() < -1 && fail() , null));
    fail();
  }
   catch ($e1) {
    $e1 = toJava($e1);
    if (!instanceOf($e1, 13))
      if (!instanceOf($e1, 15))
        throw toJs($e1);
  }
}
```

dev/core/src/com/google/gwt/dev/jjs/impl/AstDumper.java
```diff
@@ -80,25 +106,243 @@ public static void maybeDumpAST(JProgram jprogram, String versionString) {
...
+    final Set<FilterRange> files = getFilter();
+    try (PrintWriterTextOutput out = createWriter(dumpFile, fileExtension, append)) {
+      JVisitor v = new SourceGenerationVisitor(out);
+      if (files != null) {
+        v = new JFilteredAstVisitor(v, files);
+      }
+
+      v.accept(jprogram);
+    } catch (Exception e) {
+      System.out.println("Could not dump AST");
```
vegegoku, line 121: would it make sense to combine those into a single statement.

dev/core/src/com/google/gwt/dev/jjs/impl/AstDumper.java
```diff
@@ -80,25 +106,243 @@ public static void maybeDumpAST(JProgram jprogram, String versionString) {
...
+    }
+
+    @Override
+    public boolean visit(JMethod x, Context ctx) {
+      // Reset for each method
+      shouldVisitCurrentMethod = sourceFiles.stream().anyMatch(s -> s.matches(x));
+
+      // If we're already looking at this method, no need to keep checking
+      return !shouldVisitCurrentMethod;
+    }
```
vegegoku, line 265: can be written like this 

```java
@Override
public boolean visit(JMethod x, Context ctx) {
    return !(shouldVisitCurrentMethod =
            sourceFiles.stream().anyMatch(s -> s.matches(x)));
}
```
I dont understand the first comment, the 2nd one seems more usefull.

dev/core/src/com/google/gwt/dev/jjs/impl/AstDumper.java
niloc132: Sure - this is the old code, just copied here, but very reasonably should be better (its even in the TODO at the top)

dev/core/src/com/google/gwt/dev/jjs/impl/AstDumper.java
niloc132: I don't generally love inlining reading/writing into the same statement - are you okay if I leave it as two statements, and just do a better job with the comments?

zbynek (approved):

vegegoku (approved):

PR #10179, opened 2025-11-06
Use INSECURE_DELETE to support MacOS filesystems

niloc132:
This use case is only "insecure" insofar as any attacker who can already edit compiled output or temp/cache locations that the compiler is potentially now able to influence files that are deleted as well.

Fixes #10174


jnehlmeier (dismissed):
Looks good beside the checkstyle warnings. Some imports need to be moved up a line.

jnehlmeier (approved):

zbynek (approved):

PR #10180, opened 2025-11-06
Add emulation for Collection.toArray(IntFunction)

### LINKED ISSUE #10158: Implement Collection.toArray(IntFunction)
This is the only method missing in `Collection` class to get Java 17 compatibility.

niloc132:
Fixes #10158

### DIFF
--- a/user/test/com/google/gwt/emultest/java11/util/CollectionTest.java
+++ b/user/test/com/google/gwt/emultest/java11/util/CollectionTest.java
@@ -1,3 +1,18 @@
+/*
+ * Copyright 2025 GWT Project Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


user/test/com/google/gwt/emultest/java11/util/CollectionTest.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  public void testSetArrayOfMethodRef() {
+    String[] arr = Set.of("a", "b").toArray(String[]::new);
+    assertEquals(2, arr.length);
+    assertTrue(arr instanceof String[]);
+
+    // Order isn't guaranteed in a Set, so we compare against a known array, which will use the
+    // same traversal as long as nothing else has changed
+    Object[] objArr = Set.of("a", "b").toArray();
```
jnehlmeier: Maybe use `Set.of(a,b).containsAll(Arrays.asList(arr))`? From a test case point of view `toArray(IntFunction)` and `toArray()` could potentionally delegate to the same implementation method. If that methods returns [c, d] instead of [a, b] the test would be green.

niloc132: That seems reasonable, I'll make the change.

zbynek (approved):

jnehlmeier (approved):

PR #10182, opened 2025-11-07
Checkstyle: use SuppressWarnings instead of inline comments

zbynek:
SuppressWarnings makes it easier to see which warnings are actually suppressed and removes the need for on/off toggles that might get mismatched.

Moved out of #10176

### DIFF
--- a/eclipse/settings/code-style/gwt-checkstyle-tests.xml
+++ b/eclipse/settings/code-style/gwt-checkstyle-tests.xml
@@ -11,7 +11,7 @@ A more lenient set of style checks for test cases, needed because test often nee
 <module name="Checker">
     <property name="severity" value="warning"/>
     <module name="RegexpHeader">
-       <property name="fileExtensions" value="java"/> 
+       <property name="fileExtensions" value="java"/>
        <property name="severity" value="error"/>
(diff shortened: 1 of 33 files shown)


eclipse/settings/code-style/gwt-checkstyle-tests.xml
```diff
@@ -119,13 +120,6 @@ A more lenient set of style checks for test cases, needed because test often nee
             <property name="severity" value="ignore"/>
             <property name="classes" value="Boolean"/>
         </module>
-        <module name="Regexp">
```
zbynek, line 122: This rule is has a lot of false positives in general, in tests it can be just ignored since tests are not generating any user-facing error messages.

user/src/com/google/gwt/resources/rebind/context/InlineResourceContext.java
```diff
@@ -32,6 +32,7 @@ class InlineResourceContext extends StaticResourceContext {
   }
 
   @Override
+  @SuppressWarnings("checkstyle:SpaceAfterColon")
```
niloc132: For cases where we just have a single line that needs this, can we make the SuppressWarnings more specific - either just the deploy() method, or the `String encoded` local variable itself?

e.g.
```
@SuppressWarnings("checkstyle:SpaceAfterColon")
String encoded = "\"data:" + mimeType.replaceAll("\"", "\\\\\"")
    + ";base64," + base64Contents + "\"";
```

I appreciate that a lot of these cases tend to be a blanket "the whole file", but when we can make it more specific for non-test, it seems wise to do so (same as we also moved to do for errorprone suppressions).

zbynek: I've pushed some changes to restrict the suppression scope. For the `MethodName` I think keeping the class scope for classes that have more than one violating method is fine. If you think some scopes are still too broad, let me know.

niloc132 (dismissed):

niloc132 (approved):

PR #10185, opened 2025-11-13
Deprecate and disable CollectorLegacy, simplify startup

niloc132:
CollectorLegacy shouldn't be needed in any modern browser, though the Error.stack property is still non-standard. This patch leaves it in, so that it is at least possible for projects to restore it if they need it, but won't add an extra step to startup for most GWT applications.

With only a single implementation, the static final field isn't needed and likewise doesn't need to be eagerly initialized. This will remove clinits for both Impl and StackTraceCreator everywhere they are used, and by default should still compile out the collector instance into just a set of static functions.

### DIFF
--- a/user/src/com/google/gwt/core/client/impl/Impl.java
+++ b/user/src/com/google/gwt/core/client/impl/Impl.java
@@ -28,12 +28,6 @@ import com.google.gwt.core.client.JavaScriptObject;
  */
 public final class Impl {
 
-  static {
-    if (GWT.isScript() && StackTraceCreator.collector != null) {
-      // Just enforces loading of StackTraceCreator early on, nothing else to do here...
-    }
-  }
-
   private static final int WATCHDOG_ENTRY_DEPTH_CHECK_INTERVAL_MS = 2000;
 
   /**
--- a/user/src/com/google/gwt/core/client/impl/StackTraceCreator.java
+++ b/user/src/com/google/gwt/core/client/impl/StackTraceCreator.java
@@ -65,6 +65,7 @@ public class StackTraceCreator {
    * This legacy {@link Collector} simply crawls <code>arguments.callee.caller</code> for browsers
    * that doesn't support {@code Error.stack} property.
    */
+  @Deprecated
   static class CollectorLegacy extends Collector {
 
     @Override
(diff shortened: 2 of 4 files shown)


niloc132:
Combined compiled size of all examples after this change is `3320139` bytes, down about 18k (or 0.5%) from `3338616`.

vjay82:
OT: I saw that a watchdog timer is created on recursive entries (with a timeout?). I don't understand why it is necessary. Proper reset of the variable should be guaranteed through the final block, no? SchedulerImpl.runScheduledTasks(); seems to have its own exception handling. And even if that were not the case, would another try finally not be nicer than setting a timer, which just resets the variable but does not actually run scheduled commands? Am I missing something? I tried stripping that for my application (out of curiosity) and so far nothing bad has happened.

zbynek (approved):

niloc132:
@vjay82  Are you referring to `Impl.watchdogEntryDepthTimerId`? If so, the purpose there is to ensure that no changes could happen to Impl such that entry depth could ever be non-zero when leaving a stack with Impl at its base, that it always cleans up after itself.

Arguably that should be an assertion rather than always on. Also, it matters less now that JsInterop allows entry into gwt stacks without $entry at the base.

vjay82:
@niloc132 Looks like dead weight to me. Leaving it removed for now.
If bypassing $entry, scheduled tasks won't run, right? So there is a certain downside to it.

niloc132:
@vjay82 Scheduled tasks use `setTimeout`, so they will be run (though JSNI, and through $entry). What we lose in jsinterop is exit/entry hooks. Nearly no one ever uses `Scheduler.scheduleEntry` on purpose (I've only ever seen it on accident, except for some debugging I wrote once), so that's no great loss - but `scheduleFinally` is lost by this, it won't trigger in the current event loop, it won't trigger an immediate followup (like `scheduleDeferred`), but will go off at the end of the next event loop that _does_ go through `$entry`.

The solution in https://github.com/gwtproject/gwt-core is to use `Promise.resolve(null).then(..)` for this purpose - it technically doesn't give you "The end of this event loop", but it _does_ give you a microtask, which is "before the beginning of the _next_ event loop".

vjay82:
@niloc132 Yes, what I meant were these entry/exit tasks. I also have never seen any purpose of scheduleEntry and nobody has ever used in our company at least - maybe also stripping that, if not done by the compiler anyway in that case. Will have a look, thanks.

PR #10187, opened 2025-11-16
Faster implementation of long and int rotation

### LINKED ISSUE #9563: Long.rotateLeft is extremely slow
Long.rotateLeft is extremely slow. I use GWT version 2.8.2.

If I use generic rotate code : 
  `redacted, as taken directly from Oracle copyrighted code`
then it is more then 10x times faster. 
Which unsurprising looking at the implementation of rotateLeft.

The purposed fix is to use classic rotate implementation (see code ablove) rather then current immplemetation using loops.

Current workaround: do not invoke Long.rotateLeft and Long.rotateRight

zbynek:
Fixes #9563

I  picked up the implementation by @cogman from gerrit and added a few testcases.

### DIFF
--- a/user/super/com/google/gwt/emul/java/lang/Integer.java
+++ b/user/super/com/google/gwt/emul/java/lang/Integer.java
@@ -184,24 +184,17 @@ public final class Integer extends Number implements Comparable<Integer> {
   }
 
   public static int rotateLeft(int i, int distance) {
-    while (distance-- > 0) {
-      i = i << 1 | ((i < 0) ? 1 : 0);
-    }
-    return i;
+    int rotateValue = distance % SIZE;
+    int lowerBits = i >>> (SIZE - rotateValue);
+    int upperBits = i << rotateValue;
+    return upperBits | lowerBits;
   }
 
   public static int rotateRight(int i, int distance) {
-    int ui = i & MAX_VALUE; // avoid sign extension
-    int carry = (i < 0) ? 0x40000000 : 0; // MIN_VALUE rightshifted 1
-    while (distance-- > 0) {
-      int nextcarry = ui & 1;
-      ui = carry | (ui >> 1);
-      carry = (nextcarry == 0) ? 0 : 0x40000000;
-    }
-    if (carry != 0) {
-      ui = ui | MIN_VALUE;
-    }
-    return ui;
+    int rotateValue = distance % SIZE;
+    int upperBits = i << (SIZE - rotateValue);
+    int lowerBits = i >>> rotateValue;
+    return upperBits | lowerBits;
   }
 
   public static int signum(int i) {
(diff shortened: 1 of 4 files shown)


user/super/com/google/gwt/emul/java/lang/Integer.java
```diff
@@ -184,24 +184,17 @@ public static int reverseBytes(int i) {
   }
 
   public static int rotateLeft(int i, int distance) {
-    while (distance-- > 0) {
-      i = i << 1 | ((i < 0) ? 1 : 0);
-    }
-    return i;
+    int rotateValue = distance % SIZE;
```
tommyettinger: I feel like you should try shifting by any value outside the range this guarantees (which isn't 0 to 31, by the way, it's -31 to 31) in the browser's JS Console. You shouldn't just take my word for it, but the modulus is totally unnecessary here; shifts are implicitly done with a mask of 31 in JS, and in Java for ints.

I would not be at all surprised if performing modulus on JS Number types is slower than the loop before, and it's completely unnecessary. A mask would arguably be more clear, and wouldn't permit negative `rotateValue`s if that's a concern. A mask isn't needed either.

user/super/com/google/gwt/emul/java/lang/Long.java
```diff
@@ -128,24 +128,17 @@ public static long reverseBytes(long l) {
   }
 
   public static long rotateLeft(long i, int distance) {
-    while (distance-- > 0) {
-      i = i << 1 | ((i < 0) ? 1 : 0);
-    }
-    return i;
+    int rotateValue = distance % SIZE;
```
tommyettinger: Just use `distance & 63` here, if it's needed at all. GWT may already abide by the JLS and implicitly mask shifts on `long`s by 63, but I can't confirm it. This is internally a floating-point modulus because GWT `int` is JS `Number` (IEEE double-precision float that JS sometimes treats like a 32-bit integer), and floating-point modulus is the slowest basic arithmetic operation according to https://www.agner.org/optimize/ ; on Intel architectures, x87 FPREM takes reliably twice as long as FDIV, and there isn't a vector floating-point remainder operation that I can see. There doesn't appear to be an integer remainder operation there either, so it's probably produced by DIV or IDIV, which isn't much faster, unless a mask can be substituted in (which it probably can't, because if distance is negative, the result is different for `distance % 64` and `distance & 63`, even if the shifts for negative `rotateValue` behave the same).

tommyettinger (approved):
Looks good to me! All the expensive operations are gone and we have correct bitwise rotations, AKA barrel shifts. They were called barrel shifts long before Java existed. I wonder a little about what JS gets emitted here and what common JS engines compile it down to, but I'm sure the handful of loop-less, branch-less operations here will be much better.

niloc132 (approved):

Issue #10160, opened 2025-10-17
Explore method reference sharing, function deduplication


niloc132:
While looking at compiled output, I noticed that method references for something common like `String[]::new` ends up with a new class for each usage.

Might be good to fix along with #10086.

Presently, each method reference ends up looking (in PRETTY) like this:
```
cmp.MyClass$15methodref$lambda$5$Type = function MyClass$15methodref$lambda$5$Type(){
}
;
defineClass(1081, 1, {}, cmp.MyClass$15methodref$lambda$5$Type);
_.apply_2 = function apply_57(arg0){
  return initUnidimensionalArray(cggl.Ljava_lang_String_2_classLit, $intern_8, 2, arg0, 6, 1);
}
;
cggl.Lcom_my_project_MyClass$15methodref$lambda$5$Type_2_classLit = createForClass('com.my.project', 'MyClass/15methodref$lambda$5$Type', 1081);
```

In the project in question, there are 8 other classes with `apply_2` instance methods (with varying names) that have the exact same body, suggesting each was a different `String[]::new` reference. 

The compiler reasonably could rewrite all of those to be the same class, probably placed in String itself for consistency, savings of not less than 500 bytes (in OBF) just for String[]::new itself.

`initUnidimensionalArray` never optimizes out especially well, the body at best ends up as `{return a(b,c,2,d,6,1)}`. Even if we just pulled that out to its own static function (with `@DoNotInline`) and called from each place, we'd be a bit smaller, but a better answer is to always implement with the same class. Some care would need to be taken to match the same binding and interface. 

Implementing multiple interfaces for the same method reference in the same class might go to far (and may not actually ever help) - possibly worth looking at.

---

...Another, more specialized option could be to actually rewrite Stream.toArray(lambda) and Collection.toArray(lambda) to instead call `var list = stream.toList(); list.toArray(new MyClass[list.size()]);` or the like. This still has the large array creation code, but skips declaring the rest of the class (a much bigger savings, just from the `function` keywords alone). That would require a different kind of rewriting we don't support today, inlining away a lambda when directly invoked in the same block it was declared.

niloc132:
> Implementing multiple interfaces for the same method reference in the same class might go to far (and may not actually ever help) - possibly worth looking at.

[...]

PR #10188, opened 2025-11-16
Better error message for Collectors.toMap

### LINKED ISSUE #9495: Improve Lambda error messages
**GWT version: 2.8.0** 
**Browser (with version): Chrome 56.0.2924.87 (64-bit)** 
**Operating System: Windows 7** 

---

##### Description

Improve Lambda error messages. Please execute the example i've attached.
What i get running this as GWT, is:

Uncaught Error: java.lang.IllegalStateException: Can't assign multiple values to the same key
    at Fm.P [as s] (dmm-0.js:573:12483)
    at Fm.T [as v] (dmm-0.js:573:13033)

What i get running this in pure Java, is:

Exception in thread "main" java.lang.IllegalStateException: Duplicate key Item [id=1, name=1-Test]
	at java.util.stream.Collectors.lambda$throwingMerger$0(Collectors.java:133)
	at java.util.HashMap.merge(HashMap.java:1253)
	at java.util.stream.Collectors.lambda$toMap$58(Collectors.java:1320)

The code works fine, i only want highlight that the error message in Pure Java is way better than i GWT. It's much more harder to get behind the error in GWT.

##### Steps to reproduce

Run the attached Java Class as pure Java application and as GWT application.

[TestEntryPoint.txt](https://github.com/gwtproject/gwt/files/827338/TestEntryPoint.txt)

zbynek:
Fixes #9495 by changing the error message.

### DIFF
--- a/user/super/com/google/gwt/emul/java/util/stream/Collectors.java
+++ b/user/super/com/google/gwt/emul/java/util/stream/Collectors.java
@@ -380,7 +380,7 @@ public final class Collectors {
         keyMapper,
         valueMapper,
         (m1, m2) -> {
-          throw new IllegalStateException("Can't assign multiple values to the same key");
+          throw new IllegalStateException("Duplicate key " + m1);
         });
   }
 
--- a/user/super/com/google/gwt/emul/java/util/stream/Collectors.java
+++ b/user/super/com/google/gwt/emul/java/util/stream/Collectors.java
@@ -376,12 +376,31 @@ public final class Collectors {
   public static <T, K, U> Collector<T, ?, Map<K, U>> toMap(
       final Function<? super T, ? extends K> keyMapper,
       final Function<? super T, ? extends U> valueMapper) {
-    return toMap(
-        keyMapper,
-        valueMapper,
+    return Collector.of(
+        HashMap::new,
+        (map, item) -> {
+          K key = keyMapper.apply(item);
+          U newValue = Objects.requireNonNull(valueMapper.apply(item));
+          if (map.containsKey(key)) {
+            throw getDuplicateKeyException(key);
+          } else {
+            map.put(key, newValue);
+          }
+        },
         (m1, m2) -> {
-          throw new IllegalStateException("Duplicate key " + m1);
-        });
+          for (Map.Entry<K, U> entry : m2.entrySet()) {
+            if (m1.get(entry.getKey()) != null) {
+              throw getDuplicateKeyException(entry.getKey());
+            }
+            m1.put(entry.getKey(), entry.getValue());
+          }
+          return m1;
+        },
+        Collector.Characteristics.IDENTITY_FINISH);
+  }
+
+  private static RuntimeException getDuplicateKeyException(Object key) {
+    return new IllegalStateException("Duplicate key: " + key);
   }
 
   public static <T, K, U> Collector<T, ?, Map<K, U>> toMap(
@@ -423,7 +442,7 @@ public final class Collectors {
         mapSupplier,
         (map, item) -> {
           K key = keyMapper.apply(item);
-          U newValue = valueMapper.apply(item);
+          U newValue = Objects.requireNonNull(valueMapper.apply(item));
           if (map.containsKey(key)) {
             map.put(key, mergeFunction.apply(map.get(key), newValue));
           } else {
--- a/user/super/com/google/gwt/emul/java/util/stream/Collectors.java
+++ b/user/super/com/google/gwt/emul/java/util/stream/Collectors.java
@@ -376,31 +376,11 @@ public final class Collectors {
   public static <T, K, U> Collector<T, ?, Map<K, U>> toMap(
       final Function<? super T, ? extends K> keyMapper,
       final Function<? super T, ? extends U> valueMapper) {
-    return Collector.of(
-        HashMap::new,
-        (map, item) -> {
-          K key = keyMapper.apply(item);
-          U newValue = Objects.requireNonNull(valueMapper.apply(item));
-          if (map.containsKey(key)) {
-            throw getDuplicateKeyException(key);
-          } else {
-            map.put(key, newValue);
-          }
-        },
-        (m1, m2) -> {
-          for (Map.Entry<K, U> entry : m2.entrySet()) {
-            if (m1.get(entry.getKey()) != null) {
-              throw getDuplicateKeyException(entry.getKey());
-            }
-            m1.put(entry.getKey(), entry.getValue());
-          }
-          return m1;
-        },
-        Collector.Characteristics.IDENTITY_FINISH);
+    return toMap(keyMapper, valueMapper, null, HashMap::new, true);
   }
 
   private static RuntimeException getDuplicateKeyException(Object key) {
-    return new IllegalStateException("Duplicate key: " + key);
+    return new IllegalStateException("Duplicate key " + key);
   }
 
   public static <T, K, U> Collector<T, ?, Map<K, U>> toMap(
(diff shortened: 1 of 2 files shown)


user/super/com/google/gwt/emul/java/util/stream/Collectors.java
```diff
@@ -380,7 +380,7 @@ public static <T> Collector<T,?,List<T>> toList() {
         keyMapper,
         valueMapper,
         (m1, m2) -> {
-          throw new IllegalStateException("Can't assign multiple values to the same key");
+          throw new IllegalStateException("Duplicate key " + m1);
```
niloc132: I'm not sure this is what you want - `m1` and `m2` are the left/right _values_, not keys, right? The test passes because the map in the test uses the same string for key and value. I didn't test this, but it looks right based on the function below:
```
public static <T, K, U> Collector<T, ?, Map<K, U>> toMap(
      Function<? super T, ? extends K> keyMapper,
      Function<? super T, ? extends U> valueMapper,
      BinaryOperator<U> mergeFunction) {
    return toMap(keyMapper, valueMapper, mergeFunction, HashMap::new);
}
```
the BinaryOperator is generic on `U`, which is the value type, not `K`, the key type.

I think that to do this correctly, you have to call Collectors.of itself, like the toMap(keyMapper, valueMapper, mergeFunction, mapSupplier) overload does, so that you can replace the mergeAll with something that actually checks keys present on both sides.

user/super/com/google/gwt/emul/java/util/stream/Collectors.java
```diff
@@ -376,12 +376,31 @@ public static <T> Collector<T,?,List<T>> toList() {
   public static <T, K, U> Collector<T, ?, Map<K, U>> toMap(
       final Function<? super T, ? extends K> keyMapper,
       final Function<? super T, ? extends U> valueMapper) {
-    return toMap(
-        keyMapper,
-        valueMapper,
+    return Collector.of(
```
niloc132: Lambdas and method references are pretty expensive to generate, and while the new error message will be shorter, the total generated code is probably going to increase for this change.

Is there another approach we can use to avoid this, composing the toMap() call with some other collision check? At a glance, I'd either say composing something into the other lambdas we have to produce (likely parameterizing them, which isn't all that great), or maybe a call like mapping(collisionCheckHere, toMap(...))?

user/super/com/google/gwt/emul/java/util/stream/Collectors.java
```diff
@@ -376,12 +376,31 @@ public static <T> Collector<T,?,List<T>> toList() {
...
+        HashMap::new,
+        (map, item) -> {
+          K key = keyMapper.apply(item);
+          U newValue = Objects.requireNonNull(valueMapper.apply(item));
```
niloc132: The null check is surprising - it isn't specified at https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Collectors.html#toMap(java.util.function.Function,java.util.function.Function), but it does seem to be part of the current JDK at least:

```
jshell> Stream.of(1).collect(Collectors.toMap(ignore -> "", ignore -> null))
|  Exception java.lang.NullPointerException
|        at Objects.requireNonNull (Objects.java:209)
|        at Collectors.lambda$uniqKeysMapAccumulator$1 (Collectors.java:180)
|        at ReduceOps$3ReducingSink.accept (ReduceOps.java:169)
|        at Streams$StreamBuilderImpl.forEachRemaining (Streams.java:411)
|        at AbstractPipeline.copyInto (AbstractPipeline.java:509)
|        at AbstractPipeline.wrapAndCopyInto (AbstractPipeline.java:499)
|        at ReduceOps$ReduceOp.evaluateSequential (ReduceOps.java:921)
|        at AbstractPipeline.evaluate (AbstractPipeline.java:234)
|        at ReferencePipeline.collect (ReferencePipeline.java:682)
|        at (#9:1)
```

You could consider composing the existing function with `disallowNulls(...)`.

user/super/com/google/gwt/emul/java/util/stream/Collectors.java
zbynek: Of course, you're right 🤦

user/super/com/google/gwt/emul/java/util/stream/Collectors.java
zbynek: >  composing the existing function with disallowNulls

Wouldn't that instantiate more lambdas?

user/super/com/google/gwt/emul/java/util/stream/Collectors.java
zbynek: The only way I was able to reduce the number of lambdas was parametrizing the private `toMap` 🤔
The extra parameter could be avoided if we say that `mergeFunction == null` should be interpreted as `unique == true` in the private method (and possibly add a null check to the public one).

I don't think this can be easily solved with composition since either `collisionCheck` would have to return both `T` and `K` wrapped in an `Entry` or `keyMapper` would have to be called twice (in collision checking and in accumulator).

user/super/com/google/gwt/emul/java/util/stream/Collectors.java
niloc132: Instantiating lambdas isn't awful, its declaring them - any time you write `->` or `::`, you can imagine you just made a new class instead. Not just an anonymous inner class, but build the rest of the machinery in your head for there to be a constructor to pass in the outer variables/instances (fields, etc), potentially the getClass, the type info to know the supertype/interfaces of the lambda, etc.

I wouldn't bat an eye at someone adding lambdas in application code, but core JRE emulation code for something as commonly used as toMap seems worth working a little harder on - even if it means the error message might be a little subpar. Library code I'd want to work slightly harder where its easy, but not worry about it (e.g. prefer `if (optional.isPresent()) {...}` over `optional.map(...)` sorts of things).

Instantiating them is only as bad as instantiating anything else, `new Constructor(args, to, close, over)`, since at the end of the day (non-JsFunction) lambdas are just implementations of SAM interfaces. We could probably do better (...the JVM does), but it both adds complexity to the compiler, and could risk making something like incremental builds stop working. Instantiating them also only happens once, at the declaration site - but if you pull that out into a util method (like disallowNulls()), you just pay all that once, and then get to do `a(b)` to wrap the existing lambda in another lambda each time it is needed.

JsFunction lambdas are a little different in how they are built, but are actually slightly cheaper to call in GWT-generated Java than non-JsFunction lambdas - rather than `instance.apply(args)`/`a.b(c)` when you invoke, its just `instance(args)`/`a(c)`, savings of 2ish bytes per callsite.

niloc132 (approved):
This is good, thanks for taking the extra steps here. I think we can get better, but I don't think its worth doing now - and the fixes touch a lot more of the class (and besides, they are going back to the first impl that we did for this class).

I did some compiled size testing. My methodology:
 * set `test.args` in `user/build.xml` to `-ea -sourceLevel auto -setProperty compiler.useSourceMaps=true` so that we avoid using emulated stack traces, and get more realistic JS output
 * run all CollectorsTest impls by making a single CollectorsSuite
    ```java
    package com.google.gwt.emultest.java17;

    import org.junit.runner.RunWith;
    import org.junit.runners.Suite;

    @RunWith(Suite.class)
    @Suite.SuiteClasses({
      com.google.gwt.emultest.java8.util.stream.CollectorsTest.class,
      com.google.gwt.emultest.java9.util.stream.CollectorsTest.class,
      com.google.gwt.emultest.java10.util.stream.CollectorsTest.class,
      com.google.gwt.emultest.java17.util.stream.CollectorsTest.class,
    })
    public class CollectorsSuite {
    }
    ````
 * Run just that suite in prod mode in the `user/` dir: `ant test.web.htmlunit '-Dgwt.junit.testcase.includes=**/CollectorsSuite.class'`
 * Check the compiled size, with something like `ls -al ../build/out/user/test/web-htmlunit/www/com.google.gwt.emultest.EmulSuite.JUnit/*.cache.js`

HEAD, plus your test change (with two failing tests), currently d67edbb8cd: 153033
latest commit in this PR, currently  07617ec0a0, merged up to latest main: 153339

306 bytes over 150kb isn't bad, and probably not entirely driven by lambdas (additional null checks, etc). Percentage wise, its not great - but this will not scale, it'll be the max size increase that this change can cause over a full app, since its not like an app would load multiple copies of `Collectors`. Generally this is all keeping with the work done this release to improve emulation, so I'd vote worth it, and we can focus more on improved compiled size next release.

niloc132:
[Quick draft](https://github.com/niloc132/gwt/tree/refs/heads/collectors-lambda-consolidation-test) - harms readability IMHO, but improves size, down to ~~151919~~151762 using the tests above. Unsure about performance, we should add some testing here as well, but I think it is a mild improvement in at least one case (moving an `if` outside of a lambda so it is only checked once). 
Two-ish techniques taking place here:
 * First, any duplication of method references gets extracted to a helper method - those will all be rewritten to just constructor calls and inlined, so there's no extra cost to them. Might become largely unnecessary if we find a nice way to resolve https://github.com/gwtproject/gwt/issues/10160.
 * Second, when a Collector.of(...) call would take multiple lambdas/method references, make a single class that implements all of them at once. I only did this for a single case of four method references, and it only saved about 60 bytes (while creating a lot more input Java). Maybe thats not much, but if we're confident in the tests and we apply it everywhere, that's probably 2-300 bytes saved overall any application that uses most of the collectors... Still probably only barely worth it.

PR #10189, opened 2025-11-16
Fix group references in String::matches

### LINKED ISSUE #9572: Emulated String.matches(String) method introduces capturing group that breaks semantics.
**GWT version: any
**Browser (with version): any
**Operating System: any

---

##### Description

The emulated String class super/com/google/gwt/emul/java/lang/String.java defines matches(String) as
```
public boolean matches(String regex) {
    // We surround the regex with '^' and '$' because it must match the entire string.
    return new NativeRegExp("^(" + regex + ")$").test(this);
}
```
This introduces a capturing group to the regex, which changes the group numbering for any back references in the regex string.

This makes the emulated gwt regex incompatible with the same pattern executed in Java.

For example, to match a string containing repeated characters of 3 or more:

`"12333456".matches(".*(.)\\1{2,}.*")`

returns true in java and false in compiled javascript, whereas

`"12333456".matches(".*(.)\\2{2,}.*")`

returns true in compiled javascript and false in java.

If the emulated String.matches(String) method used a non-capturing group:

`    return new NativeRegExp("^(?:" + regex + ")$").test(this);`

I expect this would solve the issue.

##### Known workarounds

Use two pattern strings by adding 1 to every numbered back reference for the pattern used by compiled javascript.

(javascript regex does not support named back references)

zbynek:
Fixes #9572 by applying the change proposed by @morganga.

### DIFF
--- a/user/super/com/google/gwt/emul/java/lang/String.java
+++ b/user/super/com/google/gwt/emul/java/lang/String.java
@@ -520,7 +520,7 @@ public final class String implements Comparable<String>, CharSequence,
    */
   public boolean matches(String regex) {
     // We surround the regex with '^' and '$' because it must match the entire string.
-    return new NativeRegExp("^(" + regex + ")$").test(this);
+    return new NativeRegExp("^(?:" + regex + ")$").test(this);
   }
 
   public int offsetByCodePoints(int index, int codePointOffset) {
--- a/user/super/com/google/gwt/emul/java/lang/String.java
+++ b/user/super/com/google/gwt/emul/java/lang/String.java
@@ -518,6 +518,7 @@ public final class String implements Comparable<String>, CharSequence,
    *
    * TODO(jat): properly handle Java regex syntax
    */
+  @SuppressWarnings("checkstyle:SpaceAfterColon")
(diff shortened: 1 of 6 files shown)


user/super/com/google/gwt/emul/java/lang/String.java
```diff
@@ -520,7 +520,7 @@ public int length() {
    */
   public boolean matches(String regex) {
     // We surround the regex with '^' and '$' because it must match the entire string.
-    return new NativeRegExp("^(" + regex + ")$").test(this);
+    return new NativeRegExp("^(?:" + regex + ")$").test(this);
```
niloc132, line 524: Add suppression for this warning?

zbynek:
CI was failing builds because network errors and manual re-run failed, so I pushed some unrelated typo fixes 🙈

niloc132:
Yeah github was down https://www.githubstatus.com/incidents/5q7nmlxz30sk

I tried rerunning the build when it seemed my own access was working again, but that wasn't enough for GHA to build.

niloc132 (approved):

PR #10192, opened 2025-11-17
Emulate unsigned numbers Java 8 API.

zbynek:
Ports `unsigned` additions from gerrit -- draft.

Fixes #9643


user/super/com/google/gwt/emul/java/lang/Long.java
```diff
@@ -59,11 +59,44 @@ public static int compare(long x, long y) {
     }
   }
 
+  public static int compareUnsigned(long a, long b) {
```
niloc132, line 62: (as it happens, I was looking at this today)

PR #10202, opened 2025-12-03
Security related update of jetty and jetty related libraries

Lonzak:
This pull requests updates the jetty libraries due to a security issue. I know that gwt-dev is not for production code however explain this to the security guys ;-)

There is a [linked PR](https://github.com/gwtproject/tools/pull/38) in the gwt-tools project as well.

**Please note,** that the checks are failing since the libs are missing from the gwt-tools project. I think there is no way to link both project's here.


Lonzak:
Sorry there was a mix up with previous PR'ed commits => thus closed this one here.

PR #10204, opened 2025-12-04
Fix sporadic error with detection of instantiable subtypes

### LINKED ISSUE #10181: Sporadic [ERROR] '{type-name}' has no instantiable subtypes
We see sporadic GWT compiler errors reported during the "Computing all possible rebind results for '{some-GWT-RPC-service}'" phase. The problem occurs in roughly 10% of our builds. I'm attaching the log excerpt from when the issue occurred (bug.log) and when everything went fine (no-bug.log).

- [bug.log](https://github.com/user-attachments/files/23418210/bug.log)
- [no-bug.log](https://github.com/user-attachments/files/23418209/no-bug.log)

To see the full build log, check out https://github.com/SAP/sailing-analytics/actions/runs/19071837371 with its build log at https://github.com/SAP/sailing-analytics/actions/runs/19071837371/artifacts/4463965667, and https://github.com/SAP/sailing-analytics/actions/runs/19140961071 with its build log at https://github.com/SAP/sailing-analytics/actions/runs/19140961071/artifacts/4491062919, respectively.

As far as I was able to read from the compiler sources, the problem may be caused by the order in which all subtypes of Serializable or java.lang.Object are enumerated, differing between builds because they are provided by hashed collections (HashSet, HashMap). Ultimately, the error message is produced in SerializableTypeOracleBuilder.build(TreeLogger) calling the computeTypeInstantiability(...) method without success.

In the "good" case, the "Verifying instantiability" for java.io.Serializable starts descending into java.util.ArrayList<?>, whereas in the "bad" case the descending starts with a different type, com.sap.sse.security.shared.ServerAdminRole.

In our case, if a run fails for this problem, it is always the com.sap.sse.datamining.shared.impl.dto.StoredDataMiningReportDTOImpl type that is said to not have any instantiable subtypes. See also here: https://github.com/SAP/sailing-analytics/blob/main/java/com.sap.sse.datamining.shared/src/com/sap/sse/datamining/shared/impl/dto/StoredDataMiningReportDTOImpl.java

With only a 10% probability of this problem occurring, I have no good idea how I would work towards reproducibility of this issue...

axeluhl:
See the [issue ](https://github.com/gwtproject/gwt/issues/10181) for discussions.

This PR contains four tests based on two example type graphs that demonstrate the issue and consistently fail when executed with the ``main`` version. They all fail consistently because a class that is the only instantiable subtype of an interface has a cycle across its fields' types leading back to the class or the interface it implements.

During checking for instantiable subtypes, the type in question is marked as pending in its ``TypeInfoComputed`` (TIC). Then, when the decision isn't final after the ``checkSubtypes`` invocation, the TIC is not yet marked as done, and erroneously the TIC is then set to ``instantiable==false`` which is incorrect at that time.

The fix leaves a pending subtype pending unless a positive result was found and only updates its TIC regarding any instantiable subtypes. This way, inconsistent decisions about types not being instantiable by the RPC generator can be avoided. Four test cases that fail prior to this fix pass with the fix.

Fixes https://github.com/gwtproject/gwt/issues/10181


user/test/com/google/gwt/user/rebind/rpc/testcases/client/SimplifiedRecursiveTypeGraphInstantiability.java
```diff
@@ -0,0 +1,43 @@
...
+  /**
+   * Default serializable, but with back-reference to B for which the question of instantiable
+   * subtypes depends on the serializability of this class. The reference to {@link C} which has a
+   * back-reference to {@link A} and B being its only subtypes candidate helps reproduce issue
+   * 10181.
+   */
+  class B implements A {
+    B b;
+    C c;
+  }
```
niloc132, line 36: Tests should permutate the order of these  (fields here and in the non-simple vers, the interfaces implemented, the root type requested, etc) - if we later introduce/discover a case where we fail only for a particular encounter order, we'll have to fix it again. 

Probably best to inline these classes as strings in the test, make java resources from them as other tests do, and repeat the test with different orderings as needed.

Likewise, enforcing stability in the traversal order of types (fields already have source order enforced if i'm not mistaken) will prevent the inconsistent nature of the bug, as discussed in the ticket.

user/test/com/google/gwt/user/rebind/rpc/testcases/client/SimplifiedRecursiveTypeGraphInstantiability.java
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright 2025 Google Inc.
```
niloc132: GWT Project Authors, here, above (if you keep the files)

user/test/com/google/gwt/user/rebind/rpc/SerializableTypeOracleBuilderTest.java
```diff
@@ -1521,6 +1523,78 @@ public void testNotAllSubtypesAreSerializable() throws UnableToCompleteException
...
+        new TypeInfo(makeSourceName(RecursiveTypeGraphInstantiability.B.class.getName()), true),
+        new TypeInfo(makeSourceName(RecursiveTypeGraphInstantiability.C.class.getName()), true)};
+    validateSTO(sto, expected);
+  }
+
+  public void testInstantiabilityDetectionOnRecursiveTypeGraphStartingAtB()
+      throws UnableToCompleteException, NotFoundException {
+    TreeLogger logger = createLogger();
+    TypeOracle typeOracle = getTestTypeOracle();
+    JClassType b = typeOracle.getType(RecursiveTypeGraphInstantiability.B.class.getCanonicalName());
```
niloc132, line 1548: as in the other comment, this test is largely the same except as above for swapping A for B? Maybe make a helper method/etc so you can test with different starting points that they achieve the same result?

Starting with C should also achieve the same two types.

user/src/com/google/gwt/user/rebind/rpc/SerializableTypeOracleBuilder.java
```diff
@@ -1003,7 +1003,9 @@ TypeInfoComputed computeTypeInstantiability(TreeLogger logger, JType type, TypeP
     Set<JClassType> instantiableTypes = new HashSet<JClassType>();
     boolean anySubtypes =
         checkSubtypes(localLogger, originalType, instantiableTypes, path, problems);
-    if (!tic.isDone()) {
+    if (tic.isPendingInstantiable()) {
+      tic.setInstantiableSubtypes(anySubtypes);
+    } else if (!tic.isDone()) {
```
niloc132, line 1008: What made you end up ruling out the earlier check, are there cases where we actually need to descend a second (or more) time into checkSubtypes above, rather than exit early and not mutate the TIC further?

niloc132 (commented):
Also, could you update the title to be about the patch (though still brief), and description to follow the template - including details of "why" the change was made? A trailer at the end can then link to the bug ID, but doesn't require a future reader to go over 50+ comments to understand a standalone patch.

zbynek:
@axeluhl the phrase `Fixes #10181` needs to be part of a comment for GitHub to link issues to PRs properly. Ideally it should be the last part of the PR's initial comment (a.k.a. PR description). I've updated the titles and initial comments of both your PRs, but for this one it might be useful to make the initial comment more detailed, please edit it to include more information as Colin suggests.

user/test/com/google/gwt/user/rebind/rpc/testcases/client/SimplifiedRecursiveTypeGraphInstantiability.java
axeluhl: Done (a0ca1da0afe4dd91610afa7f48e559f1eab25084)

user/src/com/google/gwt/user/rebind/rpc/SerializableTypeOracleBuilder.java
axeluhl: Which "earlier check" are you referring to? In the ``testInstantiabilityDetectionOnRecursiveTypeGraphStartingAtA`` case, for example, ``B`` is visited to check for instantiable subtypes of ``A``. It recurses and finds itself as the only instantiable subtype of ``A`` while its TIC is still in pending state. Later, after all fields, including the field of type ``C`` which again recurses to check ``B``, have succeeded, ``B``'s TIC will finally be marked as ``instantiable==true`` which will be helpful for all later checks (none in this test case, but potentially some or even many depending on the actual case) because then we no longer need to assume that a pending TIC could suggest instantiability.

user/test/com/google/gwt/user/rebind/rpc/testcases/client/SimplifiedRecursiveTypeGraphInstantiability.java
axeluhl: There is certainly more that can be done to improve the test base. I've added two cases that approach slightly different problematic type graphs from different ends and think this helps more than only a single case, also to make undetected future regressions less likely.

Candidate traversal order stability IMHO has pros and cons. Pro: it fails once, it fails always. Con: it conincidentally does not fail, and another problem in the algorithm stays looming for a potentially longer time. But I guess this is not on me to decide.

user/test/com/google/gwt/user/rebind/rpc/SerializableTypeOracleBuilderTest.java
axeluhl: The test case with ``D`` as a second interface for me was a useful way to force a second subtype check on ``B`` as both interfaces' only implementation class. That's why I personally still find it useful.

user/src/com/google/gwt/user/rebind/rpc/SerializableTypeOracleBuilder.java
niloc132: See the second proposed patch in https://github.com/gwtproject/gwt/issues/10181#issuecomment-3604549826 - you're basing yours off the first one (which came from the earlier issue from the user `dhulme`, then was remade, incorrectly, into a gerrit review, but ultimately not merged). The second proposed patch avoids recursing into the sutypes at all, but still appears to yield the correct results.

This appears to be a superior change from my perspective - we don't descend as much into already-in-progress types, and we don't have to half-assign the TIC's knowledge of its types. It appears to actually be correct as well - the `TIC.instantiableSubtypes` field is only read via `hasInstantiableSubtypes` anyway (see the discussion in that comment), which will be true _anyway_, since the TIC is marked as in progress already.

Thus, the bug is not that it was never "marked as `instantiable==true`" (this never is set until it is marked DONE anyway), but that it was visited again while it was already in progress, and the nested visit had incomplete information. Returning early, before we get the chance to visit again, means that we not only already have the right answer, but we skip doing useless work.

I'm only about 80-90% that this is correct, I haven't had time to add more asserts/tripwires/etc in the rest of the STOB to look for other possible side effects from the extra traversal. Given how widely used GWT-RPC is in large existing apps with no plans to rewrite away from it, I would hate to change something subtly and either degrade build performance or make some _other_ use case generate incorrect code.

See also the end of that comment - it has my suggested list of important things to change to ensure that this class of bug can't happen again.

user/test/com/google/gwt/user/rebind/rpc/testcases/client/SimplifiedRecursiveTypeGraphInstantiability.java
niloc132: > Con: it conincidentally does not fail, and another problem in the algorithm stays looming for a potentially longer time. But I guess this is not on me to decide.

I think that's fair, but I don't think stochastic testing in downstream applications is the way to find that - or at least, it should be opt-in. I'd be comfortable with the sorting being optional, so it is consistent by default, but if someone wants to fuzz test (or we want in-build fuzz tests) we can support that.

user/test/com/google/gwt/user/rebind/rpc/SerializableTypeOracleBuilderTest.java
niloc132: No objection to keeping the more complex test as well.

user/src/com/google/gwt/user/rebind/rpc/SerializableTypeOracleBuilder.java
axeluhl: > See the second proposed patch in https://github.com/gwtproject/gwt/issues/10181#issuecomment-3604549826 - you're basing yours off the first one (which came from the earlier issue from the user dhulme, then was remade, incorrectly, into a gerrit review, but ultimately not merged). The second proposed patch avoids recursing into the sutypes at all, but still appears to yield the correct results.

I think it's useful to carry out the check for a "pending" TIC because finally it will have its ``instantiable`` field set correctly. This will then set the state to "done" which from there on will stop the recursion and use the definite result. You know we have a complex case, and with the current (single) patch I cannot see a significant increase in build times. We merged our fork's release with this patch included, and the builds can be seen starting with https://github.com/SAP/sailing-analytics/actions/runs/19966693686.

Feel free to add more tests, of course. I think the ones added on this PR check for the things that I think I understood why the bug was reproduced by them and how this PR fixes them. The discussion about whether or not random subclass candidate lists are a good idea we had on this other comment.

PR #10206, opened 2025-12-05
Fix AbstractLocalizableInterfaceCreator.genMethodDecl; Fixes #10205

axeluhl:
Fixes the order of AbstractLocalizableInterfaceCreator.genMethodDecl arguments; 
this reverses the arguments of genMethodDecl to their original order so that the single call matches the declaration again.

Fixes #10205

### DIFF
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -20,6 +20,10 @@ permissions:
   checks: write 
 jobs:
   build:
+    permissions:
+      contents: write
+      checks: write
+      actions: read
     if: ${{ !github.event.pull_request || contains(github.event.pull_request.labels.*.name, 'ready') }}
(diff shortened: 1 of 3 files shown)


niloc132:
This is definitely the right fix for this specific bug, but it also calls back to my [request in the initial PR](https://github.com/gwtproject/gwt/pull/10166#pullrequestreview-3366265243) that we need more tests here. I don't personally use these Creator tools, but if I were to put myself in the place of someone who did, I think I'd want to see more tests to ensure this or other errors can't happen.

At this time I'm inclined to say we should just revert the patch and re-land it when it is tested, rather than risk breakage in 2.13 for users - let it bake a little more first to be sure. 

Could you update your title/description of the PR to match what the template suggested - short concise title, descriptive commit message, ends with the "Fixes..." trailer?

PR #10208, opened 2025-12-07
Status page for JRE emulation

zbynek:
Alongside of the list of methods and fields that are supported, generate a second file listing methods and fields that are missing compared to currently running JVM.

Result of running this against `main`: https://codepen.io/zbynek/pen/NPWBzKP

Since copying of this snippet to gwt-site repo is manual, whoever's doing it can make sure it runs in Java 17.


user/super/com/google/gwt/emul/java/security/SHA256Digest.java
```diff
@@ -20,7 +20,7 @@
 /**
  * SHA256 implementation for GWT.
  */
-public class SHA256Digest extends MessageDigest {
+class SHA256Digest extends MessageDigest {
```
zbynek, line 23: This class should not be mentioned in Javadoc, it doesn't exist in JRE.

niloc132 (commented):
This is excellent, thank you for working on this! I love the idea, and the output looks great, though I'll admit I haven't looked much into the implementation yet.

I don't really love the properties format - but a lot of this is first glance, and I havent taken the time to think it through as much as you have, so take my initial feedback with a grain of salt:

The properties format seems tricky to edit - the signature needs to be just right, and newlines escaped between items... I'm guessing you have a tool that generates various values, and then you grouped them into the sets you have here? Could we go the other way, and for all missing methods, emit a properties file with an escaped key per missing member, equals sign, and a dummy group to be reassigned manually? Editing that file directly (or appending each updated entry to the existing checked in file), and not worrying about grouping order, etc.

I also think we might want to consider dealing with upcoming emul that this detects in the same sorts of ways we deal with existing emul - an issue per type/member. With the (escaped) signature as the key, one could imagine a <key>.issue=1234 pattern to link to issues as they are created, possibly making up batches of issues as we update Java versions, to be picked off by contributors as they are completed?

Some other notes on the content:
 * Varargs: as was pointed out in slack, I seem to be wrong about our need for this at all - maybe some group types can just be outright hidden, since this doesn't appear to impact developers at all?
 * nio/time: All these members makes me wish we had a `@GwtIncompatible` except for "only compile this member if the following types are available at compile time..." , so that different implementations can coexist, without one having to be checked in to the main repo to be usable at all (as is the case for j2cl for most jre emul...). Its not that simple (what if you include the member since java.time.LocalDate was present, but then Date.valueOf(LocalDate) ends up caling a LocalDate that wasn't present in that emulation), but it might be worth thinking about more.
 * logging: It might be reasonable to add more of these features - we already have decent JULI support, and I know Domino has a SLF4J impl (which has some real advantages with varargs to optionally interpolate only if the log level is set to include that message...)
 * properties - I'm not sure what different semantics you're referring to? It would be more work to do it right for sure (each method would need special handling by the compiler), and the property name would still always need to be a compile-time constant (not a variable passed from another method, etc), but it could work.
 * long: I don't _think_ we must have improved long support for most of these, and I'm not sure that bigint support actually any/all of this faster anyway. 
 * concurrent: We already have some "concurrent" apis, where users have found it to be helpful. I agree that we can't actually do the "real" concurrent implementation in any way that matters, but we have ConcurrentHashMap and Collection.parallelStream() already, they just do the non-concurrent implementation. Things like Collectors.groupingByConcurrent should be easy to implement... but the usage in the wild seems to be so low that it hasn't mattered, but I don't think that means we should insist on not doing it. Maybe this category is more of "we could do it, but someone will have to ask/contribute" indicating it is lower priority than its easy of implementation would suggest?
 *  java.nio.charset.StandardCharsets - none of these actually need nio classes, since we already support `ISO_8859_1` and `UTF_8`. I actually have an (extremely mild) need of US_ASCII, but haven't had time to just do it...
 * Double.doubleToRawLongBits (and Float too) - the JS spec actually says we can't do this correctly, but chrome/others disregard the spec here, leaving firefox as unable to support this (at least with typedarrays)
 * ResourceBundle usage is likely a problem, but the methods that use this are in Logging, Access to system resources, while I think it _probably_ is more like Localization? Not 100% sure on that.

zbynek:
With the different semantic of properties I meant that in Java they are essentially global variables whereas in GWT they are compile time constants similar to preprocessor macros inother languages. `System.setProperty(String, String)` won't work in GWT (same for `setProperties` and `clearProperty`), which reduces the usefullness of the other properties-related methods.

user/src/com/google/gwt/logging/client/DefaultLevel.java
```diff
@@ -24,7 +24,7 @@
  * gwt.xml file.
  */
 public interface DefaultLevel {
-  /**
+  /**git
```
niloc132:
```suggestion
  /**
```

(glad to see i'm not the only one who does this... though I do it at least once a week...)

zbynek: thanks! 🙈

zbynek:
> With the (escaped) signature as the key, one could imagine a .issue=1234 pattern to link to issues as they are created, possibly making up batches of issues as we update Java versions, to be picked off by contributors as they are completed? 

New approach: one .properties file per GitHub issue, each issue has a batch of methods, they can be broken down into subtasks if they turn out to be too big (but e.g. https://github.com/gwtproject/gwt/issues/10215 is just 3 closely related methods in many variants). The code generates `triage.properties` so that you can easily move the signatures into their own `gh*.properties` file.

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -16,281 +16,484 @@
...
+        } catch (IOException ex) {
+          throw new UncheckedIOException(ex);
+        }
+      });
+    } catch (IOException ex) {
+      throw new UncheckedIOException(ex);
+    }
+    File outFilePresent = new File(outputFile);
+    File outFileMissing = new File(outputFile.replace(".html", "-missing.html"));
+    outFilePresent.getParentFile().mkdirs();
```
niloc132: Files+Path instead here? `missingDir` is already a Path. Better errors for mkdirs() than just ignoring it if we can't write.

I know you didn't write this all from scratch here, just seems better to be consistent since it needs this much of a rewrite for this.

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -16,281 +16,484 @@
...
+  @Override
+  public boolean run(DocletEnvironment env) {
+    final Properties properties = new Properties();
+    Path missingDir = Path.of("build_tools/doctool/src/com/google/doctool/custom/missing");
+    try (Stream<Path> str = Files.list(missingDir)) {
+      str.filter(file -> !file.getFileName().toString().startsWith("triage")).forEach(file -> {
         try {
+          properties.clear();
+          properties.load(Files.newInputStream(file));
+          String issue = file.getFileName().toString().replaceAll("gh|\\.properties", "");
```
niloc132: I like the `gh` prefix for future proofing, but don't love that we have to hard code the templating or pathing. I don't think this is really actionable today though.

As a very tentative thought (not required to merge this), what would you think about a `link` key in the prop file rather than assuming the url structure...? Would make it easier for GWT forks to add their own issues as desired too.

Especially since we have to paste the title manually into the properties file too.

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -16,281 +16,484 @@
...
+          ghToSignatures.put(issue, (String) properties.get("members"));
+          ghToTitle.put(issue, (String) properties.get("title"));
+        } catch (IOException ex) {
+          throw new UncheckedIOException(ex);
+        }
+      });
+    } catch (IOException ex) {
+      throw new UncheckedIOException(ex);
+    }
```
niloc132: Can probably stretch this try/catch down to the end of the method

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -16,281 +16,501 @@
...
+
+  private final Map<String, String> issueToSignatures = new HashMap<>();
+  private final Map<String, String> issueToTitle = new HashMap<>();
+  private final Map<String, String> issueStatus = new HashMap<>();
+  private final List<String> triage = new ArrayList<>();
+
+  @Override
+  public boolean run(DocletEnvironment env) {
+    final Properties properties = new Properties();
+    Path missingDir = Path.of("build_tools/doctool/src/com/google/doctool/custom/missing");
```
niloc132: This tool is internal use only, but it would be nice if we instead had an option for "-out", another for "-missing", a third for "find the existing properties files here", and then finally "put any triage properties here".

It'll need to be fixed before we can rewrite the build, since these internal hardcoded paths will change

build_tools/doctool/src/com/google/doctool/custom/missing/triage.properties
```diff
@@ -0,0 +1,13 @@
+members=java.lang.String#isEmpty()\
```
niloc132: Should we be checking in triage, or just use its existence as a warning that someone needs to fix the build? I'd say we might even want to fail the build if one is created - if someone changes the min jdk or adds a new package that should be supported, they should run the build manually at least once (even in CI) and see that there is new output to deal with?

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -16,281 +16,501 @@
...
+      Set<String> allClasses = getSpecifiedPackages(env)
+        .flatMap(pack -> pack.getEnclosedElements().stream()
+        .flatMap(clazz -> withInnerClasses(clazz, pack)))
+        .collect(Collectors.toSet());
+      getSpecifiedPackages(env).forEach(pack -> {
+        Optional<Module> matchingModuleName = ModuleLayer.boot().modules().stream()
+            .filter(m -> m.getPackages().contains(pack.getQualifiedName().toString()))
+            .findFirst();
+
+        pwPresent.format("<h2 id=\"Package_%s\">Package %s</h2>\n",
```
niloc132: I really wish we could use SafeHtml here so contributors/reviewers don't have to look quite so closely at each possible string and where it came from...

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -16,281 +16,501 @@
...
+      }
+    }
+  }
+  private Object getIssueTitle(String key) {
+    if (key.isEmpty()) {
+      return "Needs triage";
+    }
+    String title = issueToTitle.get(key);
+    Status status = Status.valueOf(issueStatus.get(key).toUpperCase(Locale.ROOT));
+    return "<span title=\"" + status.title + "\">" + status.icon
```
niloc132: General note that title isn't as good as aria label, and doesn't show up on touch screens. A key or heading or some other hint is generally preferred, but this is the first draft of a new tool, so I'm not too worried.

niloc132 (commented):
Nothing dealbreaking, just want to discuss, see if anything is worth updating.

Also, we discussed privately that it likely makes sense to split out emul before merging, handle those changes separately.

build_tools/doctool/src/com/google/doctool/custom/missing/triage.properties
zbynek: This was a mistake. In the last commit I changed the file location to avoid accidental commits. Also it will fail the build from now on.

doc/build.xml
```diff
@@ -148,8 +148,14 @@
                  packagenames="${JAVA_PKGS}"
                  docletpath="${project.build}/../build_tools/doctool/bin"
                  doclet="com.google.doctool.custom.JavaEmulSummaryDoclet">
-          <arg value="-outfile" />
+          <arg value="-outFile" />
           <arg value="${project.build}/emul-ezt/fragment.html" />
+          <arg value="-missingFile" />
+          <arg value="${project.build}/emul-ezt/fragment.html" />
```
niloc132:
```suggestion
          <arg value="${project.build}/emul-ezt/fragment-missing.html" />
```
right?

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
zbynek: It's now in a separate method.

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
zbynek: Maybe adding dependency on gwt-user (or a third party library) will be easier with mavenized build, for now this should be good enough.

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
zbynek: Changed to aria-label and made visible as tooltip via CSS https://codepen.io/zbynek/pen/NPWBzKP (the CSS should go to gwt-site)

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
zbynek: I'm not sure why would the links point to another repository's issues. If someone is e.g. drafting a PR that would categorize the methods currently bunched in #10228, they can either just create the issues in gwtproject's repo prior to getting the PR reviewed, or they can use TBD in the filenames until the PR is reviewed and create the issues after. I don't see any advantage in creating "draft" issues in a fork and copying them into gwtproject's repo later. Or did you have some other usecase in mind?

niloc132: My idea is for the various GWT forks out there that aren't pushing fixes back to gwtproject/gwt, but just keeping them local (licensing, lack of tests, etc). They might want to still use this feature in some way, but will probably want to track locally rather than in the upstream repo.

zbynek: OK, I added the `link` field handler, but it still defaults to a link extracted from the filename.

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -16,281 +16,517 @@
...
+  private Reporter reporter;
+  private String outputFile;
+  private String missingFile;
+  private String triageFile;
+  private String missingPropertiesDir;
+
+  // the key in the following maps is the full URL of an issue
+  private final Map<String, String> issueToSignatures = new HashMap<>();
+  private final Map<String, String> issueToTitle = new HashMap<>();
+  private final Map<String, String> issueStatus = new HashMap<>();
```
niloc132: this is begging to be a Map<String, MyRecord> to hold all three strings in a single class/record

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -16,281 +16,517 @@
...
+  }
+
+  private String extractIssueLink(String filename) {
+    return "https://github.com/gwtproject/gwt/issues/" + filename.replaceAll("gh|\\.properties", "");
+  }
+
+  private Stream<String> withInnerClasses(Element clazz, PackageElement pack) {
+    return Stream.concat(
+      Stream.of(pack.getQualifiedName() + "." + clazz.getSimpleName().toString()),
+      clazz.getEnclosedElements().stream()
```
niloc132, line 262: technically should probably check inner-inner classes? not sure if that ever happens in the JRE though.

I see two in GWT's emul: 
Collections.UnmodifiableEntrySet.UnmodifiableEntry
EnumSet.EnumSetImpl.IteratorImpl

both are private, so wouldn't apply here.

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -16,281 +16,517 @@
...
+        .findFirst().orElse(null);
+    }
+    if (c == null) {
+      if (CURRENT_JRE_VERSION >= MAX_JRE_VERSION) {
+        throw new RuntimeException("Class does not exist in JRE: " + parts[0]);
+      } else {
+        System.out.format("Class %s not supported in Java %s%n", parts[0], CURRENT_JRE_VERSION);
+        return;
+      }
+    }
```
niloc132: If you move this up, you can drop the null check on 258

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -16,281 +16,517 @@
...
+    Class<?> superclass = c.getSuperclass() != null ? c.getSuperclass() : Object.class;
+    List<Method> superMethods = new ArrayList<>(Arrays.asList(superclass.getMethods()));
+
+    for (Class<?> parentInterface: c.getInterfaces()) {
```
niloc132:
```suggestion
    for (Class<?> parentInterface : c.getInterfaces()) {
```

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -16,281 +16,517 @@
...
+      if (!allClasses.contains(parentInterface.getTypeName())) {
+        System.out.println("Missing interface for " + c + ": " + parentInterface.getTypeName());
```
niloc132: why mix and match System.out.println/format log lines?

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -16,281 +16,517 @@
...
+          .collect(Collectors.groupingBy(signature ->
+              getStatus(pack + cls.getSimpleName() + "#" + signature)));
+      if (missingMemberGroups.containsKey("")) {
+        missingMemberGroups.get("").stream()
+            .map(s -> pack + cls.getSimpleName() + "#" + s)
+            .forEach(triage::add);
+      }
+      pwMissing.format("%n  <dt><a href=\"%s%s.html\">%s</a></dt>%n", packURL,
+        qualifiedSimpleName(cls), qualifiedSimpleName(cls));
+      for (Map.Entry<String, List<String>> entry: missingMemberGroups.entrySet()) {
```
niloc132:
```suggestion
      for (Map.Entry<String, List<String>> entry : missingMemberGroups.entrySet()) {
```

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -16,281 +16,517 @@
...
+    }
+    String title = issueToTitle.get(key);
+    Status status = Status.valueOf(issueStatus.get(key).toUpperCase(Locale.ROOT));
+    return "<span class=\"issueStatus\" aria-label=\"" + status.title + "\">" + status.icon
+        + "</span>#" + key.replaceFirst("^.*/(\\d+)$", "$1")
+        + (title == null ? "" : " (" + title + ")");
+  }
+
+  private boolean isFieldFromSuper(Field field, Class<?> c) {
+    for (Class<?> parentInterface: c.getInterfaces()) {
```
niloc132:
```suggestion
    for (Class<?> parentInterface : c.getInterfaces()) {
```

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
```diff
@@ -16,281 +16,517 @@
...
-        };
-        return new HashSet<>(Arrays.asList(options));
+    return "";
+  }
+
+  private List<String> getMethodNames(Element cls, Function<TypeMirror, String> typeNamer) {
+    return cls.getEnclosedElements()
+      .stream()
+      .filter(element -> ElementKind.METHOD == element.getKind())
+      .filter(member -> member.getModifiers().contains(Modifier.PUBLIC))
```
niloc132: There aren't many, but there are some protected methods intended to be overridden/used by JRE subclasses

AbstractList.removeRange(int,int)
Random.next(int)
PrintStream.setError
PrintStream.clearError
MessageDigestSpi.engineDigest()


This abbreviated list is from searching our own JRE emul, there could be more I'm missing (since we are missing them).

zbynek: Do we want to include protected methods in both fragments (implemented and missing)?

niloc132: WIthout knowing what we're missing its hard to saw if we need them. I could see not listing them in the implemented page, since most users won't hit them (I could also see listing them so that we offer them for cases where a user might subclass something), but if it is missing from the implementation it seems like that could mean the implementation is incorrect/incomplete?

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
zbynek: Thanks for pointing this out, the whole nested class lookup by name seems unnecessarily complicated.

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
zbynek: Changed to use recursion.

build_tools/doctool/src/com/google/doctool/custom/JavaEmulSummaryDoclet.java
zbynek: The only missing protected methods are `Object#clone()` + its overrides, `MessageDigestSpi#engineUpdate(ByteBuffer)`, `ThreadLocal#initialValue()`.
In total there are 88 protected methods + constructors and 14 fields, so including them does not bloat the page too much and it's consistent with default Javadoc pages that mention [protected fields](https://docs.oracle.com/en/java/javase/21/docs/api//java.base/java/io/ByteArrayOutputStream.html) and [methods](https://docs.oracle.com/en/java/javase/21/docs/api//java.base/java/lang/Object.html).

build_tools/doctool/src/com/google/doctool/custom/missing/gh9630.properties
```diff
@@ -0,0 +1,92 @@
...
+java.lang.Class#getAnnotatedSuperclass()\
+java.lang.Class#getAnnotatedInterfaces()\
+java.lang.Class#isNestmateOf(Class)\
+java.lang.Class#getNestMembers()\
+java.lang.Class#isSealed()\
+java.lang.Class#forPrimitiveName(String)\
+java.lang.Class#accessFlags()\
+java.lang.Enum.EnumDesc\
+java.lang.Enum#clone()\
+java.lang.Object#clone()\
```
niloc132: https://github.com/gwtproject/gwt/issues/5067 - I think this is possible, with just a little bending of the rules for JSNI/JSO restrictions.

build_tools/doctool/src/com/google/doctool/custom/missing/gh9630.properties
```diff
@@ -0,0 +1,92 @@
...
+java.lang.Class#asSubclass(Class)\
+java.lang.Class#getAnnotatedSuperclass()\
+java.lang.Class#getAnnotatedInterfaces()\
+java.lang.Class#isNestmateOf(Class)\
+java.lang.Class#getNestMembers()\
+java.lang.Class#isSealed()\
+java.lang.Class#forPrimitiveName(String)\
+java.lang.Class#accessFlags()\
+java.lang.Enum.EnumDesc\
+java.lang.Enum#clone()\
```
niloc132: This is trivial, just `return this`

niloc132 (commented):
Looks great, approved - future changes can be PRs to tweak these files as issues are made/updated.

niloc132 (dismissed):
(actually clicked approve this time)

build_tools/doctool/src/com/google/doctool/custom/missing/gh9630.properties
zbynek: The JVM implementation just throws an exception for Enums. 
Associating this with #5067 now.

zbynek:
I just split out some methods from the "leftover" ticket and updated the `clone` methods -- should be good to go from my side now.

niloc132 (approved):

PR #10231, opened 2025-12-27
StaticResourceServer must only serve static content

### LINKED ISSUE #10230: StaticResourceServer can run servlets?
Some testing of the StaticResourceServer as part of testing https://github.com/gwtproject/gwt/pull/10103 reveals that it seems to run servlets, at least in some cases. This surprises me, as I had confirmed that the web.xml was fetched as plain content rather than being treated as application code as it should be. With all Configuration types removed, I'm not sure how it is working... but it is working.

#9866 
#9863

As this defeats the purpose of having a "static" resource server for simple testing, this bug is to ensure that the webserver _only_ serves static content so we can confidently replace its implementation in the future.

niloc132:
Fixes #10230


niloc132:
Tested along with #10103 by confirming that the DynaTable example has a GWT-RPC error on startup, which can be fixed by either supplying a third-party server (service loaded via #10103 or just specified via `-server fully.qualified.ServerName`), or explicitly using the JettyLauncher.

tbroyer (approved):

Issue #10048, opened 2024-11-23
The types returned by newly added List/Set/Map.of/copyOf() cannot be passed between client and server (=are outside the scope of gwt rpc serialization)


andreaskornstaedt:
**GWT version: 2.12.1
**Browser (with version): any
**Operating System: any

---

##### Description
For example, having a GWT service method that returns a `List` instance created using `List.of()` will yield the following at runtime:

```
2024-11-23 18:36:41.736:WARN:oejshC.ROOT:qtp29640057-62: Exception while dispatching incoming RPC call
com.google.gwt.user.client.rpc.SerializationException: Type 'java.util.ImmutableCollections$ListN' was not included in the set of types which can be serialized by this SerializationPolicy or its Class object could not be loaded. For security purposes, this type will not be serialized.: instance = [1, 2, 3]
	at com.google.gwt.user.server.rpc.impl.ServerSerializationStreamWriter.serialize(ServerSerializationStreamWriter.java:699)
	at com.google.gwt.user.client.rpc.impl.AbstractSerializationStreamWriter.writeObject(AbstractSerializationStreamWriter.java:130)
	at com.google.gwt.user.server.rpc.impl.ServerSerializationStreamWriter$ValueWriter$8.write(ServerSerializationStreamWriter.java:169)
	at com.google.gwt.user.server.rpc.impl.ServerSerializationStreamWriter.serializeValue(ServerSerializationStreamWriter.java:609)
	at com.google.gwt.user.server.rpc.RPC.encodeResponse(RPC.java:644)
	at com.google.gwt.user.server.rpc.RPC.encodeResponseForSuccess(RPC.java:497)
	at com.google.gwt.user.server.rpc.RPC.invokeAndEncodeResponse(RPC.java:589)
	at com.google.gwt.user.server.rpc.RemoteServiceServlet.processCall(RemoteServiceServlet.java:349)
	at com.google.gwt.user.server.rpc.RemoteServiceServlet.processCall(RemoteServiceServlet.java:319)
	at com.google.gwt.user.server.rpc.RemoteServiceServlet.processPost(RemoteServiceServlet.java:389)
	at com.google.gwt.user.server.rpc.AbstractRemoteServiceServlet.doPost(AbstractRemoteServiceServlet.java:62)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:648)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:729)
[...]

PR #10245, opened 2026-01-06
Add missing interfaces to emulated classes

zbynek:
Some missing interfaces found when working on #10208

Out of scope here:
* adding default implementations to `RandomGenerator` (#10214)
* adding `Serializable` to any classes
* implementing methods in `Constable` / `ConstantDesc` -- only use those as tagging interfaces for now
* adding `Readable` interface as it depends on `nio`
* adding `Cloneable` to `Locale`
* adding any new interfaces to `java.lang.reflect` or `java.lang.invoke` implemented by `Class`


jnehlmeier:
In a GWT fork at work I had added `Serializable` to `ArrayDeque` (and added a GWT-RPC `CustomFieldSerializer` for it) but `ArrayDeque` does not appear here. Maybe you missed some more classes?

niloc132:
There are definitely some others that are missing too, mostly collections - I think those are going to get a closer look before adding them.

zbynek:
> Maybe you missed some more classes?

I did not add `Serializable` to any class here and I'm not planning to do that, I'm leaving that to someone with more experience with GWT RPC. 

Out of public classes, these are serializable in JRE but not in GWT:
 
* `java.lang.Class`
* `java.lang.StringBuffer`
* `java.lang.StringBuilder`
* `java.util.AbstractMap$SimpleEntry`
* `java.util.AbstractMap$SimpleImmutableEntry`
* `java.util.ArrayDeque`
* `java.util.BitSet`
* `java.util.EnumMap`
* `java.util.EnumSet`
* `java.util.EventObject`
* `java.util.Locale`
* `java.util.PriorityQueue`
* `java.util.Random`
* `java.util.concurrent.ConcurrentHashMap`
* `java.util.concurrent.atomic.AtomicReference`
* `java.util.concurrent.atomic.AtomicReferenceArray`

Also there are other issues related to serializing private collection classes (#10048, #3071)

niloc132 (approved):
LGTM, we can try to address some of the others later, though some (Locale, Random, Class to name three) probably won't make sense ever.

Issue #10011, opened 2024-10-17
Explore replacing build/project setup and modularizing codebase


niloc132:
While I'm not familiar with the history within Google of this project and how it was built, externally it has used Ant since first being open sourced. 

It appears that at least part of the build used Bazel/Blaze, but this was never formally used in the open source CI tooling (Jenkins while Google managed the project and GitHub Actions since then). It was used at least in a limited fashion for https://github.com/google/j2cl to depend on some GWT classes, but those classes have since been copied into the j2cl repository. Once there was a Bazel rules repository for GWT (https://github.com/bazelbuild/rules_gwt/), but I never saw it get real maintenance, and it pulled GWT and its dependencies from Maven Central, rather than actually using the Bazel build files in the GWT repository.

Additionally, the build files are pretty un-idiomatic Bazel, with huge globs to include hundreds thousands of classes (rather than dozens) in a single `java_library` - Bazel was likely added after the fact, with Ant as the primary way to build.

Ant's own usage surprised me 15+ years ago when I personally started working with building GWT itself - it is slow to build, there are a lot of undocumented properties that can be set to control it, IDEs are difficult to configure to use it, and the "build" code has historically been a blend of xml, Python, Java, and shell. The outputs are generally consumable from other platforms, but it can often take substantial decoding work to understand why some detail of the output is the way that it is.

Some specific goals for this task:

 * Support running the compiler without com.google.gwt.user.User and its associated classpath. This will require separating the "gwt language" and "java internal implementation" details out into their own artifact. Multiple benefits:
   * Allows decoupling "gwt-user" and "gwt-dev" versions from each other, by declaring which parts of GWT are required by the compiler, and which parts are suggested runtime code.
   * Allows dropping legacy `com.google.gwt` client/shared packages in favor of migrated external `org.gwtproject` migrated artifacts, with no unused classpath elements (flute, tapestry, gwt-user itself, etc)
 * Update long out of date dependencies, and let automated tooling help keep them up to date (e.g. the "Selenium" support is using 1.0-beta-1 of selenium-java-client-drivers, a release that can't even be found in maven central). Most of these dependencies can be obtained from maven central with no modifications, and the rest we can continue to source from the tools repo.
[...]

PR #10247, opened 2026-01-12
Rewrite SDK samples using Maven

### LINKED ISSUE #10232: Rewrite samples to move away from Ant
Samples use DevMode rather than CodeServer, and rather than updating ant templates/examples, it probably makes more sense to just move to maven. Existing reactor archetypes can be used to split server from client process, and encourage that for all new "getting started" projects. Projects that don't need a server can either use devmode with the static server, or just use the codeserver and serve the html file via public/. An example (Hello might be a good fit) that just uses CodeServer standalone also would need updated documentation (rather than the current WebAppCreator template readme).

WebAppCreator's behavior may be left alone, but the deprecation warnings should perhaps direct to a better "getting started with maven" location.

Gradle could be encouraged as well, but that's its own topic, see #10028.

niloc132:
Restructured samples to build with Maven and be designed around splitting server/client where appropriate. Includes some package changes to better differentiate when a class is shared, client-only, or server-only.

RequestFactory log requests with jakarta seem hard to use, but easy to copy into the sample project as an example of how to get around this issue.

All samples are rebuilt to use jakarta.servlet, except mobilewebapp, which uses too old of an AppEngine version to update.

Fixes #10232


common.ant.xml
```diff
@@ -144,6 +144,17 @@
     </ant>
   </presetdef>
 
+  <macrodef name="gwt.mvn">
+    <attribute name="args" default="${mvn.args}" />
+    <attribute name="dir" />
+    <sequential>
+      <echo message="Running 'mvn @{args}' in @{dir}"/>
+      <exec dir="${basedir}" executable="mvn" failonerror="true">
```
niloc132, line 152: There are indications that this may not work on windows. I'm inclined to leave it - samples aren't critical to a build, and 2.14 shouldn't get far without landing #10011 - if it does, we might need a grosser ant setup.

samples/dynatablerf/dynatablerf-shared/src/main/java/com/google/gwt/sample/dynatablerf/shared/LoggingRequest.java
```diff
@@ -0,0 +1,40 @@
+/*
+ * Copyright 2025 GWT Project Authors
```
niloc132: 2026

samples/dynatablerf/dynatablerf-server/pom.xml
```diff
@@ -0,0 +1,190 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.google.gwt.sample.dynatablerf</groupId>
```
niloc132, line 6: groupids should be consistent - leaning to leave them all as `com.google.gwt.sample.*` to match the existing projects and package structure.

samples/mail/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+  <packaging>pom</packaging>
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.gwtproject</groupId>
+        <artifactId>gwt</artifactId>
+        <version>2.12.2</version>
```
niloc132: HEAD-SNAPSHOT

distro-source/build.xml
```diff
@@ -56,14 +55,9 @@
...
-        <include name="samples/*/build.xml" />
-        <include name="samples/*/pom.xml" />
-        <include name="samples/*/README-MAVEN.txt" />
-        <include name="samples/*/src/**" />
-        <include name="samples/*/war/**" />
-      </zipfileset>
-      <zipfileset dir="${gwt.build.out}/samples-scripts" prefix="${project.distname}/samples">
-        <include name="*/*" />
+        <include name="samples/**/pom.xml" />
+        <include name="samples/*/README.txt" />
```
zbynek:
```suggestion
        <include name="samples/*/README.md" />
```
Should `COPYING` be also included?

samples/build.xml
```diff
@@ -51,39 +47,85 @@
...
 
   <target name="source" description="Copy source to the output folder">
-    <antcall target="-do">
-      <param name="target" value="source" />
-    </antcall>
+    <copy todir="${project.build}">
+      <fileset dir=".">
+        <include name="**/src/**" />
+        <include name="**/pom.xml" />
+        <include name="*/README.txt" />
```
zbynek:
```suggestion
        <include name="*/README.md" />
```

user/src/com/google/gwt/user/tools/WebAppCreator.java
```diff
@@ -142,29 +141,6 @@ public boolean setString(String str) {
     }
   }
 
-  private final class ArgHandlerLegacyJetty extends ArgHandlerFlag {
```
zbynek, line 145: Is this intentionally removed?

samples/dynatable/pom.xml
```diff
@@ -0,0 +1,118 @@
+<?xml version="1.0" encoding="UTF-8"?><project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.google.gwt.sample.dynatable</groupId>
+  <artifactId>dynatable</artifactId>
+  <version>1.0-SNAPSHOT</version>
+  <packaging>pom</packaging>
+  <properties>
+    <gwt.version>HEAD-SNAPSHOT</gwt.version>
```
zbynek, line 8: Should the READMEs suggest to change `gwt.version` to something stable?

user/src/com/google/gwt/user/tools/WebAppCreator.java
niloc132: Yes - it was only temporarily added (and immediately deprecated) so that samples wouldn't break. Now that samples don't rely on WebAppCreator to be created and built, we don't need this (and no sample relies on this feature any more either). Projects should not rely on this flag, but own for themselves how they will manage client/server classpath splits, or manage the merged classpath and Jetty runner themselves. If we commit to maintaining this flag, then we can't remove JettyLauncher, and so can't update Jetty to something that isn't (three major versions...) EOLd.

zbynek: Makes sense. Then we can also remove `@serverFQCN` from the xml template.

common.ant.xml
zbynek: `mvn verify` works for each sample individually on Windows. To make it work with `ant` I had to change this to

```suggestion
      <condition property="maven.executable" value="mvn.cmd" else="mvn">
        <os family="windows" />
      </condition>
      <exec dir="${basedir}" executable="${maven.executable}" failonerror="true">
```
It's 3 extra lines, not sure if worth adding.

zbynek (changes requested):
I think `@serverFQCN` should be removed from `build.xmlsrc`.

For building samples on Windows I think the fix is quite simple, but I don't have a strong opinion on including it.

Otherwise I couldn't find any issues.

zbynek (dismissed):

niloc132:
@vegegoku has found some issues in testing - I've confirmed the first of four and have a doc fix at least, will follow up on the others.

vegegoku (approved):

zbynek (approved):

PR #10249, opened 2026-01-13
Make jaxb optional for validation tests, remove from classpath

niloc132:
Modifies Validation and RequestFactory internals to optionally skip loading validation configuration from XML, defaulting to old behavior. Tests are updated to activate this new behavior, confirmed by removing the jaxb jar from the classpath.

Also modified the validation sample.

### DIFF
--- a/samples/validation/src/main/java/com/google/gwt/sample/validation/server/GreetingServiceImpl.java
+++ b/samples/validation/src/main/java/com/google/gwt/sample/validation/server/GreetingServiceImpl.java
@@ -27,6 +27,7 @@ import org.hibernate.validator.engine.ValidationSupport;
 import java.util.HashSet;
 import java.util.Set;
 
+import javax.validation.Configuration;
 import javax.validation.ConstraintViolation;
 import javax.validation.ConstraintViolationException;
 import javax.validation.Validation;
@@ -40,7 +41,15 @@ import javax.validation.groups.Default;
 public class GreetingServiceImpl extends RemoteServiceServlet implements
     GreetingService {
 
-  private final Validator validator = Validation.buildDefaultValidatorFactory().getValidator();
+  private final Validator validator;
+
+  public GreetingServiceImpl() {
+    Configuration<?> config = Validation.byDefaultProvider().configure();
+    if ("true".equals(System.getProperty("gwt.validation.ignoreXml"))) {
+      config = config.ignoreXmlConfiguration();
+    }
+    validator = config.buildValidatorFactory().getValidator();
+  }
 
   public SafeHtml greetServer(Person person) throws IllegalArgumentException,
       ConstraintViolationException {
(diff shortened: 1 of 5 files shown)


samples/validation/src/main/java/com/google/gwt/sample/validation/server/GreetingServiceImpl.java
```diff
@@ -40,7 +41,15 @@
 public class GreetingServiceImpl extends RemoteServiceServlet implements
     GreetingService {
 
-  private final Validator validator = Validation.buildDefaultValidatorFactory().getValidator();
+  private final Validator validator;
+
+  public GreetingServiceImpl() {
+    Configuration<?> config = Validation.byDefaultProvider().configure();
+    if ("true".equals(System.getProperty("gwt.validation.ignoreXml"))) {
```
tbroyer: nit: use Boolean.getBoolean() ?

vjay82: imho Boolean.getBoolean() is really bad API and misleading in regards of what it does

niloc132: Ironically, I am thinking I'll remove the condition from this file anyway (a sample doesn't need this, it controls its own classpath).

That said, for the other two sites, I do have a very slight preference for an explicit "is this property equal to the string `true` or not" as opposed to getBoolean's squishier contract ("exists and is equal to the string `"true"`."... but also case insensitive, plus exactly the same signature as valueOf/parseBoolean).

GWT is inconsistent here. CachedCompilationUnit and ResourceAccumulator pass System.getProperty to parseBoolean, CoverageInstrumenter, EnumOrdinalizer, JettyLauncher, Memory just check System.getProperty for being null or not. CodeSplitter, CompilingClassLoader, RemoteServiceServlet use Boolean.getBoolean.

zbynek:
From the previous PR:

> This PR fixes an issue with the dynatablerf example. Starting with GWT 2.11.0 the jabx dependencies are no longer available and need to be added.

So there is one known instance where *not* ignoring the XML configuration breaks a build.
Is there any known instance where ignoring the XML configuration could cause problems?
If not, could we swap the default so that XML config is ignored by default and a people can opt in using a property, should they need it for some unforeseen reasons?

niloc132:
I don't know of any, but rather than break builds by default, I wanted to have it be opt-in. I'm not sure what level of knowledge would be required to be confident either way.

One note though, GWT has not "officially" supported Java 8 in a while now, so teams that have been running into this issue have already fixed them in other ways - I don't expect that there are many projects which have been waiting on this fix. What this fix does then it just make it simpler to use old javax validation in new projects (other options include adopting [jakarta validation](https://gitlab.com/ManfredTremmel/gwt-bean-validators)).

zbynek (approved):

PR #10254, opened 2026-01-24
Run OptimizedOnlyCompilerSuite when draft is disabled

niloc132:
Somehow this suite was permanently disabled, and one test has broken in the meantime due to String.contains being simplified to the point it can more easily be inlined. When #10147 is fixed, this can be re-enabled.

Fixes #10253


zbynek:
With the current setup this won't run in CI. Would it be worth it to run this suite e.g. in a separate `ant` invocation in `full-check`, that would only run in optimized HTMLUnit?

niloc132:
> With the current setup this won't run in CI. Would it be worth it to run this suite e.g. in a separate `ant` invocation in `full-check`, that would only run in optimized HTMLUnit?

Ah, that's no good. It'll increase the build time, but I think worth it to get the `test.web.htmlunit` restored. Probably should look into `test.nometa.htmlunit` as well...

For what its worth, I usually run  `ant test.draft.htmlunit test.web.htmlunit` locally so it fails slightly faster if there is something egregiously wrong, but then also runs through the optimized version as well to be sure prod builds won't suffer too.

zbynek:
Maybe the CI could run draft tests with Java 11, 17, 22 and optimized tests on Java 21?

vegegoku (dismissed):

zbynek (dismissed):
Being able to run it locally is an improvement worth merging. CI changes may need a separate PR.

niloc132:
I tweaked the github actions wiring to also run optimized tests, now running at https://github.com/niloc132/gwt/actions/runs/23264900880/job/67642962566

```
+ ant clean test dist doc '-Dtest.jvmargs=-ea -Dgwt.watchFileChanges=false' -Dtest.web.htmlunit.disable=false -Dtest.nometa.htmlunit.disable=true -Dtest.emma.htmlunit.disable=true
```

We could make them mutually exclusive, but I'd really like to also eventually add various browser tests via selenium, as we have a few failures there (and it would help catch some changes like https://github.com/gwtproject/gwt/issues/10292).

niloc132:
...Excitingly, the `test.*.disable` flags seem to not booleans, but trigger if they are present at all, regardless of value (including blank).

PR #10257, opened 2026-01-26
Safer ExternalPermutationWorkerFactory connections

niloc132:
Cookies are now sent/received as 32 ascii chars before any ObjectOutputStream data is sent, and the server validates before returning the socket. An invalid socket is closed and ignored, and the next valid socket will be used instead.

Also improved cookie randomness source.

Fixes #10256


vegegoku (commented):
LGTM, might need to fix the checkstyle warnings.

dev/core/src/com/google/gwt/dev/ExternalPermutationWorkerFactory.java
```diff
@@ -61,47 +62,53 @@ public class ExternalPermutationWorkerFactory extends PermutationWorkerFactory {
...
-        throw new IllegalStateException("Too many calls to accept()");
-      }
-
-      try {
-        return sock.accept();
-      } finally {
-        if (--accepts == 0) {
-          sock.close();
-          sock = null;
+      while (!validCookies.isEmpty()) {
```
zbynek, line 84: The old code did throw an error on invalid "cookie", this seems to retry until a valid "cookie" is sent, is that an intentional change?

niloc132: Yes - rather than permit a "denial of service" where an attacking client can guess the port but get the cookie wrong and thus kill the compile job, this implementation just ignores the client since they aren't connecting correctly anyway.

If I've done this right, we still warn about the bad connection, but we don't let it kill the compiler. The only way kill the job is if it _was_ the right client, but somehow had the wrong cookie, which would have failed in the old code too.

zbynek (dismissed):

dev/core/src/com/google/gwt/dev/ExternalPermutationWorkerFactory.java
niloc132: Restating this slightly:
 * "good" clients will effectively never have this issue, but _if they do_, we will get a warning, and the client itself will crash out.
 * "bad" clients will no longer be able to just guess one of the outstanding cookies and take down the server. Technically a malicious client could guess and guess and never be rate limited... but they have 128 bits of search space to work through, which is infeasible.

zbynek (approved):
Code looks good, didn't test interactively.

(Btw. it's not clear why are those strings called cookies, but that naming didn't change here.)

niloc132:
I know I've seen other programs refer to "cookies" as auth/session tokens even when not http-related (dhcp comes to mind?), but as far as I can tell this practice originated with http's use of the term.

Also, http://www.catb.org/%7Eesr/jargon/html/M/magic-cookie.html

Issue #10242, opened 2026-01-05
Experiment with internal annotation `@HasNoSideEffects`


niloc132:
Presently JExpression has a `hasSideEffects` method that only reads from the node's constituent expressions to determine if there might be a side effect possible. For method calls, the only way that this can be set is if the method itself has the annotation. This can mostly be found on hard-to-inline methods that definitely have no user-observable side effects, such as JSNI or value boxing. Inlining is important here - if a method could be inlined, then the side effect check can happen on the contents of the method directly, and we don't need to worry about whether the method itself had side effects.

Definitionally, I've assumed "side effects" to mean "observable effects by the program outside of the return value". This means if an exception could be thrown, the call cannot be pruned, but on the other hand most constructors _can_ be pruned. To ease the implementation I've assumed that instance methods always may have side effects (both to ignore NPEs from an instance being null and to avoid looking up and down the type hierarchy for other calls that could have side effects), and trust that if the call is simple enough we can let it be made static and then address it that way. 

The current patch marks the following functions as having no side effects:
 * static methods that 
   * call only other side-effect free functions
   * only assign locals, params, and fields of values assigned locally (created by other side effect free funcs, not passed from a param or accessed from a some other field, etc)
   * never throw exceptions
 * constructors as above, plus
   * can call their own instance initializer
   * can assign their own fields, qualified by "this"

There are definitely use cases where functions produce no side effects yet are not marked as such:
 * if a function calls another function that mutates a param, but calls that function on a fresh local, that should be allowed. In the following case, ArrayList.add might become a static method and thus take `list` as a param, but will mutate one of its fields. Since ArrayList.add then isn't itself side-effect free, split() will not be considered to be side effect free either
    ```java
    public static List<String> split(String str) {
        List<String> list = new ArrayList<>();
        list.addAll(Arrays.asList(str.split(",")));
        return list;
    }
    ```

---

My expectation was that if the compiler could figure out for itself which methods/constructors had no side effects, we could fix https://github.com/gwtproject/gwt/issues/10171 (and possibly https://github.com/gwtproject/gwt/issues/10223 or at least improve it). The sample implementation I've built is able to prune a few obvious calls
 * boxing of primitives, such as via Integer.valueOf()
 * constructor calls of unhandled events, such as RowHoverEvent - in this case the event was being created but never fired since there was no handled compiled for it
 * other constructor calls, such as those created to initialize static fields, thus removing clinits entirely
 * complex method that doesn't change anything, but result is always unused

[...]

PR #10261, opened 2026-01-30
Simplify boolean expressions in JavaScript AST

### LINKED ISSUE #10280: Improve simplification of boolean expressions in JS AST
In the output I noticed statements like

`!!a&&a.b();`
which can be simplified to

`a&&a.b();`

To do that, `JsStaticEval` needs to keep better track of expressions that are coerced to boolean. By doing that, some more simplifications can be done that align the functionality of `DeadCodeElemination` with `JsJstaticEval`.

zbynek:
In the output I noticed statements like

```
!!a&&a.b();
```
which can be simplified to
```
a&&a.b();
```
In general, if we know that we ignore the result of an evaluation or we know the result is coerced to a boolean, we can apply more simplifications.

---
Other examples from compiler output this should simplify:
```
a.K==(Slh(),Klh)&&false?(n.g+=TWo,n):tNg(a,n,g)
O7d(e.yb)&&Tde(e.Qb,e.yb)&&false?Tqe(e):Pqe(e)
```

Fixes #10280


dev/core/test/com/google/gwt/dev/js/JsStaticEvalTest.java
```diff
@@ -43,7 +43,8 @@ public void testAssociativity() throws Exception {
     assertEquals("alert(a&&b||c&&d);",
         optimize("alert((a && b) || ( c && d));"));
     assertEquals("a(),b&&c();", optimize("a(), b && c()"));
-    assertEquals("a()&&b,c();", optimize("a() && b, c()"));
+    assertEquals("a()&&b(),c();", optimize("a() && b(), c()"));
```
niloc132, line 46: If the expr is in a statement like this, the output _should_ be `a(),b(),c()`. Doesn't necessarily have to be in this patch, but I think it could be fixed by passing the context as void for the whole exprstatement?

dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
```diff
@@ -144,6 +144,11 @@ public boolean visit(JsFunction x, JsContext ctx) {
     }
   }
 
+  private enum EvalMode {
+    BOOL,
+    VOID
```
niloc132: would you document the meaning of these, and what it means for `evalContext` to not have a key at all?

My understanding from (heh) context is that VOID stands for "value will be thrown away", and BOOL means "this will be coerced to boolean", while absent means we don't know, so retain everything

dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
```diff
@@ -153,12 +158,29 @@ public boolean visit(JsFunction x, JsContext ctx) {
...
+    }
+
+    public boolean visit(JsBinaryOperation x, JsContext ctx) {
+      if (evalContext.containsKey(x)
+          && (x.getOperator() == JsBinaryOperator.AND || x.getOperator() == JsBinaryOperator.OR)) {
+        evalContext.put(x.getArg1(), EvalMode.BOOL);
+        evalContext.put(x.getArg2(), evalContext.get(x));
+      }
+      if (x.getOperator() == JsBinaryOperator.COMMA) {
+        evalContext.put(x.getArg1(), EvalMode.VOID);
```
niloc132, line 187: i think we also want
`evalContext.put(x.getArg2(), evalContext.get(x));`, if it was set for `x`

dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
```diff
@@ -153,12 +158,29 @@ public boolean visit(JsFunction x, JsContext ctx) {
...
+      evalContext.put(x.getExpression(), EvalMode.VOID);
+      return true;
+    }
+
+    public boolean visit(JsBinaryOperation x, JsContext ctx) {
+      if (evalContext.containsKey(x)
+          && (x.getOperator() == JsBinaryOperator.AND || x.getOperator() == JsBinaryOperator.OR)) {
+        evalContext.put(x.getArg1(), EvalMode.BOOL);
+        evalContext.put(x.getArg2(), evalContext.get(x));
+      }
```
niloc132: else

dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
```diff
@@ -167,13 +189,20 @@ public void endVisit(JsBinaryOperation x, JsContext ctx) {
...
       JsExpression result = x;
-
-      if (op == JsBinaryOperator.AND) {
-        result = shortCircuitAnd(x);
+      if (evalContext.get(x) == EvalMode.VOID
+            && !arg2.hasSideEffects() && !op.isAssignment()) {
+        result = arg1;
+      } else if (op == JsBinaryOperator.AND) {
+        result = shortCircuitAnd(x, evalContext);
+        evalContext.remove(arg1);
```
niloc132: note that in the first if check, we're failing to remove evalContext entries for the three cases they could be added

dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
```diff
@@ -167,13 +189,20 @@ public void endVisit(JsBinaryOperation x, JsContext ctx) {
...
+        evalContext.remove(arg2);
       } else if (op == JsBinaryOperator.OR) {
         result = shortCircuitOr(x);
+        evalContext.remove(arg1);
+        evalContext.remove(arg2);
       } else if (op == JsBinaryOperator.COMMA) {
         result = trySimplifyComma(x);
+        evalContext.remove(x.getArg2());
```
niloc132: going by the patch as it stands, this should be `x.getArg1()` or just `arg1`

if you accept my suggest, we'll have to remove both

dev/core/src/com/google/gwt/dev/js/ast/JsBinaryOperation.java
```diff
@@ -68,6 +68,26 @@ public boolean isDefinitelyNull() {
     return false;
   }
 
+  @Override
+  public boolean isBooleanFalse() {
```
niloc132: I don't think this is the right place for this - it should be part of the JsStaticEval check. the comments below indicate that the converse is "already simplified", but that's only true post-JsStaticEval.

The same code that does the shortCircuitOr()/etc check should also be looking for this case - you added logic to shortCircuitAnd, does it not work for OR too?

If we need this here, we probably also need to test for `,` operations... those are messier since they imply side effects on the LHS (JsStaticEval.trySimplifyComma should remove those though).

DeadCodeElimination/DCE and Simplifier do another transformation that doesnt appear to happen in JsStaticEval - expressions like `(a,b)?c:d` are rewritten to `(a, b?c:d)` so that `b` can be statically eval'd by the Conditional. If you're running into a situation where JsStaticEval can't work out the value of a sub-expr, it could be from something like this?

DCE at least has some checks in tests to assert that it only runs once and always converges - JsStaticEval is written in a similar style, but doesn't have such a check.

niloc132 (commented):
I like the patch - it seems to go a little beyond what is discussed in the description though. Can you describe a bit more how the JS is being generated that causes this - is it from Java that isn't handled correctly, or is this JSNI (or jsinterop-base's JSNI) emitting these extra `!!`s? Might make sense to have an issue to generally track the topic too, in case there are more examples of it.

dev/core/test/com/google/gwt/dev/js/JsStaticEvalTest.java
zbynek: I don't think that's generally valid -- if `a()` evaluates to `false` we don't want to evaluate `b()`.

niloc132: No you're right of course, my reply is nonsense.

dev/core/src/com/google/gwt/dev/js/ast/JsBinaryOperation.java
zbynek: > I don't think this is the right place for this 

I wasn't sure, the `JsPrefixOperation` has this, so having it in binary operations seemed consistent. But I didn't want an implementation that checks the whole tree and you're right that what I cam up was only correct under the assumption that some simplifications are already done, which is too fragile. So I now went with your approach of replacing `&&` and `||` with `,` and adding a check for `,` in conditionals.

> The same code that does the shortCircuitOr()/etc check should also be looking for this case

I fixed the code to be symmetric now.

dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
zbynek: Changed to remove both arguments after processing.

niloc132:
Using #10263 to bring this up to date, I show that this brings samples down from 10821121 to 10800491, some 20kb, or about 0.2% - not bad for some light logic improvements!

niloc132:
Taking a little more time to go over this, unless this is exclusively a JSNI problem I think you really want this to be moved/copied to DeadCodeElimination. Doing an optimization on constants like this earlier is extremely likely to benefit the Java optimization loop more than the JS loop (where JsStaticEval lives), as removing flow control is quite likely to also improve inlinability and enhance type tightening or pruning - something that the JS passes are unable to reason about as effectively.

I'm nearly certain the two samples at the end of the description are Java rather than JSNI:
```
a.K==(Slh(),Klh)&&false?(n.g+=TWo,n):tNg(a,n,g)
```
`Slh` is almost certainly a Java `clinit`, which cannot be pruned (tests would need to show that it is promoted), and `n` is probably a stringbuilder/stringbuffer, getting a constant appended to it. DCETest has a few other cases like this that are verified, writing such a test is a little harder to do on a JS AST, since you can't just parse the JS, but need to visit the JS AST and edit it after the fact, and mark them as such - com.google.gwt.dev.js.JsDuplicateFunctionRemoverTest#setAllFromJava has an example of that, and I have a PR to land eventually for #9731 that does this too...

```
O7d(e.yb)&&Tde(e.Qb,e.yb)&&false?Tqe(e):Pqe(e)
```
Harder to be sure about anything here, but I'd bet that Tqe and Pqe are Java instance methods on `e` that were made static. If true, also implies that the java optimization loop is where this would do its best work.

Keeping the change also in the JS loop isn't bad per se, as it will still apply to JSNI or any other work that the Java loop can't optimize out. I could easily imagine that the (apparently uninlinable) pruned instance methods could have other impacts on code size that the JS loop can't necessarily recognize.

dev/core/src/com/google/gwt/dev/jjs/impl/Simplifier.java
```diff
@@ -429,9 +429,10 @@ public static JExpression simplifyAnd(JBinaryOperation expression) {
     if (rhs instanceof JBooleanLiteral) {
       if (((JBooleanLiteral) rhs).getValue()) {
         return lhs;
-      } else if (!lhs.hasSideEffects()) {
-        // Do not remove lhs if it had side effects
-        return rhs;
+      } else  {
+        // if side effect, allow rewriting a() && false && b() -> (a(), false) && b()
+        // -> (a(), false && b()) -> (a(), false)
```
niloc132: I'm wondering if this comment is useful here, since left-associativity means that we would view the whole
a() && false && b()
as
(a() && false) && b()

and so we would be relying on the JMultiExpr instanceof above for step two, and the lhs is the JBooleanLiteral on step three? I think a comment in the test makes sense to spell out that we're going to get all three different optimizations in a single pass, but probably just simplify here to  something like 
```suggestion
        // if side effect, allow rewriting a() && false -> (a(), false)
```

Also technically `(a(),false)` is a character longer than `a()&&false` - perhaps we shouldn't touch it unless we know the current binaryoperation is itself within a binaryoperation?

dev/core/test/com/google/gwt/dev/jjs/impl/DeadCodeEliminationTest.java
```diff
@@ -298,8 +298,10 @@ public void testCommuteMultiExpression() throws Exception {
             + "static int f1;"
             + "static A createA() { A.f1 = 1; return new A(); } "
             + "static boolean booleanWithSideEffects() { createA(); return true;}"
+            + "static boolean randomBooleanWithSideEffects() { createA(); return random() > .5;}"
```
niloc132: We don't usually need to add something like random() to trick the compiler in these tests, since very little of the compiler is run here, just DCE and MethodInliner (requested by this method, line 295) - specializer is still left disabled. It won't even notice that random() fails to return a value...

Given the tests you're adding, you might just prefer to make a new test method and leave the inliner off, so the compiler just lets you call a method and not try to inline it away? Then you could write `a() && false && b()` and ensure you only get `(EntryPoint$A.a(),false)` out, rather than "expect" that the random() and stuff gets inlined in?

Alternatively, JSNI can never be inlined in the Java ast, so you could also just write randomBooleanWithSideEffects() itself as jsni, and it will be assumed to have side effects, not be inlined away.

To be clear, what you have now isn't bad, but it does look like you may have been fighting the test setup a bit to get the test you wanted.

dev/core/test/com/google/gwt/dev/jjs/impl/DeadCodeEliminationTest.java
```diff
@@ -311,6 +313,12 @@ public void testCommuteMultiExpression() throws Exception {
     optimizeExpressions(false, "boolean", "false && A.booleanWithSideEffects()")
         .intoString("return false;");
 
+    optimizeExpressions(false, "int", "A.randomBooleanWithSideEffects() && false && A.randomBooleanWithSideEffects() ? 1 : 2")
+        .intoString("return (EntryPoint$A.f1 = 1, new EntryPoint$A(), (EntryPoint$A.random() > 0.5, 2));");
```
niloc132: Note that it isn't obvious if the first or second randomBooleanWithSideEffects() was inlined here - the code could be wrong and inlining the wrong one, but the test output is the same.

dev/core/test/com/google/gwt/dev/js/JsStaticEvalTest.java
```diff
@@ -154,6 +155,56 @@ public void testLiteralCompares() throws Exception {
...
+  }
+
+  public void testShortCircuitOr() throws Exception {
+    assertEquals("alert(true);", optimize("alert(true || a)"));
+    assertEquals("alert(a);", optimize("alert(false || a)"));
+
+    // these can't be simplified to maintain type
+    assertEquals("alert(a||true);", optimize("alert(a || true)"));
+    assertEquals("alert(a||false);", optimize("alert(a || false)"));
+    assertEquals("alert(!!a||!!b);", optimize("alert(!!a || !!b)"));
```
niloc132, line 202: This last one technically can be simplified further right? `!!(a||b)`

Might also be nice to get a test like the first two, but with `!!`:
```
    assertEquals("alert(!!a);", optimize("alert(!!a || false)"));
    assertEquals("alert(true);", optimize("alert(!!a || true)"));
```

dev/core/test/com/google/gwt/dev/js/JsStaticEvalTest.java
```diff
@@ -154,6 +155,56 @@ public void testLiteralCompares() throws Exception {
...
+    assertEquals("alert(a||b?c:d);", optimize("alert(!!a || !!b ? c :d)"));
+    assertEquals("alert(c);", optimize("alert(true || !!b ? c :d)"));
+    assertEquals("alert(b?c:d);", optimize("alert(false || !!b ? c :d)"));
+    assertEquals("alert(b?c:d1);", optimize("alert(b || false ? c :d1)"));
+    assertEquals("alert(c1);", optimize("alert(b || true ? c1 :d)"));
+    assertEquals("alert(c2);", optimize("alert(a || true || b ? c2 :d)"));
+  }
+
+  public void testShortCircuitOrWithSideEffects() throws Exception {
+    assertEquals("a||b();", optimize("!!a || !!b()"));
```
niloc132, line 215: Question, is this passing because it is in a void context, the value is being thrown away? Is that the goal of the test, or should this be wrapped in something (like alert()) - or should the `a` also be optimized out?

zbynek:
So there are two issues with the boolean expressions
1) extraneous `!!` operators. In the compiled code I checked (~7MB) there were over 5 thousand `!!` operators and more than a half of them by random inspection seemed redundant. These are not JSNI and do not exist in Java code, they are created during translation from Java to JS. So `JsStaticEval` sounds like the right place to fix them (other occurences of `!!` are already eliminated by JsStaticEval. 
2) dead code not eliminated from e.g. `a() && false && b()?c():d()`. In the same codebase I found 36 occurences of `&&false`, 2 of `&&true`, 1 of `||false`, so this is a much smaller problem but some of the eliminated bits might be big. In my last commit I made sure this is handled in DCE. For now I left the equivalent code also in `JsStaticEval`, but if there are concerns about increased complexity of that class, I can remove some of it.

vjay82:
Might there be another reason to use !!, like to convert to boolean or something?

niloc132:
I think the patch mostly looks good, just trying to understand the root issue, be sure we can't do less and get more from it. Nothing here seems like it could increase build times, and for how much we're saving already, it seems obviously worth it. 

I think the `!!` is being added in EqualityNormalizer - all kinds of null/not-null checks are delegated to Cast.isNull/isNotNull, which are just JSNI for `!src`/`!!src` respectively, and JSNI methods are automatically eligible for the JS inlining operation (since java's MethodInliner couldn't touch them). Once inlined, JsStaticEval 400 or so should be handling them - but the issues you've discovered with booleans in the "middle" of &&s and ||s is the root cause in interrupting that?

Concrete example:
https://github.com/gwtproject/gwt/blob/b7117c157bbb3bcfcc022433e307722a8e12fb33/user/super/com/google/gwt/emul/java/util/TreeMap.java#L828-L834
This is inlined into removeWithState:
https://github.com/gwtproject/gwt/blob/b7117c157bbb3bcfcc022433e307722a8e12fb33/user/super/com/google/gwt/emul/java/util/TreeMap.java#L882C12-L882C17
Resulting in roughly
```javascript
    if (!(!!node && node.isRed) && !$isRed(node.child[dir])) {
```
That `!!node` should definitely should be corrected, since it is unambiguously in a boolean context - your addition of `visit(JsBinaryOperation,JsContext)` should by itself correct this, since that will let the endVisit on the outer `!` remove both:
https://github.com/gwtproject/gwt/blob/b7117c157bbb3bcfcc022433e307722a8e12fb33/dev/core/src/com/google/gwt/dev/js/JsStaticEval.java#L399-L418

So I have no doubts about the patch being right, I'm sorry if I've implied that. Adding the Java side to go after `<expr> && false` and friends in the earlier loop will improve that pass, and hopefully make at least some of the JS pass unnecessary - though clearly not the `!!` checks.

I have another review pass about half done, just trying to more deeply understand what's going on here. No draft comments so far on the actual contents of the patch, just the tests and one comment I think could be cleaner.

niloc132:
> Might there be another reason to use !!, like to convert to boolean or something?

Exactly - but in cases where we know it will already be coerced to boolean, the extra `!!` is unnecessary. 

For example 
```
alert(maybeAString || false)
```
will either print the string, or if "falsey", the value `false`. If you want true/false as a string, you would need to do
```
alert(!!maybeAString || false)
```
(this is dumb, since it should just be `alert(!!maybeAString)` - that's what the second set of changes in the patch addresses)

On the other hand
```
alert(maybeAString ? 1 : 2)
```
Here we don't need the `!!` "operator" to coerce to bool, since it can only be used in a boolean context.

dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
```diff
@@ -153,12 +161,33 @@ public boolean visit(JsFunction x, JsContext ctx) {
...
+    }
+
+    public boolean visit(JsBinaryOperation x, JsContext ctx) {
+      if (evalContext.containsKey(x)
+          && (x.getOperator() == JsBinaryOperator.AND || x.getOperator() == JsBinaryOperator.OR)) {
+        evalContext.put(x.getArg1(), EvalMode.BOOL);
+        evalContext.put(x.getArg2(), evalContext.get(x));
+      } else if (x.getOperator() == JsBinaryOperator.COMMA) {
+        evalContext.put(x.getArg1(), EvalMode.VOID);
+        evalContext.put(x.getArg2(), evalContext.get(x));
```
niloc132: This doesn't look right - evalContext.containsKey(x) may or may not be false, so we could be putting `null` for the RHS's context, which means that if we read evalContext.containsKey(RHS) later, we might get "true" when we want "false".

niloc132 (commented):
Sorry, taking longer than I had anticipated with the release going on. 

Oddly, I'm seeing zero impact from this patch in the Java AST of the Showcase sample - I'm still trying to understand why. The original objective of cleaning up `!!`s is going to have to happen only in the JS AST, but the nested `&&`+`?:` examples still appear to be from Java, unless this is again just a case of JSNI method inlining. I'll keep poking, trying to find examples where we could move the work earlier.

---

https://github.com/gwtproject/gwt/blob/b7117c157bbb3bcfcc022433e307722a8e12fb33/user/src/com/google/gwt/user/cellview/client/CellTreeNodeView.java#L150-L169
This code is improved in JS by this patch, in ways that don't seem obvious right away - the COMMA operators are getting cleaned up due to lack of side effects on the _right_ side.

```
     isOpen = $containsKey(openNodes.map, key);
     outerClasses = new StringBuilder(itemStyle);
-    isOpen && (outerClasses.string += openStyle , outerClasses);
-    isRootNode && (outerClasses.string += topStyle , outerClasses);
+    isOpen && (outerClasses.string += openStyle);
+    isRootNode && (outerClasses.string += topStyle);
     isSelected = !!selectionModel && selectionModel.isSelected(value);
     ariaSelected = '' + isSelected;
-    isSelected && (outerClasses.string += selectedStyle , outerClasses);
+    isSelected && (outerClasses.string += selectedStyle);
     innerClasses = new StringBuilder(itemStyle);
     innerClasses.string += itemImageValueStyle;
-    isRootNode && (innerClasses.string += topImageValueStyle , innerClasses);
+    isRootNode && (innerClasses.string += topImageValueStyle);
     isOpen?(image = openImage):model.isLeaf(value)?(image = ($clinit_CellTreeNodeView() , LEAF_IMAGE)):(image = closedImage);
     cellBuilder = new SafeHtmlBuilder;
```
Not all cases of the "return this" are impacted by this, but these improvement sure seem like obvious wins, and occur in many other places in the showcase.

Another example here, where we get a small win, but due to apparent side effects in the clinit, we can't outright remove the whole line:
https://github.com/gwtproject/gwt/blob/b7117c157bbb3bcfcc022433e307722a8e12fb33/user/src/com/google/gwt/user/client/ui/RootPanel.java#L191-L201
```
   if (!!rp) {
     return rp;
   }
-  $size(rootPanels) == 0 && ($clinit_LocaleInfo() , false);
+  $size(rootPanels) == 0 && $clinit_LocaleInfo();
   rp = new RootPanel$DefaultRootPanel;
   $put(rootPanels, null, rp);
```

Here's a case where we probably should be removing the `!!` on the LHS, but instead leave it in place, possibly because this is a VOID context rather than BOOLEAN? Oddly, other cases like this _are_ cleaned up. Maybe the code is too simple here, and being optimized in the Java ast so that we still see this pattern?
https://github.com/gwtproject/gwt/blob/b7117c157bbb3bcfcc022433e307722a8e12fb33/user/src/com/google/gwt/cell/client/ButtonCell.java#L68-L74
```
 _.render = function render(context, data, sb){
   $append(sb.sb, '<button type="button" tabindex="-1">');
-  !!data && ($append(sb.sb, data.html) , sb);
+  !!data && $append(sb.sb, data.html);
   $append(sb.sb, '<\/button>');
 }
```

This next one is probably as optimized as this patch can get it, and shows how the old code was still getting some benefits of removing the dead code, just not as complete as what you worked out
https://github.com/gwtproject/gwt/blob/b7117c157bbb3bcfcc022433e307722a8e12fb33/user/src/com/google/gwt/user/datepicker/client/DatePicker.java#L713-L723
```
 function $refreshAll(this$static){
   $refresh(this$static.view);
   $refresh(this$static.monthAndYearSelector);
-  $isAttached(this$static) && this$static.view.lastDisplayed;
+  $isAttached(this$static);
   $setAriaSelectedCell(this$static.view, this$static.value);
 }
```
It is possible that this patch will open up other avenues for https://github.com/gwtproject/gwt/issues/10242 to benefit the codebase like this one.

This makes me mildly suspicious - is it correct to drop the `|| 0` here? This is a "number" context, right, since we're `|`ing the desired event bits against the existing ones?
https://github.com/gwtproject/gwt/blob/b7117c157bbb3bcfcc022433e307722a8e12fb33/user/src/com/google/gwt/user/client/ui/Widget.java#L239-L245
```
 _.sinkEvents = function sinkEvents(eventBitsToAdd){
-  this.eventsToSink == -1?sinkEvents(($clinit_DOM() , this.element), eventBitsToAdd | ($clinit_DOM() , ($clinit_DOM() , this.element).__eventBits || 0)):(this.eventsToSink |= eventBitsToAdd);
+  this.eventsToSink == -1?sinkEvents(($clinit_DOM() , this.element), eventBitsToAdd | ($clinit_DOM() , ($clinit_DOM() , this.element).__eventBits)):(this.eventsToSink |= eventBitsToAdd);
 }
```
My read is that if you simplify it, this looks like
```
sinkEvents(this.element), eventBitsToAdd | (this.element.__eventBits || 0));
```
and so we want the ||0 as "coerce to number".

dev/core/src/com/google/gwt/dev/jjs/impl/Simplifier.java
zbynek: It may be tricky to know in advance if the outer bracket is needed, if it's not, the `,` expression is shorter.
```
alert(foo()&&false)
alert(foo(),false)
```

niloc132: Agreed - my "technically" responses are mostly to mean "this is _possible_, but probably not easy, and doesnt save us many bytes".

dev/core/test/com/google/gwt/dev/js/JsStaticEvalTest.java
niloc132: EDIT: nope, wrong, ignore

niloc132:
I worked backwards from one of the `<expr> && false` JS cases to see what was causing them, and it seems that your Simplifier changes are at least documented as dealing with this exact case - but at least in my testing, failing to do so.

First, the method that everything is inlined into - ignore everything except the super call at the end
https://github.com/gwtproject/gwt/blob/b7117c157bbb3bcfcc022433e307722a8e12fb33/user/src/com/google/gwt/user/client/ui/DialogBox.java#L517-L529
That super call is PopupPanel.onPreviewNativeEvent:
https://github.com/gwtproject/gwt/blob/b7117c157bbb3bcfcc022433e307722a8e12fb33/user/src/com/google/gwt/user/client/ui/PopupPanel.java#L1046-L1052
which in turn calls onEventPreview (as noted, this is deprecated, so just has a default impl):
https://github.com/gwtproject/gwt/blob/b7117c157bbb3bcfcc022433e307722a8e12fb33/user/src/com/google/gwt/user/client/ui/PopupPanel.java#L704-L707
The `!onEventPreview` naturally will be inlined as just `!true` and at the end of the Java optimization loop, these are fully inlined, though we keep the `event.nativeEvent` accessor for some reason, but `!true` is now `false`
```
  protected final void onPreviewNativeEvent(Event$NativePreviewEvent event){
    final NativeEvent nativeEvent = event.nativeEvent;
    !event.isCanceled && Event.$getTypeInt((Event) event.nativeEvent) == 4 && DialogBox.$isCaptionEvent(this, nativeEvent) && DOMImplStandard.$eventPreventDefault(nativeEvent);
    event.isFirstHandler && (((Event) event.nativeEvent, false)) && (event.isCanceled = true);
  }
```

First iteration optimization loop with JS (before this patch's changes):
```
_.onPreviewNativeEvent = function onPreviewNativeEvent(event){
  var nativeEvent;
  nativeEvent = event.nativeEvent;
  !event.isCanceled && $getTypeInt(event.nativeEvent) == 4 && $isCaptionEvent(this, nativeEvent) && (nativeEvent.preventDefault() , undefined);
  event.isFirstHandler && false && (event.isCanceled = true);
}
;
```

The JS pass got the `event.nativeEvent` removal right, and then your patch correctly rewrites away the rest of the code in that line (event.isFirstHandler has no side effects, and the assignment will never be called). BUT, if Java correctly cleaned up that multiexpression (which your current DCE changes are attempting to do?). The issue is the JCastOperation - except when casting to primitive, these always are marked as having a side effect (since there could be a CCE thrown). 
https://github.com/gwtproject/gwt/blob/b7117c157bbb3bcfcc022433e307722a8e12fb33/dev/core/src/com/google/gwt/dev/jjs/ast/JCastOperation.java#L53-L63
At the same time, Event.as() is marked as "this is always safe"
https://github.com/gwtproject/gwt/blob/b7117c157bbb3bcfcc022433e307722a8e12fb33/user/src/com/google/gwt/user/client/Event.java#L464-L471
It turns out that the compiler will make the cast be free... because both Event and NativeEvent are JSOs. Event is a subclass of NativeEvent, so TypeTightener doesnt believe it is automatically safe, but really, since it is a JSO, it shouldn't check, it should just make it a free operation. 

As for why it is a free cast once we're in JS:
First, casting a JSO to a JSO is ignored by the visitor that rewrites casts to Cast.java method calls
https://github.com/gwtproject/gwt/blob/b7117c157bbb3bcfcc022433e307722a8e12fb33/dev/core/src/com/google/gwt/dev/jjs/impl/ImplementCastsAndTypeChecks.java#L100-L104
Then, when we translate the Java AST to JS, we just ignore that leftover cast:
https://github.com/gwtproject/gwt/blob/b7117c157bbb3bcfcc022433e307722a8e12fb33/dev/core/src/com/google/gwt/dev/jjs/impl/GenerateJavaScriptAST.java#L605-L608

---

So for this single case, my conclusion is that it is a bug for GWT to ignore the case during JS AST generation, when it could just as easily have tightened the cast away earlier and made all of this unnecessary.

So at least for this specific example, better casting logic would have prevented this

dev/core/src/com/google/gwt/dev/jjs/impl/Simplifier.java
zbynek: I can reproduce the issue with `||0` in a test, checking...

zbynek:
From GGB codebase

```
 $isUnbundledOrWhiteboard(this$static.cons.kernel.app)?this$static.cons.kernel.app.isUnbundledGeometry() && false?(locus.lineOpacity = 204):this$static.cons.kernel.app.isUnbundledGraphing() && false?(locus.lineOpacity = 204):this$static.cons.kernel.app.isUnbundledGraphing() && false?(locus.lineOpacity = 204):(locus.lineOpacity = 178):(locus.lineOpacity = 178);
 ```
 which comes from `setDefaultLineStyle(locus);`
 
 inlining the two methods
 
 ```
 private void setDefaultLineStyle(GeoElement geo) {
		if (geo instanceof GeoAngle
				&& cons.getApplication().isUnbundledGeometry()) {
			geo.setLineThickness(
					EuclidianStyleConstants.OBJSTYLE_DEFAULT_LINE_THICKNESS_ANGLE_GEOMETRY);
		} else {
			geo.setLineThickness(
					EuclidianStyleConstants.OBJSTYLE_DEFAULT_LINE_THICKNESS);
		}
		if (geo.hasLineOpacity()) {
			if (cons.getApplication().isUnbundledOrWhiteboard()) {
				setLineOpacity(geo);
			} else {
				geo.setLineOpacity(
						EuclidianStyleConstants.OBJSTYLE_DEFAULT_LINE_OPACITY);
			}
		}
	}

	private void setLineOpacity(GeoElement geo) {
		if (geo instanceof GeoAngle) {
			geo.setLineOpacity(
					EuclidianStyleConstants.OBJSTYLE_DEFAULT_LINE_OPACITY_ANGLE);
		} else if (geo instanceof GeoPolygon) {
			geo.setLineOpacity(
					EuclidianStyleConstants.OBJSTYLE_DEFAULT_LINE_OPACITY_POLYGON);
		} else if (geo instanceof GeoConicPart) {
			geo.setLineOpacity(
					EuclidianStyleConstants.OBJSTYLE_DEFAULT_LINE_OPACITY_SECTOR);
		} else if (cons.getApplication().isUnbundledGeometry()
				&& (geo instanceof GeoLine || geo instanceof GeoSegment
						|| geo instanceof GeoRay || geo instanceof GeoVector
						|| geo instanceof GeoPolyLine
						|| geo instanceof GeoConic)) {
			geo.setLineOpacity(
					EuclidianStyleConstants.OBJSTYLE_DEFAULT_LINE_OPACITY_GEOMETRY);
		} else if (cons.getApplication().isUnbundledGraphing()
				&& (geo instanceof GeoFunction)) {
			geo.setLineOpacity(
					EuclidianStyleConstants.OBJSTYLE_DEFAULT_LINE_OPACITY_FUNCTION_GEOMETRY);
		} else if (cons.getApplication().isUnbundledGraphing()
				&& (geo instanceof GeoCurveCartesian)) {
			geo.setLineOpacity(
					EuclidianStyleConstants.OBJSTYLE_DEFAULT_LINE_OPACITY_CURVE_GEOMETRY);
		} else {
			geo.setLineOpacity(
					EuclidianStyleConstants.OBJSTYLE_DEFAULT_LINE_OPACITY);
		}

	}
```

The `instanceof` checks can be evaluated to `false` at compile time, creating the dead branches in the conditional operator.

dev/core/test/com/google/gwt/dev/js/JsStaticEvalTest.java
zbynek: Yes, I wanted to test that in void context `!!` is removed even with side effects.
Here we can't optimize `a` out because it decides if `b()` is or is not called.
We can add some tests with `alert` and side effects.

niloc132:
Picking the inner-most ternary's condition to preserve my sanity, from the
```
this$static.cons.kernel.app.isUnbundledGeometry() && false
```
expression, I'm guessing that we really do have side effects, and this is the "(a(),false)" vs "a()&&false" difference, still preserving side effects - the former is easier to potentially optimize out, but the latter is shorter. As it happens, Simplifier.java explicitly calls this case out:
```
    * isWhatever() && false -> false, unless side effects
```

That above is embedded in a conditional if I'm reading it right, applying the multiexpr transformation we would have (what your patch now does):
```
(a(),false)?(locus.lineOpacity = 204):(locus.lineOpacity = 178)
```
Now we're in Simplifier.simplifyConditional(), and this case should cover it:
```
   * (a,b,c)?d:e -> a,b,(c?d:e)
```
so that we end up with
```
a(),(false?(locus.lineOpacity = 204):(locus.lineOpacity = 178))
```
When simplifyConditional does that, it recursively calls itself, so would collapse that `false` right away, and just leave
```
a(),locus.lineOpacity = 178;
```
Does that work with the current version of the patch? At a glance, these should apply in the correct order to take place in a single pass of DCE?

As an aside, its too bad that the 
```
      // Special case
      // if () { return ..; } else { return ..; } =>
      // return ... ? ... : ...;
```
optimization doesn't work on assignment - it would be an easy win for your code (and lots of other code) to add a case like
```
// if () { a=.. } else { a=.. } -> a = (... ? .. : ..) ;
```

niloc132:
I wrote a quick test that seems to confirm that the last step at least works as expected:
```
    runMethodInliner = true;
    addSnippetClassDecl(
        "static class A { "
            + "static int f1;"
            + "public static boolean sideEffect() { f1++; return true; } "
        + "}");

    optimize("boolean", "return A.sideEffect() ? b : b1;").intoString(
        "return (++EntryPoint$A.f1, EntryPoint.b);");
```

niloc132:
https://github.com/gwtproject/gwt/issues/9731's use case potentially is also improved by this - effectively we have `(a(), b(), false) ? c() : d()`, and the nested multiexpressions (err, JsBinExpr with COMMA) aren't being rewritten out. 

The `false` is confusing at this time for the Java AST because it isn't a literal `false`, its a Boolean object with value `false`. In JS-land, that's the same thing, so we can optimize it here, and hopefully this pattern of nested commas is something we can handle now (or can add on, or can follow up with).

dev/core/src/com/google/gwt/dev/jjs/impl/Simplifier.java
zbynek: This was fixed by not putting `null`s into the map.

dev/core/test/com/google/gwt/dev/jjs/impl/DeadCodeEliminationTest.java
zbynek: Moved to #10279

dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
```diff
@@ -144,21 +145,53 @@ public boolean visit(JsFunction x, JsContext ctx) {
...
+    BOOL, // result will be coerced to a boolean
+    VOID // result will be discarded
+  }
+
   /**
    * Does static evals.
    *
    * TODO: borrow more concepts from
    * {@link com.google.gwt.dev.jjs.impl.DeadCodeElimination}, such as ignored
    * expression results.
```
niloc132, line 163: We're quite close to throwing away ignored expression results here. Probably not worth adding a few more VOID checks at this stage

dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
```diff
@@ -607,41 +657,75 @@ public static int exec(JsProgram program) {
...
     if (arg1 instanceof CanBooleanEval) {
       CanBooleanEval eval1 = (CanBooleanEval) arg1;
-      if (eval1.isBooleanTrue()) {
+      if (eval1.isBooleanFalse()) {
+        return withSideEffect(arg2, arg1, expr.getSourceInfo());
+      } else if (eval1.isBooleanTrue()) {
+        return arg1;
+      }
+    }
+    if (arg2 instanceof CanBooleanEval && evalContext.containsKey(expr)) {
```
niloc132: should this containsKey(expr) be a get(expr) == BOOL? It might have been missed in the last update, but seems implied by the javadoc.

If it is VOID, I would assume we actually throw away arg2... but maybe a later patch (see my comment on the TODO for this nested type).

dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
```diff
@@ -607,41 +657,75 @@ public static int exec(JsProgram program) {
...
       CanBooleanEval eval1 = (CanBooleanEval) arg1;
-      if (eval1.isBooleanTrue() && !arg1.hasSideEffects()) {
-        return arg2;
+      if (eval1.isBooleanTrue()) {
+        return withSideEffect(arg2, arg1, expr.getSourceInfo());
       } else if (eval1.isBooleanFalse()) {
         return arg1;
       }
     }
+    if (arg2 instanceof CanBooleanEval && evalContext.containsKey(expr)) {
```
niloc132: Same concern, just a containsKey, so it could be explicitly VOID.

niloc132 (commented):
There are a few other possibly suspect containsKey calls - but it also might be that "VOID" still _permits_ the optimizations being made, even if removal would be better. I'm okay putting off removal, just want to be sure code/comments are clearer.

dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
zbynek: Any simplification that's correct in `BOOL` context is also valid in `VOID` context, so I'm not concertned about correctness.

> we actually throw away arg2

That's what line 203 should do. Because of that we can really just care about boolean context in the short circuit optimizations, I comments to the code.

dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
```diff
@@ -607,41 +659,79 @@ public static int exec(JsProgram program) {
...
+      if (eval2.isBooleanTrue() && !arg2.hasSideEffects()) {
+        return arg1;
+      } else if (eval2.isBooleanFalse()) {
+        return withSideEffect(arg2, arg1, expr.getSourceInfo());
+      }
+    }
     return expr;
   }
 
+  private static JsExpression withSideEffect(JsExpression main, JsExpression sideEffect,
```
niloc132: I'm pretty sure there's no bug here, but I think I would encourage the arg order to be swapped, so that we have `sideEffect` before `main`. This accurately shows the order that they will execute in, if `sideEffect` has side effects. 

All callers currently swap order, so we see a lot of `withSideEffect(arg2, arg1, expr.getSourceInfo())`, listing the calls backwards so that they are executed forwards.

Javadoc might be nice too, to state this assumption clearly, whether or not you decide to change the order.

dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
```diff
@@ -144,21 +145,55 @@ public boolean visit(JsFunction x, JsContext ctx) {
...
+     * Stores how are expression evaluations used.
+     * Missing entry = no coercion, difference between null and false matters.
+     */
+    private final Map<JsExpression, EvalMode> evalContext = new HashMap<>();
 
     /**
      * This is used by {@link #additionCoercesToString}.
      */
-    private Map<JsExpression, Boolean> coercesToStringMap = new IdentityHashMap<JsExpression, Boolean>();
+    private Map<JsExpression, Boolean> coercesToStringMap = new IdentityHashMap<>();
```
niloc132: If this is a IdentityHashMap, `evalContext` probably also should be one.

Also, this should be final.

niloc132: And for that matter, is only read/written from additionCoercesToString, which is unused.

JsExpression doesn't offer equals/hashCode so either map type is fine, just would like to be consistent.. and since we're probably better off removing this field, we can be consistent that way.

dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
```diff
@@ -607,41 +659,79 @@ public static int exec(JsProgram program) {
    * Simplify short circuit AND expressions.
    *
    * <pre>
-   * if (true && isWhatever()) -> if (isWhatever()), unless side effects
-   * if (false() && isWhatever()) -> if (false())
+   * return true() && isWhatever() -> return isWhatever(), unless true() has side effects
+   * return false() && isWhatever() -> return false()
+   * </pre>
+   * In boolean context also
```
niloc132: remove/complete this thought?

dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
zbynek: I wasn't sure whether to delete this since it's unrelated and might be somehow useful if fixed, but maybe better to remove it now and mention in  comment that expressions coerced to string might be optimized.

niloc132 (approved):
Re-reviewed the diff of compiled showcase output - all looks good.

This is the only change that wasn't a net decrease, but also not an increase.
```patch
-  this$static.focusable = ($clinit_FocusPanel() , createFocusable0(($clinit_FocusImpl() , focusHandler)?focusHandler:(focusHandler = $createFocusHandler())));
+  this$static.focusable = ($clinit_FocusPanel() , createFocusable0(($clinit_FocusImpl() , focusHandler?focusHandler:(focusHandler = $createFocusHandler()))));
```

Everything else was removing unnecessary commas or `!!` in contexts where the value would already be coerced to boolean.

I also briefly audited the generated JS, looking specifically for extra `!!`s that could be removed, and didn't spot any of note.

PR #10266, opened 2026-02-01
Update poms to pass validation for Maven Central sync

niloc132:
The new Sonatype publishing mechanism requires a different process to deploy, and seems to have stricter requirements. Removed unnecessary parent poms, added developers, scm, description, and re-grouped bom dependencies to avoid confusion from `com.google.gwt` dependencies.

Fixes #10264

### DIFF
--- a/maven/poms/gwt/pom-template.xml
+++ b/maven/poms/gwt/pom-template.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0"
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
     <groupId>org.gwtproject</groupId>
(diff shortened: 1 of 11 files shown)


zbynek:
nit: If we're cleaning this up, maybe also change 
http://maven.apache.org/maven-v4_0_0.xsd -> https://maven.apache.org/maven-v4_0_0.xsd

maven/poms/gwt/pom-template.xml
```diff
@@ -2,65 +2,54 @@
...
-    <url>http://www.gwtproject.org/</url>
+    <name>GWT Web Toolkit</name>
+    <description>BOM for GWT</description>
+    <url>https://www.gwtproject.org/</url>
     <version>${gwtVersion}</version>
     <licenses>
         <license>
             <name>GWT Terms</name>
-            <url>http://www.gwtproject.org/terms.html</url>
+            <comments>GWT itself is Apache 2.0 licensed, but contains some elements with different licenses</comments>
```
tbroyer, line 16: As long as they're not bundled in the JARs, I'd say the POM should say Apache-2.0. If some are repackaged in, then GWT Terms with this description is good.

niloc132: gwt-dev has protobuf (MIT) and rhino (MPL) bundled in. gwt-user has flute bundled (w3c software license). I don't immediately see anything specific in gwt-servlet or the rf jars to mention here, but I think the above at least justifies the arm waving about checking the site - at least for now?

niloc132:
@zbynek done

zbynek (approved):

PR #10270, opened 2026-02-04
Normalize if/else blocks

niloc132:
Fixes a longstanding TODO, and makes it easier for visitors to rewrite contents of blocks in if statements. Originally this change was intended to have no impact on generated JS, but resulted in a dangling-else problem, so changes were required to emit blocks when an if has an else.

Several tests needed to be updated to no longer assume that an unnecessary block was present after parsing. 

Fix #10239


niloc132:
As of this comment, samples are at 10807627 from 10821121, 13494 bytes saved, about 0.1% smaller.

I'm skeptical about making the new method on JBlock - though it should be used for for/while/do-while blocks as well (though not try blocks). Probably should resolve that one way or the other before merge.

The change to printing blocks in JS when an if has an else are more aggressive than they need to be - this need not happen unless the if has a child if, so we could optimize very slightly more here. Options here include keeping nested blocks for ifs manual and introducing a normalization pass rather than a feature of the JsToStringGenerator, or adding a visitor when printing the JS to see if the block is needed.

Java 17 change was done to ease test writing, and probably should be done anyway as part of the 2.14 release process.

#10240 was also addressed then reverted in this branch, to follow in a later PR.

dev/core/test/com/google/gwt/dev/js/JsToStringGenerationVisitorTest.java
```diff
@@ -75,6 +82,65 @@ public void testClassRangeMarking() throws UnableToCompleteException {
...
+      return text.toString().substring(r.getStartPosition(), r.getEndPosition());
+    }).get();
+
+    // Parse the source to be sure that the printed output has the expected characteristics
+    List<JsStatement> parsed = JsParser.parse(SourceOrigin.UNKNOWN, jsProgram.getScope(), new StringReader(entrypoint));
+
+    JsStatement result = null;
+    for (JsStatement jsStatement : parsed) {
+      if (jsStatement instanceof JsExprStmt && ((JsExprStmt) jsStatement).getExpression() instanceof JsFunction) {
+        JsFunction jsFunction = (JsFunction) ((JsExprStmt) jsStatement).getExpression();
```
zbynek:
```suggestion
      if (jsStatement instanceof JsExprStmt exprStmt 
          && exprStmt.getExpression() instanceof JsFunction jsFunction) {
```
If we're using Java 17 here, we can simplify the casts a bit.

niloc132: Agreed - the change to 17 was mostly for tests, but it could benefit the code in other places too

dev/core/src/com/google/gwt/dev/jjs/ast/JForStatement.java
```diff
@@ -81,9 +81,7 @@ public void traverse(JVisitor visitor, Context ctx) {
       if (increments != null) {
         increments = visitor.accept(increments);
       }
-      if (body != null) {
-        body = visitor.accept(body, true);
-      }
+      body = JBlock.ensureBlock(getSourceInfo(), visitor.accept(body, false));//TODO no tests fail without this change...
```
niloc132, line 84: Remove TODO after verifying with a test.

dev/core/src/com/google/gwt/dev/js/JsToStringGenerationVisitor.java
```diff
@@ -561,6 +561,11 @@ public boolean visit(JsIf x, JsContext ctx) {
     accept(x.getIfExpr());
     _rparen();
     JsStatement thenStmt = x.getThenStmt();
+    if (!(thenStmt instanceof JsBlock) && x.getElseStmt() != null) {
```
niloc132: This should have a comment that we're testing for dangling else and defensively adding {}s to the block.

This is technically not optimal - if there is no possibility of a nested `if` in the "then" branch, we don't need to do this. The simplest improvement here then would be to allow certain statement types that can never contain an if, like JsEmpty, JsThrow, JsContinue, JsExprStmt, etc, or even some subset of those. As it stands, this can produce a small regression here, such as `if(condition){;}else...` when the "then" branch is simply a JsEmpty.

dev/core/src/com/google/gwt/dev/jjs/ast/JBlock.java
```diff
@@ -105,4 +116,18 @@ public boolean unconditionalControlBreak() {
...
+    }
+    if (statements.size() == 1) {
+      JStatement jStatement = statements.get(0);
+      if (jStatement instanceof JBlock) {
+        return ((JBlock) jStatement).singleStatement();
+      }
+      return jStatement;
+    }
+    return this;
+  }
```
vegegoku, line 146: In terms of readabilty this can be enhanced, maybe like this as we put the exit conditions in order

```java
  int statementCount = statements.size();

  if (statementCount == 0) {
    return null;
  }

  if (statementCount > 1) {
    return this;
  }

  JStatement statement = statements.get(0);
  return statement instanceof JBlock
      ? ((JBlock) statement).singleStatement()
      : statement;
```
this order reads better since the exit condition shows up on sequence.

dev/core/src/com/google/gwt/dev/jjs/ast/JWhileStatement.java
```diff
@@ -43,9 +43,7 @@ public JExpression getTestExpr() {
   public void traverse(JVisitor visitor, Context ctx) {
     if (visitor.visit(this, ctx)) {
       testExpr = visitor.accept(testExpr);
-      if (body != null) {
-        body = visitor.accept(body, true);
-      }
+      body = JBlock.ensureBlock(getSourceInfo(), visitor.accept(body, false));//TODO no tests fail without this change...
```
vegegoku, line 46: Though you added this todo, it is a little big ambiguous for anyone who might read this later.

dev/core/src/com/google/gwt/dev/jjs/ast/JBlock.java
niloc132: You're right - and I do love the irony that staring at all these if/else block optimizations makes me hate seeing separate if blocks to return early (but the entire point of the patch is to let developers write what reads well, and be sure it will optimize the same).

dev/core/src/com/google/gwt/dev/jjs/ast/JWhileStatement.java
niloc132: Yeah I added a PR comment on the one in JForStatement - all of them need to be cleaned up once I have a specific test for it one way or the other.

dev/core/src/com/google/gwt/dev/jjs/ast/JBlock.java
niloc132: My one thing that I dislike about your ordering is that I was trying to handle the cases by number:

 - `0`, return null
 - `1`, if its a block, descend into the block, otherwise return the statement
 - `>1`, return this

Unless that convinces you, I'll go with your ordering, but not mixing a ternary with ifs?

dev/core/src/com/google/gwt/dev/js/JsToStringGenerationVisitor.java
niloc132: (Second try at this, github ate the first when I dared to unresolve before posting)

I built this out to be more rigorous, we should always optimally omit {}s except when there could be a nested `if` with no `else`. Doesn't remove unnecessary blocks from ifs though.

If we end up with more post-optimization normalizations, we might want to group them up outside of the string generator for readability, though this isn't the first we've added here (true/false to !0/!1, etc). Moving these outside of the string generator also would have the disadvantage of not applying the change when printing an AST node to a string for debugging or viewing an AST dump.

See also #10382

dev/core/src/com/google/gwt/dev/jjs/ast/JBlock.java
jnehlmeier: Just use `statementCount == 1` in the second if and move the block code to it. Last line in the method is then just `return this`?

vegegoku: That is convincing enough.

vegegoku (approved):

PR #10273, opened 2026-02-07
Permit failure to delete files on Windows, and close finished logs

niloc132:
Audited all other usages of `PrintWriterTreeLogger`, and found this was only one of two usages that writes to a file rather than System.out/err or a StringWriter. The other usage writes to a file in the old devmode flag `-logdir`, which is optional, and expected not to be within another directory that the compiler needs to clean.

Fixes #10272

### DIFF
--- a/dev/codeserver/java/com/google/gwt/dev/codeserver/OutboxDir.java
+++ b/dev/codeserver/java/com/google/gwt/dev/codeserver/OutboxDir.java
@@ -88,9 +88,11 @@ class OutboxDir {
     // (This is not guaranteed to delete all directories on Windows if a directory is locked.)
     for (File candidate : children) {
       if (candidate.getName().startsWith(COMPILE_DIR_PREFIX)) {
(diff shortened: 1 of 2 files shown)


niloc132:
Patch is best reviewed with "hide whitespace" https://github.com/gwtproject/gwt/pull/10273/changes?w=1

niloc132:
@keinhaar are you able to build and validate this works for you before we merge? Whatever it is about your machine/env, you're the only Windows user so far to test and see this issue.

Prebuilt SDK is available through the action results, see the links at the bottom of https://github.com/gwtproject/gwt/actions/runs/21780568450?pr=10273

keinhaar:
I tried the prebuild version gwt-java17.zip.
It work's as expected.
Thanks!

niloc132:
Since we're just about done with the 2.13.0 release process and still only the one report of this, my opinion is that we should just going to finish that and then kick off another one immediately, with this fix and any other bugs found in the wild. 

I'll make sure release notes mention and link here to avoid surprises by teams that run windows.

vegegoku (approved):

PR #10279, opened 2026-02-20
Simplify boolean expressions in Java AST

### LINKED ISSUE #10278: Improve simplification of boolean expressions in Java AST
On the Java level (in particular in `DeadCodeElimination` class) boolean expressions like `false && foo()` are simplified, but expressions like `foo() && false` are kept as-is. By changing `foo() && false` into `(foo(),false)` some more complicated expressions could be simplified further, e.g.

`foo() && false ? a() : b()` can become `(foo(), b())`

(the code for simplifying ternary operators with `,` in the condition already exists)

This issue is based  on https://github.com/gwtproject/gwt/pull/10261#discussion_r2747269907

zbynek:
Split from #10261 (see comments in that PR for discussion).

Fixes #10278

### DIFF
--- a/dev/core/src/com/google/gwt/dev/jjs/impl/Simplifier.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/impl/Simplifier.java
@@ -429,9 +429,10 @@ public class Simplifier {
     if (rhs instanceof JBooleanLiteral) {
       if (((JBooleanLiteral) rhs).getValue()) {
         return lhs;
-      } else if (!lhs.hasSideEffects()) {
-        // Do not remove lhs if it had side effects
-        return rhs;
+      } else  {
+        // if side effect, allow rewriting a() && false && b() -> (a(), false) && b()
+        // -> (a(), false && b()) -> (a(), false)
+        return lhs.hasSideEffects() ? new JMultiExpression(info, lhs, rhs) : rhs;
       }
     }
 
@@ -482,8 +483,8 @@ public class Simplifier {
     if (rhs instanceof JBooleanLiteral) {
       if (!((JBooleanLiteral) rhs).getValue()) {
         return lhs;
-      } else if (!lhs.hasSideEffects()) {
-        return rhs;
+      } else {
+        return lhs.hasSideEffects() ? new JMultiExpression(info, lhs, rhs) : rhs;
       }
     }
     return expression;
--- a/dev/core/src/com/google/gwt/dev/jjs/impl/Simplifier.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/impl/Simplifier.java
@@ -429,10 +429,14 @@ public class Simplifier {
     if (rhs instanceof JBooleanLiteral) {
       if (((JBooleanLiteral) rhs).getValue()) {
         return lhs;
-      } else  {
-        // if side effect, allow rewriting a() && false && b() -> (a(), false) && b()
+      } else if (lhs.hasSideEffects()) {
+        // if side effect, rewrite a() && false to (a(), false)
+        // so that parent node can be simplified further
+        // e.g. (a() && false) && b() -> (a(), false) && b()
         // -> (a(), false && b()) -> (a(), false)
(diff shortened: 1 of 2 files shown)


dev/core/src/com/google/gwt/dev/jjs/impl/Simplifier.java
```diff
@@ -429,9 +429,10 @@ public static JExpression simplifyAnd(JBinaryOperation expression) {
     if (rhs instanceof JBooleanLiteral) {
       if (((JBooleanLiteral) rhs).getValue()) {
         return lhs;
-      } else if (!lhs.hasSideEffects()) {
-        // Do not remove lhs if it had side effects
-        return rhs;
+      } else  {
+        // if side effect, allow rewriting a() && false && b() -> (a(), false) && b()
+        // -> (a(), false && b()) -> (a(), false)
+        return lhs.hasSideEffects() ? new JMultiExpression(info, lhs, rhs) : rhs;
```
niloc132: to be clear, this is really just `a() && false` to `(a(), false)` right? Then on the way out we would handle the next stage (and test stuff address that so we know this change composes well with existing `&&`, `?:` optimizations in DCE

For my part, I think it would read better if we had another else/if in here rather than a ternary within an else


```suggestion
      } else if (!lhs.hasSideEffects()) {
        return rhs;
      } else {
        // if side effect, allow rewriting a() && false -> (a(), false)
        // which will enable other optimizations to eval them separately
        return new JMultiExpression(info, lhs, rhs);
```

dev/core/test/com/google/gwt/dev/jjs/impl/DeadCodeEliminationTest.java
```diff
@@ -298,8 +298,10 @@ public void testCommuteMultiExpression() throws Exception {
             + "static int f1;"
             + "static A createA() { A.f1 = 1; return new A(); } "
             + "static boolean booleanWithSideEffects() { createA(); return true;}"
+            + "static boolean randomBooleanWithSideEffects() { createA(); return random() > .5;}"
             + "static boolean booleanWithoutSideEffects() { return true;}"
             + "static int arithmeticWithSideEffects() { createA(); return 4;}"
+            + "static native double random() /*-{ }-*/;"
```
niloc132: Discussed in the other PR, but this can be simpler if you split to another test, or turn off runMethodInliner before your specific tests. All methods are assumed to have side effects by default in GWT, but once they are inlined away they might no longer have side effects. You can also cheat your way out of this not with JSNI, but just a `@DoNotInline`. It makes the tests easier to read/write if you don't have to mess with this stuff

dev/core/test/com/google/gwt/dev/jjs/impl/DeadCodeEliminationTest.java
```diff
@@ -311,6 +313,12 @@ public void testCommuteMultiExpression() throws Exception {
     optimizeExpressions(false, "boolean", "false && A.booleanWithSideEffects()")
         .intoString("return false;");
 
+    optimizeExpressions(false, "int", "A.randomBooleanWithSideEffects() && false && A.randomBooleanWithSideEffects() ? 1 : 2")
```
niloc132: Add a test for just plain `&&` and `||` that fails before this patch - the more complex test is great to have as well, but we should test the simpler case.

zbynek:
Stats for GeoGebra main chunk

before:
pretty:21671810
obfuscated:7864283
gzipped:2554348

after:
pretty:21670169
obfuscated:7863399
gzipped:2553679

So this seems to be just .03% improvement.

niloc132:
> So this seems to be just .03% improvement.

Right, most of these are going to be tiny on their own, or will depend on the specific code being compiled repeating this pattern.

I still think changes like this in pattern matching simple code is worth it - we're adding a single `if` in an edge case, and DCE is specifically designed to converge in a single pass (though other optimizations can interact with it and require another later pass). The theory is that if we find other code later that can be rewritten trivially to `false`/`true`, we will get to propagate that out and clean up other code - and now that will even work if the boolean is trapped in a multiexpr.

niloc132 (approved):

PR #10291, opened 2026-03-04
Minify boolean and numeric literals in obfuscated mode

zbynek:
Fixes #10281

Also cleans up some minor code style issues in the modified files.

Does NOT implement
* Infinity -> 1/0
* undefined -> void 0
* 1.2E100 -> 12E99

### DIFF
--- a/dev/core/src/com/google/gwt/dev/jjs/JavaToJavaScriptCompiler.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/JavaToJavaScriptCompiler.java
@@ -1482,7 +1482,6 @@ public final class JavaToJavaScriptCompiler {
 
         nodeCount = jprogram.getNodeCount();
         mods = stats.getNumMods();
-        logger.log(TreeLogger.Type.TRACE, "Optimization pass " + passCount + " of " + passLimit + "\n" + stats);
       }
 
       float nodeChangeRate = mods / (float) lastNodeCount;
--- a/dev/core/src/com/google/gwt/dev/jjs/JavaToJavaScriptCompiler.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/JavaToJavaScriptCompiler.java
@@ -151,6 +151,7 @@ import com.google.gwt.dev.js.JsReportGenerationVisitor;
 import com.google.gwt.dev.js.JsStackEmulator;
 import com.google.gwt.dev.js.JsStaticEval;
 import com.google.gwt.dev.js.JsSymbolResolver;
+import com.google.gwt.dev.js.JsToStringGenerationVisitor;
 import com.google.gwt.dev.js.JsUnusedFunctionRemover;
 import com.google.gwt.dev.js.JsVerboseNamer;
 import com.google.gwt.dev.js.SizeBreakdown;
@@ -772,7 +773,9 @@ public final class JavaToJavaScriptCompiler {
           DefaultTextOutput out = new DefaultTextOutput(!options.isIncrementalCompileEnabled() &&
               options.getOutput().shouldMinimize());
           JsReportGenerationVisitor v = new JsReportGenerationVisitor(out, jjsMap,
-              options.isJsonSoycEnabled());
+              options.isJsonSoycEnabled(),
+              new JsToStringGenerationVisitor.PrintOptions(false,
+                  options.getOutput() == JsOutputOption.OBFUSCATED));
           v.accept(jsProgram.getFragmentBlock(i));
 
           StatementRanges statementRanges = v.getStatementRanges();
@@ -1479,6 +1482,7 @@ public final class JavaToJavaScriptCompiler {
 
         nodeCount = jprogram.getNodeCount();
         mods = stats.getNumMods();
+        logger.log(TreeLogger.Type.TRACE, "Optimization pass " + passCount + " of " + passLimit + "\n" + stats);
       }
 
       float nodeChangeRate = mods / (float) lastNodeCount;
--- a/dev/core/src/com/google/gwt/dev/js/JsToStringGenerationVisitor.java
+++ b/dev/core/src/com/google/gwt/dev/js/JsToStringGenerationVisitor.java
@@ -1318,14 +1318,13 @@ public class JsToStringGenerationVisitor extends JsVisitor {
   }
 
   private void _printReturnExpression(JsExpression arg) {
-    boolean space = false;
-    String value = null;
+    boolean space = true;
     if (arg instanceof JsBooleanLiteral) {
       space = !minifyLiterals;
     } else if (arg instanceof JsPrefixOperation) {
      space = ((JsPrefixOperation) arg).getOperator().isKeyword();
     } else if (arg instanceof JsNumberLiteral) {
-      value = _stringifyNumber((JsNumberLiteral) arg);
+      String value = _stringifyNumber((JsNumberLiteral) arg);
       space = value.charAt(0) != '-' && value.charAt(0) != '.';
     }
     if (space) {
@@ -1333,11 +1332,7 @@ public class JsToStringGenerationVisitor extends JsVisitor {
     } else {
       _spaceOpt();
     }
-    if (value != null) {
-      p.print(value);
-    } else {
-      accept(arg);
-    }
+    accept(arg); // this may serialize numbers again, but needed for billing
   }
 
   private void _spaceOpt() {
--- a/dev/core/src/com/google/gwt/dev/js/JsToStringGenerationVisitor.java
+++ b/dev/core/src/com/google/gwt/dev/js/JsToStringGenerationVisitor.java
@@ -239,10 +239,6 @@ public class JsToStringGenerationVisitor extends JsVisitor {
 
   @Override
   public boolean visit(JsBooleanLiteral x, JsContext ctx) {
-    if (minifyLiterals) {
-      p.print(x.getValue() ? "!0" : "!1");
-      return false;
-    }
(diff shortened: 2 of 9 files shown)


dev/core/src/com/google/gwt/dev/js/JsToStringGenerationVisitor.java
```diff
@@ -227,6 +238,10 @@ public boolean visit(JsBlock x, JsContext ctx) {
 
   @Override
   public boolean visit(JsBooleanLiteral x, JsContext ctx) {
+    if (minifyLiterals) {
+      p.print(x.getValue() ? "!1" : "!0");
```
niloc132: The `_true()`/`_false()` methods are intended to use cached `char[]` and avoid allocating new ones, we probably want to do that here too.

There's probably a case to be made for just switching to byte[] everywhere, since that's what backs String in modern Java.

dev/core/src/com/google/gwt/dev/js/JsToStringGenerationVisitor.java
```diff
@@ -790,7 +833,11 @@ public boolean visit(JsReturn x, JsContext ctx) {
     _return();
     JsExpression expr = x.getExpr();
     if (expr != null) {
-      _space();
+      if (spaceReturn(expr)) {
+        _space();
+      } else {
+        _spaceOpt();
+      }
```
niloc132: I don't have a great solution for this, but I just wanted to call out briefly that this seems gross - we have to toString the child node twice, so that we know how to concat them together. Perf-wise (though probably not logic-wise) it would seem better to just go ahead and not accept(expr) if we know it is a literal, but append directly. 

Better still would be to signal to the TextOutput "hey you already know if space is optional or not... I just finished writing a keyword, make sure the next token gets a space if it needs it". Then, when the next token is written, TextOutput decides if it needs a space? 

One missing case in the current impl is to do with parens - in the terser sample you showed me, there are some cases of extra parens being required in order to maintain operator precedence:
```
return(t=h.h>>19)!=(u=n.h>>19)?...
```

dev/core/src/com/google/gwt/dev/js/JsToStringGenerationVisitor.java
```diff
@@ -1275,6 +1322,20 @@ private boolean _spaceCalc(JsOperator op, JsExpression arg) {
     return false;
   }
 
+  private boolean spaceReturn(JsExpression arg) {
```
niloc132: Convention in this file seems to be for private helpers to start with _. I don't really like it either, but it is at least consistent.

Also, my feelings to the contrary, there are other examples of helpers like this, such as _parenPushIfCommaExpr and friends.

dev/core/src/com/google/gwt/dev/util/AbstractTextOutput.java
```diff
@@ -73,11 +73,7 @@ public void indentOut() {
 
   @Override
   public void newline() {
-    if (compact) {
-      out.print('\n');
-    } else {
-      out.print('\n');
-    }
+    out.print('\n');
```
niloc132, line 76: With this change, this is the body of the `if` in `newlineOpt`, and we could have `newlineOpt` call `newline()`?

It bothers me a little that the "text output" and "source generation" are split into separate types. I'm sure there's a good way to reconcile this

dev/core/test/com/google/gwt/dev/js/JsToStringGenerationVisitorTest.java
```diff
@@ -39,6 +43,7 @@ public void setUp() throws Exception {
     stackMode.addDefinedValue(new ConditionNone(), "STRIP");
     setProperties(new BindingProperty[] {stackMode}, new String[] {"STRIP"},
         new ConfigurationProperty[] {});
+    runDeadCodeElimination = false;
```
niloc132: why set this to false, wouldn't it still be defaulting to false? doesn't each test method get a fresh instance to work with?

dev/core/src/com/google/gwt/dev/jjs/JavaToJavaScriptCompiler.java
```diff
@@ -1479,6 +1482,7 @@ private void optimizeJavaToFixedPoint() throws InterruptedException {
 
         nodeCount = jprogram.getNodeCount();
         mods = stats.getNumMods();
+        logger.log(TreeLogger.Type.TRACE, "Optimization pass " + passCount + " of " + passLimit + "\n" + stats);
```
niloc132: Deliberate? I'm not against it, but might want to match it with one in the java loop too.

JFR output has this info and more as well, can be handy (i'm very slowly writing docs on how to use this I promise)

vjay82: This also creates work/a String, even if nothing is logged ever.

niloc132: On the one hand, yes, you're not wrong, it is work and a new heap object (very short lived, so GC will jump on it, and it might even not allocate if you get JIT involved).

On the other hand, we're concating 6 objects every 100ms or so, and probably repeating it a max of about 3-5 times (unless something goes horribly wrong, which is exactly the time you want to have it logged).

The Java optimization loop is the one that runs "a lot", and that usually terminates after 12 or so loops (in my very limited sampling, but will be higher soon if we can start collecting some JFR dumps). I won't have qualms about 6*15 concat operations either ;).

dev/core/src/com/google/gwt/dev/js/JsToStringGenerationVisitor.java
zbynek: Appending `char[]` to `StringWriter` seems 20% slower.

For `FileWriter`, appending `char[]` is about 20% faster, but in that case having a `StringWriter`, adding all the strings to it and dumping the result into `FileWriter` saves ~60%.

The JS results go into a `StringWriter`-backed output. so I'd like to just switch to using strings here. `AstDumper` can print to `FileWriter` directly -- that could be maybe optimized in an independent PR.

niloc132: What about byte[]?

We also could be a lot more clever not immediately allocating strings, but accumulating bytes until we need a string or hit IO.

dev/core/src/com/google/gwt/dev/jjs/JavaToJavaScriptCompiler.java
zbynek: Committed accidentally, this was part of some unrelated experiment. Reverted now.

dev/core/src/com/google/gwt/dev/js/JsToStringGenerationVisitor.java
zbynek: > Perf-wise (though probably not logic-wise) it would seem better to just go ahead and not accept(expr) if we know it is a literal

That's implemented in my last commit. 

> One missing case in the current impl is to do with parens

I don't have a good solution for that -- would it be OK to leave that for another time?

niloc132: Yeah, I think we wait, rethink that as we implement other size enhancements.

dev/core/src/com/google/gwt/dev/js/JsToStringGenerationVisitor.java
zbynek: Using `ByteArrayOutputStream` instead of a `StringWriter` to connect strings seems to be roughly the same, BAOS can be 5% slower if both are initiated with the right size and 15% slower if they start empty.

I'm not even sure this is something worth optimizing, but dropping the calls to `toCharArray` removes complexity and hopefully saves a few ms.

zbynek: Using `ByteArrayOutputStream` instead of a `StringWriter` to connect strings seems to be roughly the same, BAOS can be 5% slower if both are initiated with the right size and 15% slower if they start empty.

I'm not even sure this is something worth optimizing, but dropping the calls to `toCharArray` removes complexity and hopefully saves a few ms.

niloc132: That's impressive, thanks for checking - good to know we can simplify code and make things faster.

dev/core/src/com/google/gwt/dev/js/JsToStringGenerationVisitor.java
zbynek: Uh, so we need to `accept` the literal node so that subclasses can handle it (for billing). For now I reverted to the duplicated serialization -- we could introduce a flag that lets us `accept` the node without printing, but it's not obvious if that can save time.

dev/core/src/com/google/gwt/dev/js/JsToStringGenerationVisitor.java
niloc132: Unlikely to matter much, but maybe move the actual "true vs !0" check into `_true()` itself? It does read a little oddly to have a ternary and some delegated methods in an if.

niloc132 (commented):
One quick note - I have other readability complaints, but I think they're too minor to be worth bothering you with.

zbynek:
> I have other readability complaints,

Let me know if you think there's something worth fixing.

niloc132 (approved):
Nope, I think that was really the only one that seems surprising enough to be worth adjusting.

Issue #10292, opened 2026-03-04
RichTextArea is broken with Firefox 148


TatuLund:
I think this is a regression bug in Firefox 148, and I filed bug there

https://bugzilla.mozilla.org/show_bug.cgi?id=2020927

I am just creating this issue here for cross referencing.

hsivonen:
In Firefox 148, Firefox got Chrome-compatible about:blank, so the `load` event for an `iframe` without `src` now fires immediately when the `iframe` element is inserted into the DOM.

Notably, RichTextArea works, if Firefox's UA string is overridden to report Chrome's UA string instead.

CKEditor 4 has a similar problem of doing UA-sniffing and assuming browser-specific `load` event behavior. That was caught during the Nightly phase of Firefox, and we have a compatibility hack that that fires an extra `load` event for CKEditor 4's editor iframe.

If I make the same hack that we have for CKEditor 4 apply when an iframe has the class value `gwt-RichTextArea`, GWT's RichTextArea works in Firefox without an UA string override.

In the CKEditor 4 case, we didn't want to give Web developers a magic class name that would change behavior on its own. As it happened, CKEditor 4 is EOL and CKEditor 5 doesn't have this problem. We made the compatibility hack require both CKEditor's class name _and_ `window.CKEDITOR.version` starting with `4.`. (It's a bit unclear if that was wise: We found a fork that renamed `window.CKEDITOR` without renaming the class name and now a case of deploying CKEditor without performing a build step that would put the version number in the right place has been reported.)

If we wanted to check for some GWTness signal on the global scope of the document into which the RichTextArea `iframe` gets inserted into, is there some check that we could do to 1) get confidence of GWTness beyond the `gwt-RichTextArea` class name and 2) turn the compat hack off for future versions of GWT?

TatuLund:
If I monkey patch RichtTextAreaImplMozilla in my project with this the RichTextArea works again, apparently the old workarounds are no longer needed (maybe at quick glance).

```
package com.google.gwt.user.client.ui.impl;

/**
 * Mozilla-specific implementation of rich-text editing.
 */
public class RichTextAreaImplMozilla extends RichTextAreaImplStandard {

  @Override
  public String getBackColor() {
    return queryCommandValue("HiliteColor");
  }
}
```

TatuLund:
> no longer needed (maybe at quick glance).

~Perhaps premature, now I can't tab into it.~

Well that is also the original behavior, so not related to removed workarounds

[...]

PR #10304, opened 2026-03-24
Fix Safari timer throttling in CrossSiteIframeLinker bootstrap iframe

### LINKED ISSUE #10303: CrossSiteIframeLinker bootstrap iframe causes 2-4+ minute load delays in Safari
## Summary

GWT's default `CrossSiteIframeLinker` creates a hidden bootstrap iframe with
`position:absolute; width:0; height:0; left:-1000px; top:-1000px`. In Safari,
this iframe has zero intersection with the viewport, which triggers aggressive
DOM timer throttling. As a result, GWT's `onModuleLoad()` is delayed by
2–4+ minutes on Safari while it fires in ~1 second on Chrome under identical
conditions.

## Environment

- **GWT version**: 2.13.0
- **Linker**: `CrossSiteIframeLinker` (default `xsiframe`)
- **Affected browsers**: Safari 18.6 macOS (10–115s delay), Safari 26.2 macOS (~244s delay)
- **Unaffected browsers**: Chrome 146 (~1s load time on same hardware/network)

## Observed Symptom

1. `<module>nocache.js` and `<hash>.cache.js` both load immediately (within ~1 second).
2. The browser then sits idle for 2–4+ minutes with a loading spinner.
3. All subsequent artifacts (images, RPC calls) load in a burst once `onModuleLoad()` finally fires.

During the idle window, the server receives **zero** GWT RPC requests — the delay is entirely client-side.

## Root Cause

Two contributing factors:

### 1. Iframe style puts it outside the viewport

Safari (WebKit) throttles DOM timers (`setTimeout` / `setInterval`) in iframes with an empty viewport intersection. An iframe at `left:-1000px; top:-1000px` with `width:0; height:0` has exactly such a zero visible rect.

Note: `visibility:hidden` and `opacity:0` also trigger the throttle — only a genuinely rendered element with non-zero viewport intersection prevents it. `opacity:0.01` works; `opacity:0` does not.

### 2. `document.write` triggers a throttled navigation cycle

`installLocationIframe.js` calls `frameDoc.open()` / `frameDoc.write()` / `frameDoc.close()` for all non-Chrome browsers (Safari included). Commit `43ca6fb` skipped `document.write` in Chrome but not Safari. The `document.write` + `document.close()` sequence triggers a navigation cycle inside the iframe, putting Safari's browsing context into a throttled state even on reload.

### Why the existing workaround is insufficient

The existing comment notes `display:none` was avoided because it prevents code from running. The chosen alternative (`position:absolute` at −1000px) still produces a zero-intersection viewport rect in modern Safari, triggering the same throttle by a different path.

## Evidence

| Browser | Load time |
|---|---|
| Chrome 146 | ~1 second |
| Safari 18.6 (warm cache) | 10–115 seconds |
| Safari 26.2 (clean load) | ~244 seconds |

The throttle triggers reliably on the first Shift+Reload in Safari. Server-side RPC logs confirm zero requests arrive during the idle window — the bottleneck is entirely within the browser before `onModuleLoad()` fires.

## Suggested Fix

Three changes to `installLocationIframe.js`:

**1. Iframe style** — use a fixed, nearly-invisible in-viewport element:
```javascript
scriptFrame.style.cssText = 'position:fixed; left:0; top:0; width:50px; height:50px;'
  + ' border:none; opacity:0.01;';
```

**2. Skip `document.write` for Safari** — extend the existing Chrome check:
```javascript
var isSafari = navigator.vendor && navigator.vendor.indexOf("Apple") != -1;
if (navigator.userAgent.indexOf("Chrome") == -1 && !isSafari) { ... }
```

**3. Clean up stale iframes** — add before creating the new iframe:
```javascript
var stale = $doc.getElementById('__MODULE_NAME__');
if (stale && stale.parentNode) { stale.parentNode.removeChild(stale); }
```

## Related Links

- [`installLocationIframe.js` source](https://github.com/gwtproject/gwt/blob/main/dev/core/src/com/google/gwt/core/ext/linker/impl/installLocationIframe.js)
- [GWT commit `43ca6fb` — avoid `document.write` in Chrome](https://github.com/gwtproject/gwt/commit/43ca6fb)
- [WICG Interventions Issue #9 — out-of-view iframe throttling](https://github.com/WICG/interventions/issues/9)

jeanphil77:
Safari (WebKit) throttles DOM timers in iframes with an empty viewport
intersection, causing `onModuleLoad()` to be delayed by 2–4+ minutes
(vs ~1s on Chrome) when the bootstrap iframe is positioned off-screen.

Three changes to `installLocationIframe.js`:

1. **Iframe style**: replace the off-screen `position:absolute` at `-1000px`
   with `position:fixed` at `(0,0)` with `width:50px; height:50px; opacity:0.01`.
   The element is in the viewport (non-zero intersection), preventing WebKit
   throttling, while remaining imperceptible to users.
   Note: `opacity:0` and `visibility:hidden` also trigger throttling.

2. **Skip `document.write` for Safari**: `document.write()` + `document.close()`
   triggers a navigation cycle that puts Safari's browsing context into a
   throttled state. Extends the existing Chrome bypass (commit `43ca6fb`)
   to cover Safari as well.

3. **Remove stale iframes** before creating a new one: prevents accumulation
   across reloads and avoids inheriting a throttled browsing-context state
   (most visible on Shift+Reload in Safari).

Affected: Safari 18.6 (10–115s delay), Safari 26.2 (~244s delay).
Unaffected browsers see no functional or visual change.

Fixes #10303

### DIFF
--- a/dev/core/src/com/google/gwt/core/ext/linker/impl/installLocationIframe.js
+++ b/dev/core/src/com/google/gwt/core/ext/linker/impl/installLocationIframe.js
@@ -28,25 +28,46 @@ function getInstallLocation() {
 
 function setupInstallLocation() {
   if (frameDoc) { return; }
+
+  // Remove any stale bootstrap iframe left over from a previous page load
+  // (e.g. after a Shift+Reload in Safari) to prevent accumulation and avoid
+  // inheriting a throttled browsing-context state.
+  var stale = $doc.getElementById('__MODULE_NAME__');
+  if (stale && stale.parentNode) {
+    stale.parentNode.removeChild(stale);
+  }
+
   // Create the script frame, making sure it's invisible, but not
   // "display:none", which keeps some browsers from running code in it.
+  //
+  // The iframe must have a non-zero intersection with the viewport.
+  // Safari (WebKit) throttles DOM timers in iframes whose visible rect is
+  // empty, which includes iframes placed off-screen at negative coordinates
+  // as well as those hidden via visibility:hidden or opacity:0.  A 50x50px
+  // fixed element at the top-left corner with opacity:0.01 is imperceptible
+  // to users while reliably preventing the throttle.
   var scriptFrame = $doc.createElement('iframe');
   scriptFrame.id = '__MODULE_NAME__';
-  scriptFrame.style.cssText = 'position:absolute; width:0; height:0; border:none; left: -1000px;'
-    + ' top: -1000px;';
+  scriptFrame.style.cssText = 'position:fixed; left:0; top:0; width:50px; height:50px;'
+    + ' border:none; opacity:0.01;';
   scriptFrame.tabIndex = -1;
   $doc.body.appendChild(scriptFrame);
 
   frameDoc = scriptFrame.contentWindow.document;
 
-  // The following code is needed for proper operation in Firefox, Safari, and
+  // The following code is needed for proper operation in Firefox and
   // Internet Explorer.
   //
   // In Firefox, this prevents the frame from re-loading asynchronously and
   // throwing away the current document.
   //
   // In IE, it ensures that the <body> element is immediately available.
-  if (navigator.userAgent.indexOf("Chrome") == -1) {
+  //
+  // Safari is excluded because document.write() + document.close() triggers
+  // a navigation cycle that puts Safari's browsing context into a throttled
+  // state, causing the same multi-minute delay as the off-screen iframe style.
+  var isSafari = navigator.vendor && navigator.vendor.indexOf("Apple") != -1;
+  if (navigator.userAgent.indexOf("Chrome") == -1 && !isSafari) {
     frameDoc.open();
     var doctype = (document.compatMode == 'CSS1Compat') ? '<!doctype html>' : '';
     frameDoc.write(doctype + '<html><head></head><body></body></html>');


dev/core/src/com/google/gwt/core/ext/linker/impl/installLocationIframe.js
```diff
@@ -28,25 +28,46 @@ function getInstallLocation() {
...
   // throwing away the current document.
   //
   // In IE, it ensures that the <body> element is immediately available.
-  if (navigator.userAgent.indexOf("Chrome") == -1) {
+  //
+  // Safari is excluded because document.write() + document.close() triggers
+  // a navigation cycle that puts Safari's browsing context into a throttled
+  // state, causing the same multi-minute delay as the off-screen iframe style.
+  var isSafari = navigator.vendor && navigator.vendor.indexOf("Apple") != -1;
+  if (navigator.userAgent.indexOf("Chrome") == -1 && !isSafari) {
```
niloc132, line 70: IE no longer exists, so this is just a very verbose "is it firefox" check, better to be rewritten as such.

Do we still need this check for FF, or can we remove it outright?

zbynek: Back when I implemented https://github.com/gwtproject/gwt/pull/9836 it was needed for the then-stable FF version, but as discussed in https://github.com/gwtproject/gwt/issues/10292 the behavior of anonymous frames changed in 148.

jeanphil77:
Closing PR for now as we are unable to reliably assess the root cause of the issue and whether the fix actually addressess it. See associated issues for details.

Issue #10295, opened 2026-03-10
Audit and remove very old Firefox / Gecko UA detection


hsivonen:
A search for Gecko and for Firefox in this repo finds a number of UA detection cases that appear to have been added before Firefox 5.0, which was released in 2011.

It seems unlikely that projects using GWT today care about the results working in very old browsers, and, depending on the specifics, browser detection poses the risk of things going wrong in the future.

It would make sense to audit the UA detection cases, remove very old ones, and if some things don't work without browser detection, trying to come up with alternative code that works without browser detection in multiple present-day engines.

See #10292 for a recent case of GWT's Gecko-specific code breaking when Gecko's behavior aligned with Blink.

niloc132:
The last time I dug in, RTL locale behavior was still quite different between "webkit/blink" and "gecko" browsers, so it is no where near as simple as just removing the checks. My recollection is that Gecko ended up having a nicer implementation to work with, so the "hacks" exist on both sides, and we wouldn't want to just pick one implementation across the board, but need to test each case individually. 

Additionally, there are still JS level differences as well - Firefox for example canonicalizes all NaNs when read/written to typedarrays, while at least Chrome does not and preserves the original value (try running https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN#observably_distinct_nan_values in chrome/ff to observe differences). I'm not even sure what one should do to "feature detect" numbers losing data like this - we've just forbidden that feature entirely, since there is no way good to emulate it.

I don't disagree that it would be better to not have the differences, but the browsers still have notable differences in how they behave, so each edge case would need to be revisited, understood, tested, validated against LTS releases of any browser... Even "very old" differences are not necessarily exempt. Definitely worth doing, but quite time intensive.

niloc132:
Small update, most of the version checks in com.google.gwt.dom.client.DOMImplMozilla are entirely unused, and `isGecko19()` is used only for (really really) old RTL behavior. We still naturally need to know the difference between gecko and others (I looked into RTL, webkit-derived browsers evidently required scrollleft values to be compensated with scrollwidth and clientwidth, and as of a year or three ago they still did).

Most issues I looked at are long since closed, with a few exceptions: https://bugzilla.mozilla.org/show_bug.cgi?id=208427 was closed as of FF 148, so we'll be keeping that workaround for a while yet.

[...]

PR #10308, opened 2026-04-15
Remove very old FireFox version checks

zbynek:
Removes some very old checks, partial fix for #10295

### DIFF
--- a/user/src/com/google/gwt/dom/client/DOMImplMozilla.java
+++ b/user/src/com/google/gwt/dom/client/DOMImplMozilla.java
@@ -56,24 +56,6 @@ class DOMImplMozilla extends DOMImplStandard {
   }
 
   @Override
-  public native EventTarget eventGetRelatedTarget(NativeEvent evt) /*-{
-//  fixed
-    // Hack around Mozilla bug 497780 (relatedTarget sometimes returns XUL
-    // elements). Trying to access relatedTarget.nodeName will throw an
-    // exception if it's a XUL element.
-    var relatedTarget = evt.relatedTarget;
-    if (!relatedTarget) {
-      return null;
-    }
-    try {
-      var nodeName = relatedTarget.nodeName;
-      return relatedTarget;
-    } catch (e) {
-      return null;
-    }
-  }-*/;
-
-  @Override
   public int getAbsoluteLeft(Element elem) {
     return getAbsoluteLeftImpl(elem.getOwnerDocument().getViewportElement(),
         elem);
@@ -88,22 +70,12 @@ class DOMImplMozilla extends DOMImplStandard {
   @Override
   public native int getBodyOffsetLeft(Document doc) /*-{
     var style = $wnd.getComputedStyle(doc.documentElement, null);
-    if (style == null) {
-    //fixed
-      // Works around https://bugzilla.mozilla.org/show_bug.cgi?id=548397
-      return 0;
-    }
     return parseInt(style.marginLeft, 10) + parseInt(style.borderLeftWidth, 10);
   }-*/;
 
   @Override
   public native int getBodyOffsetTop(Document doc) /*-{
     var style = $wnd.getComputedStyle(doc.documentElement, null);
-    if (style == null) {
-    //fixed
-      // Works around https://bugzilla.mozilla.org/show_bug.cgi?id=548397
-      return 0;
-    }
     return parseInt(style.marginTop, 10) + parseInt(style.borderTopWidth, 10);
   }-*/;
 
@@ -118,7 +90,7 @@ class DOMImplMozilla extends DOMImplStandard {
       //
       // See https://bugzilla.mozilla.org/show_bug.cgi?id=208427
       // and http://code.google.com/p/google-web-toolkit/issues/detail?id=1909
-      // fixed FF 148, reevaluate once ESR > 148
+      // Fixed FF 148, reevaluate once ESR > 148
       return 0;
     }
   }-*/;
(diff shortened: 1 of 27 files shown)


user/src/com/google/gwt/user/cellview/client/AbstractCellTable.java
```diff
@@ -1949,8 +1915,8 @@ protected void onBrowserEvent2(Event event) {
           boolean unhover = true;
           if (!skipRowHoverFloatElementCheck) {
             // Ignore events happening directly over the hovering row. If there are floating element
-            // on top of the row, mouseout event should not be triggered. This is to avoid the flickring
-            // effect if the floating element is shown/hide based on hover event.
+            // on top of the row, mouseout event should not be triggered. This is to avoid the
+            // flickring effect if the floating element is shown/hide based on hover event.
```
niloc132:
```suggestion
            // flickering effect if the floating element is shown/hide based on hover event.
```
as long as we're tweaking this code

niloc132 (commented):
I have a local branch that was also attempting to do this, I'll make a PR to your PR and see if we can't get all the fixes in

user/src/com/google/gwt/dom/client/DOMImplMozilla.java
```diff
@@ -46,47 +46,7 @@ private static native int getNativeGeckoVersion() /*-{
...
-
-  /**
-   * Return true if using Gecko 2.0.0 (Firefox 4.0) or earlier.
-   * 
-   * @return true if using Gecko 2.0.0 (Firefox 4.0) or earlier
-   */
-  private static boolean isGecko2OrBefore() {
-    int geckoVersion = getGeckoVersion();
-    return (geckoVersion != -1) && (geckoVersion < 2000000);
+    return geckoVersion >= 1009000;
```
niloc132: It looks like my version made this always return true, then inlined the method away, so that we both don't need to compute the version and don't need the isRTL checks later in the file.

niloc132:
Take a look at https://github.com/zbynek/gwt/pull/6 when you get the chance, I think it completes this work.

user/src/com/google/gwt/dom/client/DOMImplMozilla.java
```diff
@@ -156,20 +70,12 @@ public int getAbsoluteTop(Element elem) {
   @Override
   public native int getBodyOffsetLeft(Document doc) /*-{
     var style = $wnd.getComputedStyle(doc.documentElement, null);
-    if (style == null) {
-      // Works around https://bugzilla.mozilla.org/show_bug.cgi?id=548397
-      return 0;
-    }
     return parseInt(style.marginLeft, 10) + parseInt(style.borderLeftWidth, 10);
```
niloc132: Note that I think this might be wrong in other browsers - the superclass only returns 0, mozilla is the only impl that actually reads values from the dom. Ditto offsettop below. Probably out of scope here, just seeking to remove this extra null check.

user/test/com/google/gwt/http/server/RequestBuilderTestServlet.java
```diff
@@ -104,6 +104,7 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
          * text the status message will be undefined. So, we make sure that the
          * post returns some data. See
          * http://bugs.webkit.org/show_bug.cgi?id=3810.
+         * //fixed
```
niloc132: Whoops, I forgot to remove this from my initial research steps

```suggestion
```

user/src/com/google/gwt/user/cellview/client/AbstractCellTable.java
```diff
@@ -638,44 +638,14 @@ protected void replaceAllRowsImpl(AbstractCellTable<?> table, TableSectionElemen
     }
   }
 
-  /**
-   * Implementation of {@link CellTable} used by Firefox.
-   */
-  @SuppressWarnings("unused")
-  private static class ImplMozilla extends Impl {
```
niloc132, line 645: Looking more carefully, maybe we should keep the class with no contents, as the Trident classes, just in case someone has a custom replace-with

niloc132 (commented):
"Self review" of a sort - I can push you a commit if you'd prefer

user/src/com/google/gwt/dom/client/DOMImplMozilla.java
zbynek: The FF implementation matches the name, but the WebKit implementation results in more sensible behavior.

1) go to https://samples.gwtproject.org/samples/Showcase/Showcase.html#!CwBasicPopup
2) run `document.documentElement.style.margin="100px";` in the console
3) click the image

Expected: popup centered
Actual: popup too far left in FF (but centered in WebKit)

user/src/com/google/gwt/user/cellview/client/AbstractCellTable.java
zbynek: The class does not change behavior in any existing browser, so the fix for anyone using it is to simply remove the `replace-with`. With the Trident stuff it might be possible (though very unlikely) that someone depends on the changes in behavior it provides. I think it's fine to remove this.

user/src/com/google/gwt/user/client/impl/WindowImplMozilla.java
```diff
@@ -17,14 +17,16 @@
 
 /**
  * Mozilla implementation of {@link com.google.gwt.user.client.impl.WindowImpl}.
+ * @deprecated This class is only used to make tests work with HtmlUnit 2.55.0, and will
```
zbynek, line 20: @niloc132 I didn't verify that this can be actually removed with HtmlUnit 4+, but I think it's better to first do that upgrade and then get back to this.

PR #10312, opened 2026-04-22
Include GWT version, commit in JFR output

niloc132:
Enhances the "manual, opt-in analytics" by allowing users to upload JFR dumps of a build and detect which particular version or build resulted in the dump.

Deliberately omits command line args at this time - users can leave the default JVM events in place so these are detected or specify them manually. An automatic approach would need more code, able to remove details like paths and module names from the output.

Fixes #10269


dev/core/src/com/google/gwt/dev/About.java
zbynek:
```suggestion
   * Tag used for text replacement of the Git version (split up to avoid
```

vegegoku (dismissed):

vegegoku:
There is a checkstyle issue.

zbynek:
https://github.com/gwtproject/gwt/blob/6a2216c337e8e8ae2f6e728f71c744a3d7a256b7/dev/core/test/com/google/gwt/dev/AboutTest.java#L27 needs updating too

zbynek (approved):

PR #10313, opened 2026-04-22
Use Objects.equals to compare record fields for null support

### LINKED ISSUE #10293: equals works incorrectly for Java Records if some fields are initialized with null
If some fields in java record are initialized with `null` the equals return `false`.

```java
public record A(String v) {}
```
```java
boolean equals1 = new A("A").equals(new A("A"));
boolean equals2 = new A(null).equals(new A(null));
```
Both `equals1` and `equals2` have to be `true` but `equals2` is `false`

GWT 2.13.0

niloc132:
This likely generates slightly more code than before, but the code it emits is correct.

Fixes #10293


dev/core/src/com/google/gwt/dev/jjs/impl/ImplementRecordComponents.java
```diff
@@ -190,28 +193,18 @@ private void implementEquals(JRecordType type, JMethod method, SourceInfo info)
...
         JFieldRef otherField = new JFieldRef(info, typedOther.createRef(info), field, type);
-        final JBinaryOperation equals;
+        final JExpression equals;
         if (field.getType().isPrimitiveType()) {
           equals = new JBinaryOperation(info, JPrimitiveType.BOOLEAN,
                   JBinaryOperator.EQ,
                   myField,
                   otherField);
         } else {
-          // We would like to use Objects.equals here to be more concise, but we would need
```
zbynek, line 205: So this is now resolved with `@SpecializeMethod` 👍 

The alternative solution would be to just add Strings to the same branch here as primitive types

zbynek (approved):

vegegoku (approved):

PR #10314, opened 2026-04-27
Add boolean matches(String selectors) method

Greunlis:
I propose adding a matches method as defined here: https://developer.mozilla.org/en-US/docs/Web/API/Element/matches

### DIFF
--- a/user/src/com/google/gwt/dom/client/Element.java
+++ b/user/src/com/google/gwt/dom/client/Element.java
@@ -531,6 +531,15 @@ public class Element extends Node {
   }
 
   /**
+   * Tests whether the element would be selected by the specified CSS selector.
+   * @param selectors A string containing valid CSS selectors to test the Element against.
+   * @return true if the Element matches the selectors. Otherwise, false.
+   */
+  public final native boolean matches(String selectors) /*-{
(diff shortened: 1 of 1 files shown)


niloc132:
Thanks for the submission - at the very least, we would want to have tests to accompany it. I would also be concerned about a caller who cannot tell if "false" means "doesn't match" or "matches, but the method doesn't exist".

Note however that this is already supported in `elemental2.dom.Element`, as well as many other other methods not found in dom.client.Element and subclasses. As Elemental2 is auto generated (from [hand maintained externs](https://github.com/google/closure-compiler/blob/master/externs/)) it will tend to be more up to date with what the rest of the browser world can support.

Does it make sense to just use elemental2 for your use cases where modern APIs are required?

Greunlis:
Thanks for pointing me to elemental2! Adding it just for this method alone seems like an overkill, but I might find other use cases in our codebase, I'll give it some more thought and either add a test or close this pull request as unnecessary.

niloc132:
Closing for now - I think elemental is the better way forward here.

Greunlis:
Thank you, closing, I started using elemental2 for this.

PR #10322, opened 2026-05-27
Make StringQuoter date parsing thread-safe

metsw24-max:
RequestFactory and AutoBean decode Date values on the server through StringQuoter.tryParseDate, which shared two static SimpleDateFormat instances across all request threads. SimpleDateFormat parsing mutates internal Calendar state, so concurrent requests race and return wrong dates or throw. Give each thread its own formatter via ThreadLocal.

### DIFF
--- a/user/test/com/google/web/bindery/autobean/vm/StringQuoterJreTest.java
+++ b/user/test/com/google/web/bindery/autobean/vm/StringQuoterJreTest.java
@@ -1,0 +1,115 @@
+/*
+ * Copyright 2026 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.web.bindery.autobean.vm;
+
+import com.google.web.bindery.autobean.shared.impl.StringQuoter;
+
+import junit.framework.TestCase;
+
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Locale;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Tests for the server-side {@link StringQuoter#tryParseDate(String)}.
+ */
+public class StringQuoterJreTest extends TestCase {
+
+  private static final String ISO8601_PATTERN = "yyyy-MM-dd'T'HH:mm:ss.SSSz";
+  private static final String RFC2822_PATTERN = "EEE, d MMM yyyy HH:mm:ss Z";
+
+  public void testTryParseDateMillis() {
+    Date d = new Date(1234567890123L);
+    assertEquals(d, StringQuoter.tryParseDate(Long.toString(d.getTime())));
+  }
+
+  public void testTryParseDateIso8601() {
+    SimpleDateFormat fmt = new SimpleDateFormat(ISO8601_PATTERN, Locale.getDefault());
+    Date d = new Date(1234567890123L);
+    assertEquals(d, StringQuoter.tryParseDate(fmt.format(d)));
+  }
+
+  public void testTryParseDateZuluSuffix() throws Exception {
+    SimpleDateFormat fmt = new SimpleDateFormat(ISO8601_PATTERN, Locale.getDefault());
+    Date expected = fmt.parse("2024-01-15T10:30:00.000+0000");
+    assertEquals(expected, StringQuoter.tryParseDate("2024-01-15T10:30:00.000Z"));
+  }
+
+  public void testTryParseDateRfc2822() {
+    // RFC 2822 has second resolution.
+    SimpleDateFormat fmt = new SimpleDateFormat(RFC2822_PATTERN, Locale.getDefault());
+    Date d = new Date(1234567890000L);
+    assertEquals(d, StringQuoter.tryParseDate(fmt.format(d)));
+  }
+
+  public void testTryParseDateUnparseable() {
+    assertNull(StringQuoter.tryParseDate("not a date"));
+  }
+
+  /**
+   * SimpleDateFormat is not thread-safe; sharing a single instance across
+   * request threads on the server produced either ParseExceptions (returned as
+   * null) or silently wrong Dates. Hammer tryParseDate from many threads and
+   * make sure every call returns the expected value.
+   */
+  public void testTryParseDateConcurrent() throws Exception {
+    SimpleDateFormat fmt = new SimpleDateFormat(ISO8601_PATTERN, Locale.getDefault());
+    final Date expected = new Date(1234567890123L);
+    final String input = fmt.format(expected);
+
+    final int threads = 16;
+    final int perThread = 2000;
+    final CountDownLatch start = new CountDownLatch(1);
+    final AtomicInteger nulls = new AtomicInteger();
+    final AtomicInteger wrong = new AtomicInteger();
+    ExecutorService pool = Executors.newFixedThreadPool(threads);
+    try {
+      for (int i = 0; i < threads; i++) {
+        pool.submit(new Runnable() {
+          @Override
+          public void run() {
+            try {
+              start.await();
+            } catch (InterruptedException e) {
+              Thread.currentThread().interrupt();
+              return;
+            }
+            for (int j = 0; j < perThread; j++) {
+              Date r = StringQuoter.tryParseDate(input);
+              if (r == null) {
+                nulls.incrementAndGet();
+              } else if (!expected.equals(r)) {
+                wrong.incrementAndGet();
+              }
+            }
+          }
+        });
+      }
+      start.countDown();
+      pool.shutdown();
+      assertTrue("threads did not finish in time", pool.awaitTermination(60, TimeUnit.SECONDS));
+    } finally {
+      pool.shutdownNow();
+    }
+    assertEquals("parses returned null", 0, nulls.get());
+    assertEquals("parses returned wrong date", 0, wrong.get());
+  }
+}
--- a/user/test/com/google/web/bindery/autobean/vm/StringQuoterJreTest.java
+++ b/user/test/com/google/web/bindery/autobean/vm/StringQuoterJreTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2026 Google Inc.
+ * Copyright 2026 GWT Project Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License"); you may not
  * use this file except in compliance with the License. You may obtain a copy of
--- a/user/test/com/google/web/bindery/autobean/vm/StringQuoterJreTest.java
+++ b/user/test/com/google/web/bindery/autobean/vm/StringQuoterJreTest.java
@@ -19,12 +19,13 @@ import com.google.web.bindery.autobean.shared.impl.StringQuoter;
 
 import junit.framework.TestCase;
 
-import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Date;
-import java.util.Locale;
+import java.util.List;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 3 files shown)


niloc132:
Interesting - the code is definitely wrong, but does this actually fail in any real use case? It looks like you could manufacture such a scenario by manually writing JSON, but clearly AutoBeans (and thus, RequestFactory) won't serialize dates using these formats. I imagine these were speculatively added, but never actually used.

https://github.com/gwtproject/gwt/blob/4e6ae10e9798e73fa490dabc64b7a742e933a66b/user/src/com/google/web/bindery/autobean/shared/ValueCodex.java#L113-L115

That seems contradicted by discussion at https://github.com/gwtproject/gwt/issues/6330 (esp https://github.com/gwtproject/gwt/issues/6330#issuecomment-110838704) and https://github.com/gwtproject/gwt/commit/3df95e6fef5c303d9e2fc0ca8c16c736cac99cdf. Nevertheless, there are still no tests using date strings that I can find.

I'm apprehensive about merging this despite its correctness, given the abandoned https://github.com/gwtproject/gwt/pull/10307 and the behavior of the account (many PRs to many unrelated repos with weak followup and tendency towards LLM-like content).

vjay82:
If the code was never used, then it could be removed, no? However, if it is in use, it looks more correct than before, proposed by a LLM or not, no?

niloc132:
Right - the code looks right, but the use case needs tests. The linked issue makes me suspect it is (or at least "was") used in some capacity, so removing a feature of a public api is at least somewhat dangerous, and there's probably no downside... but especially if it is tool-generated code, it should come with a test.

As a maintainer, I think it is appropriate for me to have a degree of skepticism for "more correct than before" code coming from low-effort sources, as a way to gain trust and attack the project. I'm definitely not accusing @metsw24-max here - the fixes to all the various repos they've sent code to seem well-intentioned, but very few have tests or actual use cases.

metsw24-max:
Fair point on the missing tests. Pushed a JRE test (StringQuoterJreTest) that covers the millis, ISO-8601, Z-suffix, RFC 2822, and unparseable paths, plus a concurrent hammer on tryParseDate that fires 16 threads x 2000 parses at the same date string. On the unpatched code it reliably produces wrong Dates (got ~11 on my box); with the ThreadLocal it sits at zero. Also wired into AutoBeanSuite.

niloc132:
Thanks - locally I haven't had the test pass without the fix, though I did see only one wrong date value (out of 16*2000 checks) - usually single digit, but rather less than 5, so the test does what it is there to do.

zbynek:
Why wasn't this caught by https://errorprone.info/bugpattern/DateFormatConstant ? The rule [exists in 2.33](https://github.com/google/error-prone/blob/v2.33.0/core/src/main/java/com/google/errorprone/bugpatterns/DateFormatConstant.java).

user/test/com/google/web/bindery/autobean/vm/StringQuoterJreTest.java
```diff
@@ -0,0 +1,115 @@
...
+ * Tests for the server-side {@link StringQuoter#tryParseDate(String)}.
+ */
+public class StringQuoterJreTest extends TestCase {
+
+  private static final String ISO8601_PATTERN = "yyyy-MM-dd'T'HH:mm:ss.SSSz";
+  private static final String RFC2822_PATTERN = "EEE, d MMM yyyy HH:mm:ss Z";
+
+  public void testTryParseDateMillis() {
+    Date d = new Date(1234567890123L);
+    assertEquals(d, StringQuoter.tryParseDate(Long.toString(d.getTime())));
```
zbynek: I think it would be more transparent if all arguments for `tryParseDate` in this file were string literals to make it obvious what kinds of inputs `tryParseDate` accepts.

user/test/com/google/web/bindery/autobean/vm/StringQuoterJreTest.java
```diff
@@ -0,0 +1,115 @@
...
+    assertNull(StringQuoter.tryParseDate("not a date"));
+  }
+
+  /**
+   * SimpleDateFormat is not thread-safe; sharing a single instance across
+   * request threads on the server produced either ParseExceptions (returned as
+   * null) or silently wrong Dates. Hammer tryParseDate from many threads and
+   * make sure every call returns the expected value.
+   */
+  public void testTryParseDateConcurrent() throws Exception {
```
jnehlmeier, line 69: There is quite a bit going on here:

1. `pool.submit()` does return a `Future` and that `Future` is not recorded and checked for exceptions. Without the `ThreadLocal` code in `StringQuoter` the `SimpleDateFormat.parse()` method can actually throw exceptions other than `ParseException` (for example `NumberFormatException`) during concurrent execution. Thus `StringQuoter.tryParseDate()` might throw exceptions which are only recorded in the `Future`. In that case the test might be green because neither `null` or `wrong` have been increased.
2. The `CountDownLatch.countDown()` method is called directly after the submit `for`-loop. This does not guarantee that all 16 threads are actually already waiting. Sometimes they do, sometimes they don't. Submitted work can be put into a queue. What you really want is calling `start.countDown()` followed by `start.await()` inside the `run()` method and use a `CountDownLatch(threads)`. Otherwise a second `CountDownLatch` is needed if we really want a dedicated signal to start the work in the test method itself.
3. `InterruptedException` should be rethrown as a runtime exception. If 1. is implemented the test can check for 16 `InterruptedException` in the Future to figure out if the test has actually been executed or not. The test is generally brittle and we might not want to trust a green test if too many (or all threads) have been interrupted before actually doing anything.
4. `start.await()` should probably use a timeout as well and some `AssertionError` with a message should be thrown if `await` returns via timeout. This is not strictly needed but gives a hint that thread synchronization did not happen in a timely manner. Also it kind of protects against a stupid future issue if for whatever reason `CountDownLatch` and `ExecutorService` have not been initialized with the same number (e.g. latch > pool threads).

niloc132:
@zbynek nice find - it looks like that is configured as a warning rather than an error, and common.ant.xml sets `nowarn` so we don't get warnings in the logs. There are a lot of warnings though... some might be valid, but a large number of them are javadoc or deprecation related.

https://github.com/google/error-prone/issues/424 exists to let projects pick out warnings to disable or let them fail the build, but hasn't been fixed.

When I turn warnings on, I don't see the DateFormatConstant in the output - we might have too many warnings to see all of them...

Here's a count of each warning type, via `ant clean dist-dev | grep bugpattern | sort -nr | uniq -c | sort -nr`
```
   4679     [javac]     (see https://errorprone.info/bugpattern/MissingOverride)
    341     [javac]     (see https://errorprone.info/bugpattern/JavaUtilDate)
    284     [javac]     (see https://errorprone.info/bugpattern/ReferenceEquality)
    248     [javac]     (see https://errorprone.info/bugpattern/InvalidInlineTag)
    241     [javac]     (see https://errorprone.info/bugpattern/NonApiType)
    213     [javac]     (see https://errorprone.info/bugpattern/UnnecessaryParentheses)
    198     [javac]     (see https://errorprone.info/bugpattern/InlineMeSuggester)
    146     [javac]     (see https://errorprone.info/bugpattern/SameNameButDifferent)
    143     [javac]     (see https://errorprone.info/bugpattern/JdkObsolete)
    137     [javac]     (see https://errorprone.info/bugpattern/UnusedVariable)
    116     [javac]     (see https://errorprone.info/bugpattern/BadImport)
     89     [javac]     (see https://errorprone.info/bugpattern/StringSplitter)
     75     [javac]     (see https://errorprone.info/bugpattern/DefaultCharset)
     72     [javac]     (see https://errorprone.info/bugpattern/IdentityHashMapUsage)
     71     [javac]     (see https://errorprone.info/bugpattern/OperatorPrecedence)
     67     [javac]     (see https://errorprone.info/bugpattern/NotJavadoc)
     65     [javac]     (see https://errorprone.info/bugpattern/InvalidLink)
     64     [javac]     (see https://errorprone.info/bugpattern/TypeParameterUnusedInFormals)
     63     [javac]     (see https://errorprone.info/bugpattern/ProtectedMembersInFinalClass)
     59     [javac]     (see https://errorprone.info/bugpattern/AnnotateFormatMethod)
     55     [javac]     (see https://errorprone.info/bugpattern/InvalidBlockTag)
     43     [javac]     (see https://errorprone.info/bugpattern/UnrecognisedJavadocTag)
     35     [javac]     (see https://errorprone.info/bugpattern/ClassCanBeStatic)
     34     [javac]     (see https://errorprone.info/bugpattern/ClassNewInstance)
     33     [javac]     (see https://errorprone.info/bugpattern/InvalidParam)
     32     [javac]     (see https://errorprone.info/bugpattern/StaticAssignmentInConstructor)
     24     [javac]     (see https://errorprone.info/bugpattern/StringCharset)
     22     [javac]     (see https://errorprone.info/bugpattern/FallThrough)
     20     [javac]     (see https://errorprone.info/bugpattern/MissingCasesInEnumSwitch)
     18     [javac]     (see https://errorprone.info/bugpattern/ImmutableEnumChecker)
     17     [javac]     (see https://errorprone.info/bugpattern/EqualsGetClass)
     16     [javac]     (see https://errorprone.info/bugpattern/BoxedPrimitiveConstructor)
     14     [javac]     (see https://errorprone.info/bugpattern/OutlineNone)
     12     [javac]     (see https://errorprone.info/bugpattern/InlineFormatString)
     11     [javac]     (see https://errorprone.info/bugpattern/ReturnAtTheEndOfVoidFunction)
     11     [javac]     (see https://errorprone.info/bugpattern/MixedMutabilityReturnType)
     11     [javac]     (see https://errorprone.info/bugpattern/LongDoubleConversion)
     11     [javac]     (see https://errorprone.info/bugpattern/DoNotCallSuggester)
     11     [javac]     (see https://errorprone.info/bugpattern/CatchAndPrintStackTrace)
     10     [javac]     (see https://errorprone.info/bugpattern/FutureReturnValueIgnored)
      9     [javac]     (see https://errorprone.info/bugpattern/UnusedMethod)
      9     [javac]     (see https://errorprone.info/bugpattern/SynchronizeOnNonFinalField)
      9     [javac]     (see https://errorprone.info/bugpattern/InlineMeInliner)
      8     [javac]     (see https://errorprone.info/bugpattern/InconsistentCapitalization)
      8     [javac]     (see https://errorprone.info/bugpattern/HidingField)
      6     [javac]     (see https://errorprone.info/bugpattern/URLEqualsHashCode)
      6     [javac]     (see https://errorprone.info/bugpattern/SuperEqualsIsObjectEquals)
      6     [javac]     (see https://errorprone.info/bugpattern/ReturnFromVoid)
      6     [javac]     (see https://errorprone.info/bugpattern/EscapedEntity)
      6     [javac]     (see https://errorprone.info/bugpattern/EqualsUnsafeCast)
      5     [javac]     (see https://errorprone.info/bugpattern/UnsynchronizedOverridesSynchronized)
      5     [javac]     (see https://errorprone.info/bugpattern/OrphanedFormatString)
      4     [javac]     (see https://errorprone.info/bugpattern/UnusedTypeParameter)
      4     [javac]     (see https://errorprone.info/bugpattern/UnusedNestedClass)
      4     [javac]     (see https://errorprone.info/bugpattern/TypeParameterShadowing)
      4     [javac]     (see https://errorprone.info/bugpattern/TypeEquals)
      4     [javac]     (see https://errorprone.info/bugpattern/ShortCircuitBoolean)
      4     [javac]     (see https://errorprone.info/bugpattern/OverrideThrowableToString)
      4     [javac]     (see https://errorprone.info/bugpattern/NarrowCalculation)
      4     [javac]     (see https://errorprone.info/bugpattern/LoopOverCharArray)
      4     [javac]     (see https://errorprone.info/bugpattern/IntLongMath)
      4     [javac]     (see https://errorprone.info/bugpattern/DateFormatConstant)
      4     [javac]     (see https://errorprone.info/bugpattern/DateChecker)
      3     [javac]     (see https://errorprone.info/bugpattern/UnicodeEscape)
      3     [javac]     (see https://errorprone.info/bugpattern/UndefinedEquals)
      3     [javac]     (see https://errorprone.info/bugpattern/ToStringReturnsNull)
      2     [javac]     (see https://errorprone.info/bugpattern/StaticAssignmentOfThrowable)
      2     [javac]     (see https://errorprone.info/bugpattern/ModifyCollectionInEnhancedForLoop)
      2     [javac]     (see https://errorprone.info/bugpattern/LabelledBreakTarget)
      2     [javac]     (see https://errorprone.info/bugpattern/JavaLangClash)
      2     [javac]     (see https://errorprone.info/bugpattern/IncrementInForLoopAndHeader)
      2     [javac]     (see https://errorprone.info/bugpattern/EqualsIncompatibleType)
      2     [javac]     (see https://errorprone.info/bugpattern/EmptyTopLevelDeclaration)
      2     [javac]     (see https://errorprone.info/bugpattern/ArgumentSelectionDefectChecker)
      1     [javac]     (see https://errorprone.info/bugpattern/WaitNotInLoop)
      1     [javac]     (see https://errorprone.info/bugpattern/ThreadPriorityCheck)
      1     [javac]     (see https://errorprone.info/bugpattern/InvalidThrows)
      1     [javac]     (see https://errorprone.info/bugpattern/InlineTrivialConstant)
      1     [javac]     (see https://errorprone.info/bugpattern/Finally)
      1     [javac]     (see https://errorprone.info/bugpattern/AlmostJavadoc)
```

The four DateFormatConstant issues are this one class, repeated twice. 

To repro this list:
```diff
diff --git a/common.ant.xml b/common.ant.xml
index d5e0312342..c10d87c16c 100755
--- a/common.ant.xml
+++ b/common.ant.xml
@@ -55,7 +55,7 @@
   <property name="javac.debuglevel" value="lines,vars,source"/>
   <property name="javac.encoding" value="utf-8"/>
   <property name="javac.release" value="11"/>
-  <property name="javac.nowarn" value="true"/>
+  <property name="javac.nowarn" value="false"/>
 
   <!-- javac and errorprone instructions from https://errorprone.info/docs/installation#ant -->
   <path id="errorprone.processorpath.ref">
@@ -184,6 +184,8 @@
         <compilerarg value="-J--add-exports=jdk.compiler/com.sun.tools.javac.util=ALL-UNNAMED" />
         <compilerarg value="-J--add-opens=jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED" />
         <compilerarg value="-J--add-opens=jdk.compiler/com.sun.tools.javac.comp=ALL-UNNAMED" />
+        <compilerarg value="-Xmaxwarns" />
+        <compilerarg value="99999" />
         <compilerarg value="-XDcompilePolicy=simple"/>
         <compilerarg value="-processorpath"/>
         <compilerarg pathref="mergedprocessorpath.ref"/>
```

jnehlmeier (changes requested):

user/test/com/google/web/bindery/autobean/vm/StringQuoterJreTest.java
niloc132: 1 for certain should be fixed - 2 could/should be combined with it, countdown as each thread is ready to run, then get() all threads before shutdown (then we shouldnt even need shutdown vs shutdownNow, since by definition all work is finished). 4 could spuriously succeed too if somehow we were slow enough to just not run through all the iterations (in GHA... i've seen stranger things, so that might actually be important).

3 seems important too if interrupts played a role in test code - though I wouldn't bet on us handling them appropriately across production compiler code, which may be acceptable for batch 
code. 

The good news is that the test fails without the fix, consistently. The latch bothered me at first read, but "it fails when wrong, and succeeds when obviously correct" is already a lot better than we had.

user/test/com/google/web/bindery/autobean/vm/StringQuoterJreTest.java
metsw24-max: Good call. Switched every tryParseDate argument to a string literal so the accepted formats are visible at the call site. They all point at the same instant (2009-02-13T23:31:30.123 UTC) with explicit +0000 offsets, so the expected Dates are deterministic regardless of the box's timezone.

user/test/com/google/web/bindery/autobean/vm/StringQuoterJreTest.java
metsw24-max: Reworked the harness along these lines:

1. The Futures are now collected and get() with a timeout after the start signal, so anything a worker throws (the NumberFormatException path included) is rethrown and fails the test instead of being lost.
2. ready is now a CountDownLatch(threads): each worker counts down and the main thread awaits it before releasing start, so all 16 are parked before the race begins.
3. InterruptedException in the worker is rethrown as a RuntimeException, which surfaces through Future.get().
4. Every await (ready, start, get) has a 60s timeout with a message, so a synchronization stall fails loudly rather than passing by accident.

Since we now get() every worker before shutting the pool down, all the work is guaranteed finished by then.

metsw24-max:
Bumping this one. The harness rework is in 9efec11 and covers jnehlmeier's four points along with zbynek's string-literal note. It still fails reliably without the ThreadLocal and stays clean with it, which lines up with what you saw locally.

From the errorprone thread: this change also clears the DateFormatConstant warnings for this class, since those are flagging the two shared SimpleDateFormat constants the PR replaces. Happy to make any further tweaks needed to land it.

user/test/com/google/web/bindery/autobean/vm/StringQuoterJreTest.java
```diff
@@ -0,0 +1,118 @@
+/*
+ * Copyright 2026 Google Inc.
```
jnehlmeier: Given it is a new file it should be `Copyright 2026 GWT Project Authors`.

jnehlmeier (changes requested):
Looks good. After the copyright change I am fine with it.

user/test/com/google/web/bindery/autobean/vm/StringQuoterJreTest.java
metsw24-max: Done, header now reads Copyright 2026 GWT Project Authors. Pushed in dbf5570.

zbynek (approved):

jnehlmeier (approved):

PR #10323, opened 2026-05-27
Improve incremental tracking of method references and lambdas

### LINKED ISSUE #9670: Constructor method reference cached incorrectly in superdevmode 2.8.2
**GWT version:** 2.8.2
**Browser (with version):** Google Chrome Version 76.0.3809.100 (Official Build) (64-bit)
**Operating System:** Ubuntu 16.04.2 LTS

---

##### Description

Constructor method references are cached in super dev mode in a way that changing the class is not recognized on subsequent recompiles.  Lambdas do not have the same issue.

##### Steps to reproduce

Execute the following code after a SDM compile:
final List<Object> l  = Stream.generate(Object::new).limit(3).collect(Collectors.toList());
GWT.log("list: " + l.toString());

The browser console will show:
list: [java.lang.Object@5a, java.lang.Object@5b, java.lang.Object@5c]

Now change the code to:
final List<Object> l  = Stream.generate(String::new).limit(3).collect(Collectors.toList());

Expected output is list: [, , ], however the browser reports list: [java.lang.Object@5a, java.lang.Object@5b, java.lang.Object@5c]

Clearing the cache and recompiling resolves the problem.

niloc132:
Ensures that all GWT JDeclaredTypes are correctly marked as changed, including method references and lambdas.

Changes naming convention for method reference types - this slightly degrades debugability, but better ensures that fresh compiles and incremental compiles will result in the same output for a given class, improving how much testing we can do.

Note that adding/removing inner classes results in different incremental/initial output, so one of the new tests must opt out of that check.

Fixes #9670


dev/core/src/com/google/gwt/dev/jjs/impl/UnifyAst.java
```diff
@@ -1030,11 +1030,11 @@ private void assimilateSourceUnit(CompilationUnit unit, boolean reportErrors) {
...
     // Staleness calculations need to be able to trace from CompilationUnit name to the names of
     // immediately nested types. So record those associations now.
-    if (incrementalCompile) {
-      compilerContext.getMinimalRebuildCache().recordNestedTypeNamesPerType(unit);
-    }
     // TODO(zundel): ask for a recompile if deserialization fails?
     List<JDeclaredType> types = unit.getTypes();
+    if (incrementalCompile) {
+      compilerContext.getMinimalRebuildCache().recordNestedTypeNamesPerType(unit, types);
+    }
```
niloc132, line 1037: This change lets us only deserialize the set of GWT types once. This is the only call to CompilationUnit.getTypes() and also the only call to recordNestedTypeNamesPerType.

dev/core/src/com/google/gwt/dev/MinimalRebuildCache.java
```diff
@@ -659,14 +658,23 @@ public void recordNestedTypeName(String compilationUnitTypeName, String nestedTy
...
-      recordNestedTypeName(compilationUnitTypeName, nestedTypeName);
+
+    // Record all GWT types that were derived from that compilation unit
+    for (JDeclaredType type : types) {
+      String typeName = type.getName();
+      if (!nestedTypeNamesByUnitTypeName.containsEntry(compilationUnitTypeName, typeName)) {
+        recordNestedTypeName(compilationUnitTypeName, typeName);
+      }
     }
   }
```
niloc132, line 676: THis pair of loops could probably be more efficient if we didn't remove-all and add-all, but just looked for the actual added/removed changes. GWT UI classes tend to have lambdas, anon inner classes, so this usually isn't a very short list, but it probably isn't long enough to really matter - but its also pretty likely most of the classes that were present last time are still there.

dev/core/src/com/google/gwt/dev/MinimalRebuildCache.java
```diff
@@ -659,14 +658,23 @@ public void recordNestedTypeName(String compilationUnitTypeName, String nestedTy
...
     // For the root type in the compilation unit the source name and binary name are the same.
     String compilationUnitTypeName = compilationUnit.getTypeName();
 
+    // Clean up the reverse map for old nested type names, then clear all entries
+    Collection<String> oldNestedTypeNames = nestedTypeNamesByUnitTypeName.get(compilationUnitTypeName);
+    for (String oldNestedTypeName : oldNestedTypeNames) {
+      compilationUnitTypeNameByNestedTypeName.remove(oldNestedTypeName);
+    }
     nestedTypeNamesByUnitTypeName.removeAll(compilationUnitTypeName);
-    for (CompiledClass compiledClass : compilationUnit.getCompiledClasses()) {
```
vegegoku, line 667: Maybe we can just remove in 2 steps instead like this 

```java
    Collection<String> oldNestedTypeNames = nestedTypeNamesByUnitTypeName.removeAll(compilationUnitTypeName);
    compilationUnitTypeNameByNestedTypeName.keySet().removeAll(oldNestedTypeNames);
```

dev/core/src/com/google/gwt/dev/MinimalRebuildCache.java
```diff
@@ -659,14 +658,23 @@ public void recordNestedTypeName(String compilationUnitTypeName, String nestedTy
...
-      String nestedTypeName = InternalName.toBinaryName(compiledClass.getInternalName());
-      recordNestedTypeName(compilationUnitTypeName, nestedTypeName);
+
+    // Record all GWT types that were derived from that compilation unit
+    for (JDeclaredType type : types) {
+      String typeName = type.getName();
+      if (!nestedTypeNamesByUnitTypeName.containsEntry(compilationUnitTypeName, typeName)) {
+        recordNestedTypeName(compilationUnitTypeName, typeName);
+      }
```
vegegoku: Since we know we removed in the previous step, do we still need this !exists check?

niloc132: You're right - I had an intermediate version which had the old loop and so this loop "needed" that check to avoid doing a little extra work, but if I can remove the old loop (it only did a subset of the work that this loop does), then this can unconditionally add each item.

niloc132:
This is back in draft to avoid merging until I've finished going over the new test added via https://github.com/gwtproject/gwt/pull/10390 and understanding the differences, possibly merging the fixes.

PR #10324, opened 2026-05-28
Add missing Override annotations

zbynek:
Used  ErrorProne's patch functionality to add override annotations where possible. Used regular expressions to make sure overrides are on a separate line.

This should make it easier to show errorprone warnings during compilation without going through much spam -- this is the most common errorprone as pointed out in https://github.com/gwtproject/gwt/pull/10322#issuecomment-4566968342.


.github/workflows/quick-check.yml
```diff
@@ -56,7 +56,7 @@ jobs:
           cd gwt
           for f in build/out/**/checkstyle*.xml ; do
             echo $f
-            reviewdog -f=checkstyle -filter-mode=diff_context -reporter=github-pr-annotations -level=info < $f
+            reviewdog -f=checkstyle -filter-mode=diff_context -reporter=github-pr-annotations -level=info < $f || true
```
zbynek, line 59: Huge PRs bound to run into limitation on the number of annotations.

niloc132:
I'm a little hesitant to fix these, like a lot of the other checkstyle issues
 - it isn't indicative of a problem, just not the style that we want
 - gwt was originally built for old enough java that override didn't exist (so the style wasn't even appropriate)
 - there isn't an actual bug being fixed here, just a missing annotation. This could be wrong in two cases - if one of these fails the build with the annotation (but it doesn't), or if the author meant to override a method but wasnt able to or didn't at the time and got the signature wrong (but we can't know what they meant).

From this I tend to say this is better to add in diffs that change related code, to ensure that the author of a given diff did what they expected.


I wonder if we can collect warnings that only apply to the current diff like we do with checkstyle and ensure we fix when we change things, rather than bulk changes like this? From https://github.com/reviewdog/reviewdog/issues/1087 it looks like this should be possible in theory (includes an example with errorprone), but it apparently doesn't work. (My [quick attempt to match errorprone+ant](https://reviewdog.github.io/errorformat-playground/?efms=%25W++++%5Bjavac%5D+%25f%3A%25l%3A+warning%3A+%25m%25Z%0A&text=++++%5Bjavac%5D+%2Fhome%2Fcolin%2Fworkspace%2Fgwt%2Fbuild%2Fout%2Fuser%2Fno-servlet-src%2Fcom%2Fgoogle%2Fweb%2Fbindery%2Fautobean%2Fshared%2Fimpl%2FSplittableSimpleMap.java%3A221%3A+warning%3A+%5BMissingOverride%5D+remove+implements+method+in+Iterator%3B+expected+%40Override%0A++++%5Bjavac%5D+++++++++++public+void+remove%28%29+%7B%0A++++%5Bjavac%5D+++++++++++++++++++++++%5E%0A++++%5Bjavac%5D+++++%28see+https%3A%2F%2Ferrorprone.info%2Fbugpattern%2FMissingOverride%29%0A++++%5Bjavac%5D+++Did+you+mean+%27%40Override+public+void+remove%28%29+%7B%27%3F%0A++++%5Bjavac%5D+%2Fhome%2Fcolin%2Fworkspace%2Fgwt%2Fbuild%2Fout%2Fuser%2Fno-servlet-src%2Fcom%2Fgoogle%2Fweb%2Fbindery%2Fautobean%2Fshared%2Fimpl%2FStringQuoter.java%3A33%3A+warning%3A+%5BDateFormatConstant%5D+DateFormat+is+not+thread-safe%2C+and+should+not+be+used+as+a+constant+field.%0A++++%5Bjavac%5D+++private+static+final+DateFormat+ISO8601+%3D+new+SimpleDateFormat%28ISO8601_PATTERN%2C+Locale%0A++++%5Bjavac%5D+++++++++++++++++++++++++++++++++++%5E%0A++++%5Bjavac%5D+++++%28see+https%3A%2F%2Ferrorprone.info%2Fbugpattern%2FDateFormatConstant%29%0A++++%5Bjavac%5D+++Did+you+mean+%27private+static+final+ThreadLocal%3CDateFormat%3E+iso8601+%3D+ThreadLocal.withInitial%28%28%29+-%3E+new+SimpleDateFormat%28ISO8601_PATTERN%2C+Locale%27+or+%27private+static+final+DateFormat+iso8601+%3D+new+SimpleDateFormat%28ISO8601_PATTERN%2C+Locale%27%3F%0A++++%5Bjavac%5D+%2Fhome%2Fcolin%2Fworkspace%2Fgwt%2Fbuild%2Fout%2Fuser%2Fno-servlet-src%2Fcom%2Fgoogle%2Fweb%2Fbindery%2Fautobean%2Fshared%2Fimpl%2FStringQuoter.java%3A37%3A+warning%3A+%5BDateFormatConstant%5D+DateFormat+is+not+thread-safe%2C+and+should+not+be+used+as+a+constant+field.%0A++++%5Bjavac%5D+++private+static+final+DateFormat+RFC2822+%3D+new+SimpleDateFormat%28RFC2822_PATTERN%2C+Locale%0A++++%5Bjavac%5D+++++++++++++++++++++++++++++++++++%5E%0A++++%5Bjavac%5D+++++%28see+https%3A%2F%2Ferrorprone.info%2Fbugpattern%2FDateFormatConstant%29%0A++++%5Bjavac%5D+++Did+you+mean+%27private+static+final+ThreadLocal%3CDateFormat%3E+rfc2822+%3D+ThreadLocal.withInitial%28%28%29+-%3E+new+SimpleDateFormat%28RFC2822_PATTERN%2C+Locale%27+or+%27private+static+final+DateFormat+rfc2822+%3D+new+SimpleDateFormat%28RFC2822_PATTERN%2C+Locale%27%3F%0A), untested via command line, but seems to capture enough for our purposes)

---

The flip side of course is that this is an obvious change (...after one reviews all 2126 changed lines), and if it compiles it must be correct.

If we want this to be required, the build should also be changed to promote this to a compile failure.

jnehlmeier:
> If we want this to be required, the build should also be changed to promote this to a compile failure.

At work we configure errorprone checks which default to warning as error if we prefer the general intent of that check. And yes we do bulk updates to fix these checks (especially when updating errorprone and new checks have been added). We thought about doing it incremental as we came across it but given the code base is large it would probably never come to an end. So these checks would need to default to warning forever. 

Fixing them and configure them as error improves code style consistency, makes code more readable, produces less noise during reviews and developers learn something if they run into such compile errors. If the build output is too verbose then warnings are not always noticed and they end up in reviews as well. At least that is my experience.

So in general I would be fine with that bulk change here.

niloc132:
Agreed, I'm not against it, just want to caution against bulk "I fixed every typo in the project" changes (or worse a deluge of small "fix: Removed final from a private method" PRs from autonomous agents that gotta get those stats up). Ant output is pretty gross already too - though we could turn it up specifically for CI to capture related warnings.

Override is sort of a special case too, since there are so many missing, more than 10x the next most common errorprone warning, about about 2x every other warning _combined_. If we do fix it _and_ turn up warnings overall, we're adding 2k or so more warnings in CI builds... though at least a few hundred of the next most common types are just a facet of GWT/JS and not going away any time soon.

But if it's important enough to fix, we should ratchet up warn to error, and at least start capturing/alerting other warnings so we can fix/ratchet or downgrade.

tbroyer:
> If we do fix it _and_ turn up warnings overall, we're adding 2k or so more warnings in CI builds... though at least a few hundred of the next most common types are just a facet of GWT/JS and not going away any time soon.

Then those probably should be disabled?

niloc132:
They all are disabled - `-nowarn` has been passed to javac via the ant build since the initial svn commit, nearly 20 years ago (though the line has moved a bit according to git, but the value has always been the same):

https://github.com/gwtproject/gwt/blob/4e6ae10e9798e73fa490dabc64b7a742e933a66b/common.ant.xml#L58

The greater context is that if you turn that back on (and ask javac to print more than a handful of warnings), you get all these EP issues. https://github.com/gwtproject/gwt/pull/10322#issuecomment-4566968342

Maybe you're saying "_if_ we do this change and _if_ we turn on those other warnings and _if_ we somehow don't make any other changes to isolate/permit accepted issues" then we should disable those checks? But the premise is "let's enable appropriate checks" so I think that's a given?

jnehlmeier:
> Agreed, I'm not against it, just want to caution against bulk "I fixed every typo in the project" changes (or worse a deluge of small "fix: Removed final from a private method" PRs from autonomous agents that gotta get those stats up). 

Yes, but we could also settle on some contribution guidelines saying that such bulk updates are generally only accepted from core members with the goal to make better use of errorprone.

> If we do fix it _and_ turn up warnings overal

At work we have javac warnings turned on (with some javac warnings excluded) and we disabled ALL errorprone checks which default to warning and showed up when we introduced errorprone or when we update errorprone so we do not get overwhelmed. Then we turn them on in errorprone one by one, fix them and switch them to error.

So yes we should not just remove `-nowarn`. Errorprone need to be configured first (I don't know out of my head if there is any ep configuration in the ant script with regard to checks).

tbroyer:
I was specifically talking about those “at least a few hundred of the next most common types [that] are just a facet of GWT/JS”: if we enable warnings (as a whole), then disable those "next most common types" that we don't care about (we've ignored them for all these years, so even if there are "real positives" there, it might not be worth getting the spam of the ones we consider "false positives").

And BTW, if we otherwise don't care at all about warnings (any warning, like today), then we should probably use [`-epDisableAllWarnings`](https://errorprone.info/docs/flags) in addition to `-nowarn` to avoid the work of detecting those warnings and then ignore them anyway (technically, there could be "mandatory warnings" but Error Prone don't produce any of them).

We could also `-XepDisableAllWarnings -Xep:MissingOverride:WARN` or some other combination (e.g. bumping to error).

zbynek:
With the last commit the most common patterns are suppressed + reviewdog should  report the less common ones as comments (we can use separate invocation of reviewdog since we only care about annotations and no merged summary).

Also `MissingOverride` is bumped to trigger an error from now on.

.github/workflows/quick-check.yml
```diff
@@ -56,8 +56,9 @@ jobs:
...
           done
+          reviewdog -efm="    [javac] %f:%l: warning: %m" -filter-mode=diff_context -reporter=github-pr-annotations -level=info < log.txt || true
```
niloc132, line 61: Could you introduce a deliberate EP warning so we can validate that this is working?

dev/core/src/com/google/gwt/dev/shell/remoteui/RemoteMessageProto.java
```diff
@@ -3,6 +3,9 @@
 
 package com.google.gwt.dev.shell.remoteui;
 
+import javax.annotation.processing.Generated;
+
+@Generated("remotemessage.proto")
```
niloc132, line 8: Teeeeeechnically protobuf uses the javax.annotation.Generated annotation, though now they also use `@com.google.protobuf.Generated`

https://github.com/protocolbuffers/protobuf/commit/31e57fc64a87d52a57d3373d7088c60a84ae79ff

This isn't worth changing, since our version of protobuf is so old (and can't change, since Google never open sourced the build for the dev mode plugins...), just pointing it out.

common.ant.xml
```diff
@@ -161,7 +161,7 @@
     <attribute name="release" default="${javac.release}" />
     <attribute name="excludes" default="" />
     <attribute name="processorpath" default="" />
-    <attribute name="errorprone.args" default="" />
+    <attribute name="errorprone.args" default="-XepDisableWarningsInGeneratedCode -xEp:MissingOverridde:ERROR -Xep:UnrecognisedJavadocTag:OFF -Xep:InvalidBlockTag:OFF -Xep:AnnotateFormatMethod:OFF -Xep:ProtectedMembersInFinalClass:OFF -Xep:TypeParameterUnusedInFormals:OFF -Xep:InvalidLink:OFF -Xep:NotJavadoc:OFF -Xep:OperatorPrecedence:OFF -Xep:IdentityHashMapUsage:OFF -Xep:DefaultCharset:OFF -Xep:StringSplitter:OFF -Xep:BadImport:OFF -Xep:UnusedVariable:OFF -Xep:JdkObsolete:OFF -Xep:SameNameButDifferent:OFF -Xep:InlineMeSuggester:OFF -Xep:UnnecessaryParentheses:OFF -Xep:NonApiType:OFF -Xep:InvalidInlineTag:OFF -Xep:ReferenceEquality:OFF -Xep:JavaUtilDate:OFF -xEp:MissingSummary:OFF -xEp:EmptyBlockTag:OFF" />
```
niloc132: It seems that this is selectively disabling some options, but by doing so in errorprone.args, the suggestion seems to be that we want to re-enable them for certain projects. Instead however, only user/build.xml overrides this, by disabling all warnings.

Wouldn't it make more sense to move these to the compilerarg tags where we want them disabled globally? Otherwise, if we were to, for example, attempt to selectively disable one more check in dev/build.xml, we would end up re-enabling all of the rest? Or, something like only put `-XepDisableWarningsInGeneratedCode -xEp:MissingOverridde:ERROR` here in a compiler arg, then leave user disabling all warnings (except that), and leave the rest of these as warnings until we fix them as we interact with the code?

No firm thoughts here except these are all _potential_ bugs.

zbynek: I didn't really consider using different settings for different modules. Maybe we can look into that once we need it?

The only override is in the no-servlet compilation (`user/build.xml`, as you pointed out) that prevents all issues from `user` module to be reported *twice* (technically there could be some issues specific to Jakarta, but IMHO not worth the overhead).

The selection of things to ignore is based on number of reports which is not a great metric, but should avoid having too many annotations in PRs. Follow-up tickets can be made to reduce this list.

zbynek:
<details>
<summary>Reported warnings with the current filter:</summary>

```
     1 AmbiguousMethodReference
      1 ExtendsObject
      1 IncorrectMainMethod
      1 InputStreamSlowMultibyteRead
      1 MockNotUsedInProduction
      1 MutablePublicArray
      1 StaticGuardedByInstance
      2 AlmostJavadoc
      2 AssertEqualsArgumentOrderChecker
      2 EmptyTopLevelDeclaration
      2 Finally
      2 InlineTrivialConstant
      2 LogicalAssignment
      2 MissingImplementsComparable
      2 MultipleUnaryOperatorsInMethodCall
      2 NarrowingCompoundAssignment
      2 StaticAssignmentOfThrowable
      2 ThreadPriorityCheck
      2 TypeNameShadowing
      2 WaitNotInLoop
      3 UnnecessaryStringBuilder
      4 DateFormatConstant
      4 IncrementInForLoopAndHeader
      4 LabelledBreakTarget
      4 ModifyCollectionInEnhancedForLoop
      4 NonCanonicalType
      4 ToStringReturnsNull
      5 EqualsIncompatibleType
      5 NegativeCharLiteral
      5 TypeEquals
      5 UnusedTypeParameter
      6 ArgumentSelectionDefectChecker
      6 EscapedEntity
      6 JavaLangClash
      6 LoopOverCharArray
      6 ReturnFromVoid
      6 SuperEqualsIsObjectEquals
      7 HidingField
      7 InvalidThrows
      7 ShortCircuitBoolean
      7 TimeUnitConversionChecker
      8 EqualsUnsafeCast
      8 NarrowCalculation
      8 URLEqualsHashCode
      9 ComplexBooleanConstant
      9 OverrideThrowableToString
      9 TypeParameterShadowing
     10 OrphanedFormatString
     11 BigDecimalEquals
     11 InconsistentCapitalization
     11 IntLongMath
     11 StringCaseLocaleUsage
     12 ModifiedButNotUsed
     13 DefaultPackage
     13 DoNotCallSuggester
     13 InlineFormatString
     13 LongDoubleConversion
     13 removal
     14 OutlineNone
     15 UnusedNestedClass
     16 FutureReturnValueIgnored
     16 InlineMeInliner
     17 UnicodeEscape
     18 SynchronizeOnNonFinalField
     20 BadInstanceof
     20 MixedMutabilityReturnType
     21 UnsynchronizedOverridesSynchronized
     23 FloatingPointLiteralPrecision
     23 ReturnAtTheEndOfVoidFunction
     24 MissingFail
     25 AssertionFailureIgnored
     25 CatchAndPrintStackTrace
     31 CollectionUndefinedEquality
     31 EqualsGetClass
     31 ImmutableEnumChecker
     32 MissingCasesInEnumSwitch
     33 StaticAssignmentInConstructor
     35 CatchFail
     35 StringCharset
     43 ClassNewInstance
     43 InvalidParam
     44 FallThrough
     45 UndefinedEquals
     61 UnnecessaryAsync
     84 UnusedMethod
     94 EmptyCatch
    136 ClassCanBeStatic
 ```
    
   </details>

**Comment by niloc132** [comment]
Looking at https://github.com/gwtproject/gwt/pull/10348, it doesnt look like they are being reported in the PR still?

**Comment by zbynek** [comment]
> it doesnt look like they are being reported in the PR still?

@niloc132 now it worked (I forgot reviewdog only considers the changed parts of the file), but it reports it twice

File: user/super/com/google/gwt/emul/java/util/logging/Level.java [hunk 5]
```diff
@@ -37,48 +37,71 @@ public class Level implements Serializable {
...
+    }
   }
 
   private static class LevelConfig extends Level {
-    @Override public String getName() { return "CONFIG"; }
-    @Override public int intValue() { return 700; }
+    @Override
+    public String getName() { return "CONFIG"; }
+    @Override
+    public int intValue() { return 700; }
```
niloc132: Cange these consistently? If its worth fixing, its probably worth fixing consistently, so we don't get another pass to fix again.

user/test/com/google/gwt/emultest/java/lang/FloatTest.java
```diff
@@ -22,6 +22,7 @@
  * Unit tests for the Javascript emulation of the Float/float autoboxed
  * fundamental type.
  */
+@SuppressWarnings("FloatingPointLiteralPrecision")
```
niloc132: wrong patch?

user/test/com/google/gwt/editor/rebind/model/EditorModelTest.java
```diff
@@ -96,6 +96,18 @@ public CharSequence getContent() {
...
 
+  private static class OverrideStrippedJavaResource  extends RealJavaResource {
+
+    public OverrideStrippedJavaResource(Class<?> clazz) {
+      super(clazz);
+    }
+
+    @Override
+    public CharSequence getContent() {
+      return super.getContent().toString().replace("@Override", "");
```
niloc132: Why does RFED need to have the overrides removed, but nothing else does?

user/test/com/google/gwt/junit/client/DefaultExceptionAsserter.java
```diff
@@ -26,7 +26,7 @@
 public class DefaultExceptionAsserter extends Assert implements ExceptionAsserter {
 
   @GwtIncompatible
-  @Override
+  @SuppressWarnings("MissingOverride")
```
niloc132: Why remove the override and replace with the suppression?

user/build.xml
```diff
@@ -228,6 +228,7 @@
               release="8"
               srcdir="${project.build}/jakarta-src"
               destdir="${javac.out}-jakarta"
+              errorprone.args="-XepDisableAllWarnings"
```
niloc132, line 234: Does it make sense to suppress every warning here, when this patch just fixed a ton of override warnings? As written, I believe this is replacing (rather than appending to) the defaults provided in common.ant.xml - maybe common should always add those flags rather than make them defaults, then user can add this one single flag?

niloc132 (commented):
Sorry it took me a while, but i've read it all now. Just a few comments.

user/test/com/google/gwt/junit/client/DefaultExceptionAsserter.java
zbynek: I added the suppression to other `ExceptionAsserter` and pasted it here without much thinking. Actually the better fix is to have `@GwtIncompatible` in all the asserters.

user/build.xml
zbynek: I added a comment now explaining that this just prevents it from printing them twice.

user/test/com/google/gwt/editor/rebind/model/EditorModelTest.java
zbynek: Because the source of `EditorDriver` is modified (mocked), RFED source cannot contain overrides of methods that were hidden. With my last commit I tried the opposite approach though, to make it clear which method is causing the issue.

user/build.xml
niloc132: Oh got it, that makes sense, thanks for adding the comment.

niloc132 (approved):

PR #10326, opened 2026-05-31
fix off-by-one out-of-bounds write in EmulatedCharset.getBytes

metsw24-max:
Noticed LatinCharset.getBytes writes bytes[i] while i runs from offset, but bytes is only count long. So OutputStreamWriter.write(buf, offset, count) with a non-zero offset on ISO-8859-1/Latin-1 writes past the output array and leaves bytes[0..offset) unwritten; the precondition check only validates offset/count against the input buffer. Index the destination with i - offset.

### DIFF
--- a/user/super/com/google/gwt/emul/javaemul/internal/EmulatedCharset.java
+++ b/user/super/com/google/gwt/emul/javaemul/internal/EmulatedCharset.java
@@ -38,7 +38,7 @@ public abstract class EmulatedCharset extends Charset {
       int n = offset + count;
       byte[] bytes = new byte[count];
       for (int i = offset; i < n; ++i) {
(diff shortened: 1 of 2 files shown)


niloc132:
Thanks - at a quick glance that looks correct, and confirmed that the test fails without the change (and passes with it). 

The fork Google made also appears to need this fix: https://github.com/google/j2cl/blob/master/jre/java/javaemul/internal/EmulatedCharset.java

metsw24-max:
Good point, the LatinCharset code is identical over there. I'll send the same one-line fix to j2cl.

zbynek (approved):

metsw24-max:
gentle ping

PR #10328, opened 2026-06-02
validate permutation strong name before building symbol map paths

metsw24-max:
The permutation strong name comes straight from the client `X-GWT-Permutation` header and is concatenated into the symbol/source map file name that `openInputStream` resolves:

    resymbolize(trace, "../../../../etc/passwd")
    -> openInputStream("../../../../etc/passwd.symbolMap")

`ServerSerializationStreamReader` already restricts the strong name to `[a-zA-Z0-9_]+` before loading the policy file; this applies the same check inside the symbol/source map loaders so an out-of-range name is treated as not found.

### DIFF
--- a/user/test/com/google/gwt/core/server/StackTraceDeobfuscatorTest.java
+++ b/user/test/com/google/gwt/core/server/StackTraceDeobfuscatorTest.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2026 GWT Project Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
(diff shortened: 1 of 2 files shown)


user/test/com/google/gwt/core/server/StackTraceDeobfuscatorTest.java
```diff
@@ -0,0 +1,56 @@
+/*
+ * Copyright 2026 Google Inc.
```
niloc132: GWT Project Authors

metsw24-max:
gentle ping

niloc132 (commented):
Sorry, I had started a review some time ago, but didn't finish it. Just one nit (unless you work for Google and want to keep the attribution like that)

user/test/com/google/gwt/core/server/StackTraceDeobfuscatorTest.java
metsw24-max: Good catch, I'm not at Google, so I've changed it to GWT Project Authors to match the other files.

Samin061:
This is the same issue I filed in #10356 (now closed), with the same [a-zA-Z0-9_]+ check mirroring ServerSerializationStreamReader, so no objection to keeping just this one.

One minor point from my version: putting the guard directly in getSymbolMapInputStream/getSourceMapInputStream (where the name is actually concatenated into the path) keeps the check at a single choke point, so the source map fragment path and any future caller of those protected methods are covered by construction. Your placement in loadSymbolMap/loadSourceMap covers both current call sites already, so it's really just a defense-in-depth preference. Either way the traversal is closed.

niloc132 (approved):

PR #10330, opened 2026-06-03
Fix empty ID assertion, simplify equality check

zbynek:
This fixes two issues found by nullness analysis:
* the assertion of non-empty IDs for ARIA purposes was logically wrong. When `Id` is created from an element, the `id` attribute is never null and the empty string check was skipped. When `Id` is created from a string (never happens in GWT itself) and the string happens to be `null`, the assertion would result in NPE rather than assertion error.
* the other issue is with slightly inefficient re-implementation of `Objects.equals` which had an unused `null` check (I assume the API didn't exist in JDK back when this wasimplemented, but can be used now).

### DIFF
--- a/user/src/com/google/gwt/cell/client/EditTextCell.java
+++ b/user/src/com/google/gwt/cell/client/EditTextCell.java
@@ -84,7 +84,7 @@ public class EditTextCell extends
 
     @Override
     public boolean equals(Object o) {
-      if (!(o instanceof  ViewData)) {
+      if (!(o instanceof ViewData)) {
         return false;
       }
       ViewData vd = (ViewData) o;
--- a/user/src/com/google/gwt/cell/client/EditTextCell.java
+++ b/user/src/com/google/gwt/cell/client/EditTextCell.java
@@ -33,6 +33,7 @@ import com.google.gwt.text.shared.SafeHtmlRenderer;
 import com.google.gwt.text.shared.SimpleSafeHtmlRenderer;
 
 import java.util.Locale;
+import java.util.Objects;
 
 /**
  * An editable text cell. Click to edit, escape to cancel, return to commit.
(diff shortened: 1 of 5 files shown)


user/src/com/google/gwt/cell/client/EditTextCell.java
```diff
@@ -83,12 +84,12 @@ public ViewData(String text) {
 
     @Override
     public boolean equals(Object o) {
-      if (o == null) {
+      if (!(o instanceof  ViewData)) {
```
niloc132:
```suggestion
      if (!(o instanceof ViewData)) {
```

user/src/com/google/gwt/cell/client/EditTextCell.java
```diff
@@ -83,12 +84,12 @@ public ViewData(String text) {
...
         return false;
       }
       ViewData vd = (ViewData) o;
-      return equalsOrBothNull(original, vd.original)
-          && equalsOrBothNull(text, vd.text) && isEditing == vd.isEditing
+      return Objects.equals(original, vd.original)
```
niloc132, line 91: alternatively, use [pattern match instanceof](https://openjdk.org/jeps/394) in the return here, something like

```suggestion
      return o instanceof ViewData vd 
          && Objects.equals(original, vd.original)
```

zbynek: Do we want to start using Java 17 features?

niloc132: Sorry right, we might not want to do that yet. I had assumed that because the compiler could handle it (even on Java 8) that we would want it, but that might not make sense after all.

niloc132 (approved):

PR #10335, opened 2026-06-07
Remove commons-collections from classpath

### LINKED ISSUE #10334: Update old dependency on commons-collections
`gwt-dev` has a dependency on `commons-collections`. Users are thus also forced to transitively pull in that dependency, which is bad for the following reasons:
- The last release was in 2015. Since then, it has been replaced with `commons-collections4` and users should migrate.
- Nexus IQ, Sonatype's commercial security scanner, reports a vulnerability sonatype-2024-3350 in that library with CVSS 8.7.

Image

Recommended actions:
- Check if the dependency is needed at all. If not, remove it.
- If it is needed, update to `commons-collections4`.

niloc132:
Only seems to be used for testing GWT's own collection types.

Fixes #10334


zbynek (dismissed):

dev/build.xml
```diff
@@ -83,7 +84,6 @@
           <include name="tomcat/tomcat-servlet-api-8.0.28.jar"/>
           <include name="tomcat/tomcat-websocket-api-8.0.28.jar"/>
           <include name="tomcat/tomcat-annotations-api-8.0.28.jar"/>
-          <include name="apache/commons/commons-collections-3.2.2.jar"/>
```
zbynek, line 86: These classes are still referenced in `com.google.gwt.dev.util.collect.Hash(Map|Set)`'s Javadoc, those references should be changed to http links or `@code`.

zbynek (dismissed):

zbynek (approved):

PR #10336, opened 2026-06-08
Raise Java language level to 17, update errorprone, asm

niloc132:
ErrorProne version is raised to the newest release that still supports running on Java 17. There are however two bugs in EP that cause it to fail when compiling GWT with this release - these are fixed in 2.45, so a workaround is applied for now. Several classes were also updated to address new EP violations.

The asm version is also updated to latest, and confirmed to work on Java 17, and should be forward compatible to Java 27.


niloc132:
Depends on https://github.com/gwtproject/tools/pull/39

niloc132:
Converted to draft while moving new APIs to triage and waiting for tools merge.

niloc132:
CollectReferencesVisitorTest failure is due to asm version being 9.6 - the test only passes up to Java 22. This doesn't impact the compiler actually working, but is a failure specific to the test reading Object.clas.

CompilationUnitTypeOracleUpdaterFromByteCodeTest has a similar problem, though it is designed to read bytecode from JDT's output, which is currently limited to a max of java 17. The test however still grabs Object.class, as above, which will be specific to whatever JRE is actually running the test.

zbynek (changes requested):
(Just noting that the tools repo branch reference needs to be reverted here.)

Otherwise looks good 👍

zbynek (approved):

PR #10344, opened 2026-06-15
Allow overriding Super Dev Mode server URL via prompt

schteff:
If the SDM server is unreachable, a prompt now allows users to specify a new server URL. This URL is persisted in sessionStorage so it can be used during the subsequent page reload.

This makes it possible to use https://github.com/localtunnel/localtunnel or other similar services (that tunnel specific ports) to share the local dev environment with other people without having to build and deploy the app on some other server.

### DIFF
--- a/dev/codeserver/java/com/google/gwt/dev/codeserver/stub.nocache.js
+++ b/dev/codeserver/java/com/google/gwt/dev/codeserver/stub.nocache.js
@@ -20,8 +20,19 @@
 (function($wnd, $doc){
   // Compute some codeserver urls so as the user does not need bookmarklets
   var hostName = $wnd.location.hostname;
-  var serverUrl = 'http://' + hostName + ':__SUPERDEV_PORT__';
   var module = '__MODULE_NAME__';
+
+  // The default Super Dev Mode server url. The user may override it (and retry)
+  // through the prompt shown by the error handler below; the override is kept
+  // for the session so the reload picks up the new server.
+  var serverUrlKey = '__gwt_sdm__server_url';
+  var defaultServerUrl = 'http://' + hostName + ':__SUPERDEV_PORT__';
+  var serverUrl = defaultServerUrl;
+  try {
+    serverUrl = $wnd.sessionStorage.getItem(serverUrlKey) || defaultServerUrl;
+  } catch (e) {
+    // sessionStorage may be unavailable; fall back to the default url.
+  }
   var nocacheUrl = serverUrl + '/recompile-requester/' + module;
 
   // Insert the superdevmode nocache script in the first position of the head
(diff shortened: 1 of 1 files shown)


dev/codeserver/java/com/google/gwt/dev/codeserver/stub.nocache.js
```diff
@@ -20,8 +20,19 @@
...
+  // The default Super Dev Mode server url. The user may override it (and retry)
+  // through the prompt shown by the error handler below; the override is kept
+  // for the session so the reload picks up the new server.
+  var serverUrlKey = '__gwt_sdm__server_url';
+  var defaultServerUrl = 'http://' + hostName + ':__SUPERDEV_PORT__';
+  var serverUrl = defaultServerUrl;
+  try {
+    serverUrl = $wnd.sessionStorage.getItem(serverUrlKey) || defaultServerUrl;
+  } catch (e) {
+    // sessionStorage may be unavailable; fall back to the default url.
```
niloc132: What's the risk/likelihood that session storage is unavailable? `dev_mode_on.js` relies on it and won't function without it, so I think maybe we can assume it is there?

dev/codeserver/java/com/google/gwt/dev/codeserver/stub.nocache.js
```diff
@@ -32,15 +43,22 @@
...
             "server at " + serverUrl + ".\n" +
             "Please make sure this server is ready.\n" +
-            "Do you want to try again?")) {
+            "Press OK to try again:",
+            serverUrl);
+        if (newServerUrl) {
+          try {
+            $wnd.sessionStorage.setItem(serverUrlKey, newServerUrl);
+          } catch (e) {
+            // Without storage the reload falls back to the default url.
```
niloc132: (assuming still relevant) Should probably notify that the prompt had no effect?

Maybe test for storage usage before attempting to use prompt to begin with?

niloc132 (commented):
I like the idea, and at a quick glance the implementation seems okay - but doesn't the bookmarklet setup solve this too? That is, if you first start mysdm:9876 and visit that URL, the bookmarklets let you enable SDM in session storage?

My current usage is atypical, so I may well be misremembering.

schteff:
> I like the idea, and at a quick glance the implementation seems okay - but doesn't the bookmarklet setup solve this too? That is, if you first start mysdm:9876 and visit that URL, the bookmarklets let you enable SDM in session storage?
> 
> My current usage is atypical, so I may well be misremembering.

I haven't used the bookmarklets in so long, I thought they were abandoned with the old dev mode and replaced by SDM.

I'll remove the check for sessionstorage availability, I was not aware of it being used for other GWT functionality. 👍

niloc132:
They aren't necessary, but they are allowed to be used, and can give you control like what you're describing - arguably better to be able to enable/disable and control urls directly through the click of a button than prompts and manual typing (when it might just be that your server is stopped and you forgot when you hit refresh).

schteff:
Just to clarify, the original URL is pre-filled in the prompt. Just clicking OK on the prompt will yield the exact same result as clicking OK on the original alert. Manual typing would only be needed if you would want to change the default URL. And since it's persisted it's only needed once (if typed correctly).

I haven't tried the bookmarklet variant yet, but an alternative solution for this PR might be to simply change the message in the original alert to give the user the tip of using the bookmarklet if they want to manually change the URL?

jnehlmeier:
@schteff If you try the bookmarklets then you might need to adjust the gwt property `devModeUrlWhitelistRegexp` defined in `CrossSiteIframeLinker.gwt.xml`. By default GWT checks for `127.0.0.1 / localhost`.

zbynek:
Another approach would be to put the right URL of the code server into the JS similar to this proposal:
https://gwt-review.googlesource.com/c/gwt/+/17180

(except instead of the binding address, a new setting would be used to allow use of local tunnel)

PR #10346, opened 2026-06-17
Pin GitHub actions to SHA hashes

zbynek:
Pointing to mutable tags is bad, because they can be force-pushed by repository owners or bad actors. All the tags GWT is currently referencing are mutable (e.g. list of [reviewdog](https://github.com/reviewdog/action-setup/tags) tags shows the distinction; for [setup-java](https://github.com/actions/setup-java/tags) all tags are mutable). The risk is [well known](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions
), but [until recently](https://github.com/aquasecurity/trivy/security/advisories/GHSA-69fq-xp46-6x23) it wasn't exploited for widely used actions. Now might be a good time to fix this.

This PR replaces all tags with hashes (even if immutable tag would be available, for consistency). The hash + comment format is well [supported by Dependabot](https://docs.github.com/en/actions/reference/security/secure-use#keeping-the-actions-in-your-workflows-secure-and-up-to-date).

### DIFF
--- a/.github/workflows/full-check.yml
+++ b/.github/workflows/full-check.yml
@@ -27,17 +27,17 @@ jobs:
         java-version: [ '11', '17', '21', '22' ]
     steps:
       - name: Checkout GWT itself into one directory
-        uses: actions/checkout@v6
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #v6.0.3
         with:
           path: 'gwt'
       - name: Checkout GWT tools into a sibling directory
-        uses: actions/checkout@v6
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 #v6.0.3
         with:
           repository: 'gwtproject/tools'
           path: 'tools'
       - name: Set up JDK ${{ matrix.java-version }}
         # GWT requires Java 11+ to build
-        uses: actions/setup-java@v5
+        uses: actions/setup-java@ad2b38190b15e4d6bdf0c97fb4fca8412226d287 #v5.3.0
         with:
           java-version: ${{ matrix.java-version }}
           distribution: 'temurin'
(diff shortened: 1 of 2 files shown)


niloc132:
This "broken by design" attitude of GHA almost has me wanting to leave github by itself - tags are mutable (WHY? git itself assumes they aren't, you have to go out of your way to pull moved tags, unlike branches), and as a result we "should" pin to repo/action@sha tuples... but that is flawed by design too - you can reference a hash that isn't part of `repo/action`, just like you can for git commits - e.g. `https://github.com/gwtproject/gwt/commit/a69f715a28a9ab0bcad8bce94630d037b19784b8` is a commit in my fork, but i can reference it via `gwtproject/gwt`. When you craft such a URL, at least you get a warning banner at the top, but using the wrong action sha lets you pull commits from any fork of the action. Contrast this to tags, which _must_ come from the original repo (EDIT: this is faulty too, see https://www.stepsecurity.io/blog/actions-cool-issues-helper-github-action-compromised-all-tags-point-to-imposter-commit-that-exfiltrates-ci-cd-credentials)... but can move. Tags also have the benefit of being autoupdated by dependabot (EDIT: I'm wrong here too, as you called out at the end of your commit message).

GIthub's docs tell you about this in a rather opaque way https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/find-and-customize-actions#using-shas
> When selecting a SHA, you should verify it is from the action's repository and not a repository fork. 
Any action change then needs to be manually verified to actually be the tag it claims to be - but since tags move, it might not be correct for long? 

How on earth is this how they intended this feature to work?

There is apparently upcoming work to let tags be used while forcing them to be pinned to a commit, which seems like the best option. It also locks upstream dependencies, to avoid a given action from itself being attacked in this way, which is definitely a step forward.
https://github.blog/news-insights/product-news/whats-coming-to-our-github-actions-2026-security-roadmap/

Reviewing https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax, this apparently isn't complete yet.

zbynek:
I know it's not ideal, but it's what a lot of projects are doing and out of the options that are available right now it seems to be the safest 🤷

niloc132 (approved):
Manually verified all hashes match the repo+tag they reference

PR #10349, opened 2026-06-22
Add support for Bazel 9 to the j2cl parts

aschleck:
Adds a MODULE.bazel and a @rules_java dep to pull in `java_library`. Bazel complains that the javax glob is empty but I thought it was best to not mess with the list for consistency

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -1,0 +1,3 @@
+module(name = "org_gwtproject_gwt")
+
+bazel_dep(name = "rules_java", version = "9.6.1")
--- a/user/BUILD.bazel
+++ b/user/BUILD.bazel
@@ -1,5 +1,8 @@
 # Description:
 #   subset of bazel rules used by j2cl.
+
+load("@rules_java//java:defs.bzl", "java_library")
+
 package(default_visibility = ["//visibility:public"])
 
 licenses(["notice"]) # Apache2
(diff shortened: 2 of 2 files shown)


niloc132:
Are you continuing to use the bazel parts? They haven't been maintained, and J2CL doesn't use GWT any more (directly, they forked the sources that the use). If not, we're intended to remove them sooner than later.

aschleck:
Thank you for looking at this! My problem is that j2cl depends on elemental2 and elemental2 depends on this via https://github.com/google/elemental2/blob/5f4a508c7d1235edd1eac6e8f887ff607c3080d3/MODULE.bazel#L47 for gwt-array-stamper  https://github.com/google/elemental2/blob/5f4a508c7d1235edd1eac6e8f887ff607c3080d3/java/elemental2/core/BUILD#L63 and https://github.com/google/elemental2/blob/5f4a508c7d1235edd1eac6e8f887ff607c3080d3/third_party/BUILD#L327 so the build fails at analysis time. If this dep can be deleted from elemental2 that sounds good to me!

aschleck:
@niloc132 do you have any thoughts on what I should do here?

`javaemul.internal.ArrayStamper` is used in the generated(?) `elemental2.core.JsArray#asArray`:

```
  @JsOverlay
  public final T[] asArray(T[] reference) {
    return ArrayStamper.stampJavaTypeInfo(this, reference);
  }
```

zbynek:
Closing in favor of https://github.com/google/elemental2/issues/188

Issue #9881, opened 2023-12-20
Compile-time configuration to disable rpc.enhancedClasses feature


niloc132:
Following https://github.com/gwtproject/gwt/issues/9709 and the PR https://github.com/gwtproject/gwt/pull/9879 to mitigate this by disabling the feature, this issue is to track efforts to entirely disable it. This would allow projects to safely leave JDO/JPA annotations in their DTO sources, without needing to filter or sign/verify requests and responses.

At GWT compile time, some hint would tell the GWT-RPC codegen to not only leave the `@ClientFields` code out of the serialization policy, but avoid generating client code to handle those enhanced payloads. This should amount to little more than ensuring that `JClassType.isEnhanced()` or `TypeInfoComputed.maybeEnhanced()` always return false in these cases.

Disabling this could be by annotations on the DTO type itself, on the RemoteService type that the policy file would pertain to, or perhaps globally, with something like a "deny" to match the "allow" that the `rpc.enhancedClasses` configuration property currently allows (perhaps with wildcards, rather than excluding on a class-by-class basis). 

By excluding the `ClientFields` directive entirely from the serialization policy file, the changes made in #9879 would never need to warn/error about these types at runtime.

Issue #9880, opened 2023-12-20
Server-side validate for rpc.enhancedClasses payloads


niloc132:
Following https://github.com/gwtproject/gwt/issues/9709 and the PR https://github.com/gwtproject/gwt/pull/9879 to mitigate this by disabling the feature, this issue is to track efforts to make this feature safe to use again. There are several approaches that can be used to make this safe, and reasons that no one solution is a guaranteed fit for all applications, so I propose that we find a general extension point to add, and if populated by an application, permit the use of enhanced classes without warning or error.

Basic approaches:
 1. Sign serialized data with a trusted private key and serialize the signature+public key with the data. Then, when reading the data back, the server can confirm the public key is trusted and that the signature is correct. A single server might create an in-memory key on startup (no persistence/coordination required), but then server restarts mean that clients must re-fetch all entities. A persisted key resolves this issue, at the cost of some additional configuration. A cluster of servers performing any kind of load balancing or handoff that allows entities from one server to be passed to another further requires coordination of keeping separate private keys but sharing trusted public keys (or using a common CA/etc), or all servers sharing the same keypair.
 2. Authenticate first before deserializing. This assumes that no authenticated user could be malicious, not usually an acceptable assumption. `onBeforeRequestDeserialized` can be overridden today to read from the request or session to validate in this way.
 3. Allow developer control/configuration of the `ObjectInputStream` and surrounding code. In some cases it could be sufficient to offer a limited set of allowed classes that ObjectInputStream.resolveClass could read, and to check each one as it is read that it matches the expected field to write. See `ServerSerializationStreamReader.checkResolve` for the current implementation here.

Option 1 seems to be the superior choice in terms of granting flexibility and not requiring trusting malicious clients, but it also likely requires the most flexible configuration. We could go so far as to delegate this as a simple interface such as 

```java
public interface SharedClassKeyValidation {
  /**
   * Given a plain payload, returns a signed payload that will be later accepted by verifyAndUnwrap.
   * It must always be true that verifyAndUnwrap(signAndAppend(payload)) returns the original payload.
  String signAndAppend(String payload);
  /**
   * Given a signed payload, confirms the signature is valid and the key accepted, and returns the
   * underlying payload. If the signature or key is invalid, returns empty.
   */
  Optional<String> verifyAndUnwrap(String signedPayload);
}
```

and offer a simple (single server), in-memory only (won't survive restarts) implementation to start from, and configuration to provide a custom implementation.

[...]

PR #10354, opened 2026-06-26
Restrict enhanced-class field deserialization to declared types

Samin061:
The server-only fields of an enhanced JPA/JDO class are read back through a plain ObjectInputStream in ServerSerializationStreamReader.deserializeClass, and that blob round-trips through the client, so a tampered RPC request can make readObject instantiate an arbitrary class and run a gadget chain once gwt.enhancedClasses.enabled is set. This constrains resolveClass to the field types the enhanced class actually declares (plus strings and boxed primitives), so the real server-only values still load while unrelated classes are rejected before instantiation.

Related to #9881 / #9880. This is not a comprehensive fix and does not by itself make the feature safe to re-enable; see the discussion thread for the limitations.

### DIFF
--- a/user/src/com/google/gwt/user/server/rpc/impl/ServerSerializationStreamReader.java
+++ b/user/src/com/google/gwt/user/server/rpc/impl/ServerSerializationStreamReader.java
@@ -27,7 +27,10 @@ import com.google.gwt.user.server.rpc.ServerCustomFieldSerializer;
 
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.InvalidClassException;
 import java.io.ObjectInputStream;
+import java.io.ObjectStreamClass;
 import java.lang.reflect.Array;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
@@ -39,6 +42,7 @@ import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.IdentityHashMap;
 import java.util.LinkedList;
 import java.util.Map;
@@ -799,7 +803,8 @@ public final class ServerSerializationStreamReader extends AbstractSerialization
         if (encodedData != null) {
           byte[] serializedData = Base64Utils.fromBase64(encodedData);
           ByteArrayInputStream baos = new ByteArrayInputStream(serializedData);
-          ObjectInputStream ois = new ObjectInputStream(baos);
+          ObjectInputStream ois =
+              new EnhancedFieldDeserializer(baos, instanceClass, clientFieldNames);
 
           int count = ois.readInt();
           for (int i = 0; i < count; i++) {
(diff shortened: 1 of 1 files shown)


niloc132:
I believe you may have linked the wrong issue - at present, 9709 is already resolved and enhanced classes are safe by default. Instead, maybe you would like to link to https://github.com/gwtproject/gwt/issues/9881, though you are sort of doing the opposite of what it describes, and should probably file a new issue to discuss. This isn't necessarily bad - you're making a "I know these types are definitely safe" fix, as opposed to the suggested "I know these types are annotated, but don't serialize their fields" fix. If this were a comprehensive way to resolve #9709, then the previous checks/warnings should be removed. I however suspect that #9880 is likely to be the only safe way to reenable this feature.

---

With that said, is this going to work to let a user re-enable the feature without negative impacts? For example, if there is an Object field, the security flaw will be wide open, since `allowed.isAssignableFrom(element)` will always be true, and then the flaw is wide open again. The EnhancedFieldDeserializer constructor loop also doesn't descend into field typers, so a field of type Person would itself be deserializable, but if Person in turn had an Address field, it would fail. Likewise generics - List/Maps/etc will only be deserializable if empty.

Can you confirm how you've tested this and your motivations for fixing this? If this is general security research and not rigorously tested in a production app, we're unlikely to be interested in this route to a fix.

Samin061:
You're right on the issue link, that was my mistake. I've dropped the 9709 reference from the description and I'll move the discussion to a separate issue rather than framing this as resolving it.

On the gaps, agreed on all of them. The Object case defeats the allowlist entirely since Object.isAssignableFrom is always true, and the same holds for any interface-typed or otherwise broad field. The constructor also only allows the directly-declared field types, so nested server-only fields (Person -> Address) and non-empty collections/maps fall outside the set and get rejected. So this narrows the surface for entities whose server-only fields are scalars/strings/concrete leaf types, but it doesn't make the feature generally safe to re-enable, and you're right that the existing checks/warnings shouldn't be removed on the back of it.

On testing and motivation: this came out of auditing the RPC deserialization path, not a production deployment. I validated the mechanism with a standalone reproducer (a tampered blob that runs readObject on an attacker class through the plain ObjectInputStream, and the same blob rejected with InvalidClassException before instantiation under the subclass), but I haven't exercised it against a real app with real enhanced entities. Given the holes above, I agree the don't-serialize-the-fields approach in #9880 is the sounder path. Happy to move this to a new issue to discuss.

niloc132:
What version were you testing against? Anything after 2.10.1 should be safe already (in that the feature is disabled outright, but can be selectively reenabled based on awareness of risk, serialization vulnerable dependencies, authentication, other controls), in that a serialization policy should never have client fields enabled, so this code path shouldn't be possible.

Samin061:
I wasn't testing against a released build. I reconstructed the enhanced-field blob format from deserializeClass (writeInt(count) followed by the name/value pairs) in a standalone harness and exercised the resolveClass logic there, rather than driving it through an actual serialization policy. So you're right that I never confirmed the path is reachable on 2.10.1+, where the policy shouldn't carry client fields to begin with. That guard sits in front of this code and I didn't validate against it.

Given that plus the holes from earlier (Object/interface-typed fields, no descent into nested field types, empty-collections-only), I don't think this is the right route. Happy to close this out and leave the discussion on #9880's approach.

niloc132:
Yep, I think we'll close this in favor of either a fix for #9881 or #9880. I wouldn't say no to something that let the user explicitly build a list for ObjectInputStream to handle, but that seems like more trouble than its worth from here.

Samin061:
Sounds good, agreed. The explicit-list option would still put the burden on the user to get the full transitive closure right (nested field types, collection element types), so it has the same sharp edges as the annotation route without much upside. The don't-serialize-the-fields approach in #9880 avoids that class of mistake entirely, so I think that's the right place to land. Thanks for the back-and-forth.

PR #10356, opened 2026-07-01
validate permutation strong name in StackTraceDeobfuscator

Samin061:
StackTraceDeobfuscator builds the symbol and source map file names by concatenating the permutation strong name, and that name comes straight from the client X-GWT-Permutation header via the remote logging service, so a value such as ../../ reads files outside the configured symbolMaps directory. The RPC path already constrains the strong name to a bare compiler hash in ServerSerializationStreamReader; this applies the same check in the deobfuscator before the name is turned into a path, keeping valid hashes working while rejecting traversal.


zbynek:
Maybe you can compare this to #10328 and add your comments to that PR, it's probably better to have just one open PR for the same issue. Thanks.

Samin061:
Makes sense, one PR for this is cleaner. Left my comparison notes on #10328. Fine to keep that one.

PR #10357, opened 2026-07-02
normalize module base path before resolving rpc policy resource

Samin061:
ServerSerializationStreamReader validates the strong name before policy lookup, but the module base URL from the same request reaches loadSerializationPolicy untouched, where new URL(moduleBaseURL).getPath() is joined with the strong name and the .gwt.rpc suffix and opened through getResourceAsStream. A value such as http://host/ctx/../../WEB-INF/foo still satisfies the startsWith(contextPath) guard, so the ../ segments survive into the resource path and can point the policy load outside the module directory; getRequestModuleBasePath resolves the symbol-map path the same way. Normalize the path with URI.normalize() before the containment check at both call sites so traversal segments collapse and the existing check rejects anything that leaves the module directory.

### DIFF
--- a/user/src/com/google/gwt/user/server/rpc/RemoteServiceServlet.java
+++ b/user/src/com/google/gwt/user/server/rpc/RemoteServiceServlet.java
@@ -25,9 +25,8 @@ import com.google.gwt.user.client.rpc.SerializationException;
 
 import java.io.IOException;
 import java.io.InputStream;
-import java.net.MalformedURLException;
 import java.net.URI;
-import java.net.URL;
+import java.net.URISyntaxException;
 import java.text.ParseException;
 import java.util.HashMap;
 import java.util.Map;
@@ -60,8 +59,8 @@ public class RemoteServiceServlet extends AbstractRemoteServiceServlet
     String modulePath = null;
     if (moduleBaseURL != null) {
       try {
-        modulePath = normalizeModulePath(new URL(moduleBaseURL).getPath());
-      } catch (MalformedURLException ex) {
+        modulePath = normalizeModulePath(moduleBaseURL);
+      } catch (URISyntaxException ex) {
         // log the information, we will default
         servlet.log("Malformed moduleBaseURL: " + moduleBaseURL, ex);
       }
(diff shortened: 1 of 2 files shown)


user/src/com/google/gwt/user/server/rpc/RemoteServiceServlet.java
```diff
@@ -137,6 +138,26 @@ static SerializationPolicy loadSerializationPolicy(HttpServlet servlet,
...
+   * cannot walk outside the module directory before the "same web application"
+   * containment check is applied. Returns {@code null} if the path cannot be
+   * parsed.
+   */
+  private static String normalizeModulePath(String path) {
+    if (path == null) {
+      return null;
+    }
+    try {
+      return URI.create(path).normalize().getRawPath();
```
zbynek: > This method is provided for use in situations where it is known that the given string is a legal URI [...]. The constructors, which throw `URISyntaxException` directly, should be used situations where a URI is being constructed from user input

per https://docs.oracle.com/en/java/javase/25/docs/api/java.base/java/net/URI.html#create(java.lang.String)

user/src/com/google/gwt/user/server/rpc/RemoteServiceServlet.java
```diff
@@ -59,7 +60,7 @@ static SerializationPolicy loadSerializationPolicy(HttpServlet servlet,
     String modulePath = null;
     if (moduleBaseURL != null) {
       try {
-        modulePath = new URL(moduleBaseURL).getPath();
+        modulePath = normalizeModulePath(new URL(moduleBaseURL).getPath());
```
zbynek: The argument passed to the normalization method is in both cases obtained through `new URL(baseURL).getPath()`, so now we're effectively doing
```java
new URI(new URL(baseURL).getPath()).normalize().getRawPath()
```
Couldn't we simplify that to
```java
new URI(baseURL).normalize().getRawPath()
```
and replace the checks for `MalformedURLException` with `URISyntaxException`?

Samin061: Good call, that's cleaner. Pushed: normalizeModulePath now takes the base URL and does new URI(baseURL).normalize().getRawPath() directly, so the intermediate new URL().getPath() is gone and both call sites catch URISyntaxException instead of MalformedURLException. Normalization output is unchanged for the traversal cases (/MyModule/../../../secret still collapses to /../../secret and gets rejected).

user/src/com/google/gwt/user/server/rpc/RemoteServiceServlet.java
Samin061: Right, since this is user input the constructor is the correct choice. That folded into the same change: it now uses new URI(moduleBaseURL) and lets URISyntaxException propagate to the existing catch blocks, so URI.create is gone.

Samin061:
gentle ping

zbynek (approved):
My previous concerns are addressed and the code makes sense, but I don't know much about RPC so I can't review this fully.

niloc132 (approved):
I think we could do better, handling the exception and null param inside the new helper method, but I also don't think its worth doing it now - this code doesnt get a lot of changes, and isn't needing more churn.

PR #10358, opened 2026-07-03
Fix compile errors for records with custom constructors, update JDT

zbynek:
Updating JDT by itself is enough to fix the issue described in #10045 .

There is also a similar issue with

```java
record RecordWithDefaultConstructor(String foo) {
  public RecordWithDefaultConstructor {
    Objects.requireNonNull(foo);
  }
}
```

where JDT provides two canonical constructors:

```java
// this one has no binding
RecordWithDefaultConstructor(String foo) {
  super();
  .foo = foo;
}

// this one has a binding
public RecordWithDefaultConstructor(String foo) {
  super();
  Objects.requireNonNull(foo);
  this.foo = foo;
}
```
Skipping the one without a binding prevents a NPE.

Unit tests added for both, minor tweaks needed to make tests pass with new JDT.

Fixes #10045

Relies on https://github.com/gwtproject/tools/pull/41

PR #10359, opened 2026-07-04
Avoid warning when compiling records

zbynek:
Fixes #10252 by rearranging the compilation steps -- record components have to be implemented earlier.

### DIFF
--- a/dev/core/test/com/google/gwt/dev/jjs/impl/JsInteropRestrictionCheckerTest.java
+++ b/dev/core/test/com/google/gwt/dev/jjs/impl/JsInteropRestrictionCheckerTest.java
@@ -2522,7 +2522,7 @@ public class JsInteropRestrictionCheckerTest extends OptimizerTestBase {
     assertBuggySucceeds();
   }
 
-  public void testUnusuableByJsWarns() throws Exception {
+  public void testUnusableByJsWarns() throws Exception {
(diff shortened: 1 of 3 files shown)


dev/core/test/com/google/gwt/dev/jjs/impl/JsInteropRestrictionCheckerTest.java
```diff
@@ -2625,6 +2625,18 @@ public void testUnusableByJsSyntheticMembersSucceeds() throws Exception {
     assertBuggySucceeds();
   }
 
+  public void testRecordSucceeds() throws Exception {
+    // override equals and hashCode so that we don't need the source for java.lang.Objects
+    addSnippetClassDecl("""
+        public class Hidden { }
+        public record Buggy(Hidden val) {
+          public boolean equals(Object o) {return o == this;}
```
niloc132: If instead you add java.util.Objects with just hash and equals in JavaResourceBase (like the changes made in https://github.com/gwtproject/gwt/pull/10358/changes#diff-5ec48ee761efac77bce719e54d010bd9cf322235a29da742abdce91e61bfdf87R251), you won't need this workaround. Like the other classes/methods, they can be no-op impls, and not all methods are required.

niloc132 (dismissed):

niloc132 (approved):

PR #10361, opened 2026-07-06
Make source map cache in StackTraceDeobfuscator thread-safe

Samin061:
ConcurrentHashMap's computeIfAbsent is used now to populate the cache safely when entries are missing.

### DIFF
--- a/user/src/com/google/gwt/core/server/StackTraceDeobfuscator.java
+++ b/user/src/com/google/gwt/core/server/StackTraceDeobfuscator.java
@@ -154,8 +154,7 @@ public abstract class StackTraceDeobfuscator {
   private static final int LINE_NUMBER_UNKNOWN = -1;
   private static final String SYMBOL_DATA_UNKNOWN = "";
 
-  private final Map<String, SourceMapping> sourceMaps =
-      new ConcurrentHashMap<String, SourceMapping>();
+  private final Map<String, SourceMapping> sourceMaps = new ConcurrentHashMap<>();
   private final SymbolCache symbolCache = new SymbolCache();
   private boolean lazyLoad = false;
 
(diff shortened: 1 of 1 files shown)


user/src/com/google/gwt/core/server/StackTraceDeobfuscator.java
```diff
@@ -154,7 +154,8 @@ Map<String, String> getAll(String strongName, Set<String> symbols) {
   private static final int LINE_NUMBER_UNKNOWN = -1;
   private static final String SYMBOL_DATA_UNKNOWN = "";
 
-  private final Map<String, SourceMapping> sourceMaps = new HashMap<String, SourceMapping>();
+  private final Map<String, SourceMapping> sourceMaps =
+      new ConcurrentHashMap<String, SourceMapping>();
```
niloc132: Generic args need not be explicitly specified

```suggestion
  private final Map<String, SourceMapping> sourceMaps = new ConcurrentHashMap<>();
```

niloc132 (commented):
Unless I'm mistaken, this is still slightly racy when it is accessed - the get/put calls should be replaced with computeIfAbsent so that we only write once? Probably not a dangerous race, but in the same cases where the old code could lead to a race/deadlock, we could end up doing the code twice. If you're cleaning this up, let's clean it all?

Related, but not necessarily blocking: I think SymbolCache could be more efficient too, at least not requiring exclusive readers, only writing should block reading (and writes are only possible for the first read, likewise a computeIfAbsent).

Samin061:
Good point. Switched loadSourceMap to computeIfAbsent so the parse only runs once per key, and dropped the explicit generic args per your inline note.

I left SymbolCache alone to keep this focused on the sourceMaps race. The read/write split there is a bit more involved since the inner maps currently rely on the synchronized block, so I'd rather do it as a separate change if that's alright.

user/src/com/google/gwt/core/server/StackTraceDeobfuscator.java
Samin061: Done, applied.

Samin061:
gentle ping

Samin061:
any update?

niloc132 (approved):

PR #10365, opened 2026-07-09
Make some of the c.g.g.view.client.*SelectionModelTest tests async.

### LINKED ISSUE #9622: Various broken SelectionModel tests in com.google.gwt.view.client
**GWT version:** HEAD
**Browser (with version):** all
**Operating System:** all

---

##### Description

All these selection models

- `DefaultSelectionModel`
- `MultiSelectionModel`
- `NoSelectionModel`
- `SingleSelectionModel`

fire `SelectionChangeEvents` using a finally command in their `setSelected()` method.

However their corresponding test classes contain tests that 

1. are not async using `delayTestFinish` and a `Timer` to make sure `SelectionChangeHandler`s are executed within the context of the test.
2. have broken implementation because they assume the code under test executes synchronously. These tests currently rely on ordering of `setSelected()` and `addSelectionChangeHandler` but that ordering is useless because the code under test works asynchronous.

Because these tests are not asynchronous they currently succeed, however in reality they fail asynchronously when finally commands are executed. This somehow cause other, unrelated, asynchronously tests to fail in Gradle. For some reason these unrelated tests succeed when using ANT, as on build.gwtproject.org, but you can still see HTMLUnit throwing these unexpected exceptions caused by the broken, failing tests listed below.

[Example: DefaultSelectionModelTest.testScheduleSelectionChangeEvent logs exception originating from failed assert in DefaultSelectionModelTest.testNoDuplicateChangeEvent](http://build.gwtproject.org/job/gwt/1332/testReport/com.google.gwt.view.client/DefaultSelectionModelTest/testScheduleSelectionChangeEvent/)

You can also see exceptions in the browser console using manual mode, however they are not reported back to the JUnit backend as it terminates beforehand.

The broken tests are:
- `DefaultSelectionModelTest`
   - `testNoDuplicateChangeEvent()`
- `SingleSelectionModelTest`
  - `testNoDuplicateChangeEvent()`
  - `testNoDuplicateChangeEvent2()`
- `MultiSelectionModelTest`
  - `testNoDuplicateChangeEvent()`
  - `testNoDuplicateChangeEvent2()`
  - `testNoDuplicateChangeEventWithKeyProvider()`

niloc132:
SelectionModel fires selection events using a finally command and some tests using SelectionChangeEvent.Handler were not async. In addition these tests had an implementation that let the test fail (but ANT never noticed because tests were sync).

This CL makes sure all tests using SelectionChangeEvent.Handler are async and updates implementation slightly to correclty assert event behavior.

Previously submitted as https://gwt-review.googlesource.com/c/gwt/+/21060.

Fixes #9622

### DIFF
--- a/user/test/com/google/gwt/view/client/DefaultSelectionModelTest.java
+++ b/user/test/com/google/gwt/view/client/DefaultSelectionModelTest.java
@@ -16,6 +16,9 @@
 package com.google.gwt.view.client;
 
 
+import com.google.gwt.core.client.Scheduler;
+import com.google.gwt.user.client.Timer;
+
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


user/test/com/google/gwt/view/client/DefaultSelectionModelTest.java
zbynek, line 85: Here the `AssertOneSelectionChangeEventOnlyHandler` is still listening, so it would fail in case another event is fired. We shouldn't need to ad a second handler for that.

user/test/com/google/gwt/view/client/MultiSelectionModelTest.java
zbynek, line 143: Again, this handler may not be needed.

user/test/com/google/gwt/view/client/SingleSelectionModelTest.java
zbynek, line 82: Again, this handler may not be needed.

zbynek (commented):
The code looks correct, but some of the handlers are not needed as far as I understand (see inline comments).
Alternatively, instead of two types of mock handlers it should be possible to have one type that counts events and let the test make assertions on those counts.

user/test/com/google/gwt/view/client/DefaultSelectionModelTest.java
jnehlmeier: `MockSelectionChangeHandler` requires you to call `assertEventFired` and it will clear its internal state after that call, so you always have to do event -> check -> event -> check. `MockSelectionChangeHandler` will fail if it sees a second event in a row without calling `assertEventFired` in between. 

If we would remove the failing handler then the first `model.setSelected()` following this line would only be recorded and only the second call would trigger the exception in `MockSelectionChangeHandler` because you haven't called `assertEventFired` in between two events. That makes the test a little imprecise. If you would want to call `assertEventFired` in between then it must happen async again which bloats the code.

The failing handler exists to simplify the logic and to make sure that the first additional event already fails the test and not the second.

zbynek (approved):

user/test/com/google/gwt/view/client/DefaultSelectionModelTest.java
zbynek: I missed that `assertEventFired` is also resetting the state 👍

PR #10366, opened 2026-07-10
use SecureRandom for external compile worker auth cookie

Samin061:
The one-time cookie that authenticates a CompilePermsServer worker connecting back to the compiler's ephemeral ServerSocket is generated with java.util.Random, whose 48-bit seed makes the value guessable, so someone able to reach the port during a compile could predict the cookie and feed crafted serialized data to the worker socket's ObjectInputStream. Generate it with SecureRandom instead.

Partial #10256


niloc132:
https://github.com/gwtproject/gwt/pull/10257 is a full patch for the linked issue - would you consider reviewing that instead?

Samin061:
Makes sense. #10257 covers the whole issue, including this SecureRandom swap plus authenticating the cookie before the ObjectInputStream is created and rejecting/logging bad cookies. Mine was only the entropy piece, so it's a strict subset. Closing this in favor of #10257, I'll review that one.

PR #10368, opened 2026-07-11
validate language subtag in GwtLocale fromString

Samin061:
On the server the active locale is read straight from the request by GwtServletBase.getGwtLocale (the locale query parameter or a cookie) and stored as the thread 'locale' property, and GwtLocaleFactoryImpl.fromString then parses that string into subtags. It checks the script, region and variant components against length and character rules but takes the primary language subtag as-is with only a lower-casing pass, and the code still carries a TODO to verify the language tag. Because the split only happens on '-' and '_', a value such as locale=java.lang.Runtime survives intact as the language, and LocalizableInstantiator concatenates that language into class names it passes to Class.forName().newInstance() when server code calls GWT.create on a Localizable, so '.', '$' and '/' can be smuggled into the reflective lookup. I came across it while tracing where the untrusted locale property finally lands. The fix validates the assembled language tag against the same BCP47 subtag shape the other components already require and throws IllegalArgumentException otherwise, kept inside fromString so every caller is covered without disturbing the parse flow. Existing locales, including extended-language and private-use forms like zh-cmn and x-foo123, are unaffected.

### DIFF
--- a/user/src/com/google/gwt/i18n/server/GwtLocaleFactoryImpl.java
+++ b/user/src/com/google/gwt/i18n/server/GwtLocaleFactoryImpl.java
@@ -36,6 +36,19 @@ public class GwtLocaleFactoryImpl implements GwtLocaleFactory {
     return matches(str, min, max, false);
   }
 
+  private static boolean isAlphaNumeric(String str, int min, int max) {
+    int len = str.length();
+    if (len < min || len > max) {
+      return false;
+    }
+    for (int i = 0; i < len; ++i) {
+      if (!Character.isLetterOrDigit(str.charAt(i))) {
+        return false;
+      }
+    }
+    return true;
+  }
+
   /**
    * Check if the supplied string matches length and composition requirements.
    * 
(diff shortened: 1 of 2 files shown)


user/src/com/google/gwt/i18n/server/GwtLocaleFactoryImpl.java
```diff
@@ -172,6 +204,10 @@ && isAlpha(part, 3, 3)) {
             part = localeParts.get(partIdx);
           }
         }
+        if (language.length() > 0 && !isValidLanguage(language)) {
```
zbynek: So in the middle of a method that splits the locale string into tags and makes assertions about each tag we now call another method that splits the string into tags and makes assertions about each tag? That makes the code much harder to follow. I didn't check how `fromString` behaves on your test strings before and after, but I would expect the `throw` on line 200 in the original code to take care of the invalid cases. If it doesn't, maybe some of the checks in the existing `fromString` could be stricter instead of introducing a validation method.

zbynek (changes requested):

user/src/com/google/gwt/i18n/server/GwtLocaleFactoryImpl.java
Samin061: Good point, and you're right that the throw on 200 doesn't catch these. The malicious inputs have no '-'/'_' so the split yields a single part and there's nothing left over to trip it. The reason it slips through is that the split lets any character that isn't '-'/'_' stay inside a part, and the primary language subtag was the one part never checked for composition.

I dropped the separate method and made the existing parse stricter instead: right after the split, each subtag now has to be 1-8 alphanumeric chars, so '.', '$' and '/' can't survive into a subtag. No re-splitting, and the later role checks are unchanged. english_USA still throws as before, and en-ext/zh-min-nan and the private-use forms still parse.

zbynek: Thanks, looks better! Some characters are still checked twice this way, but I think this is a reasonable trade-off between readable code and optimal performance.

zbynek (approved):

Samin061:
any update?

PR #10374, opened 2026-07-16
fix fragment id capture in StackTraceDeobfuscator file name regex

Samin061:
When a frame has no symbol map entry, which is the normal case for anonymous functions, the deobfuscator falls back to reading the code split fragment number out of the frame's file name. That name comes from the `//# sourceURL=<moduleName>-<fragmentId>.js` comment CrossSiteIframeLinker writes at the end of every fragment, so it is the only place the fragment number is still available on the server. The pattern used to read it back is `.*(\d+)\.js`, and because `.*` is greedy it backs off only far enough to leave a single digit for the capture group, so `app-12.js` reports fragment 2 and `app2-104.js` reports fragment 4. Under ten fragments that happens to be correct, which is presumably why it has gone unnoticed, but past that the wrong `_sourceMap<N>.json` is loaded and the frame gets resolved against unrelated source. Requiring a non-digit ahead of the captured run fixes the capture, and it also drops the quadratic backtracking the old pattern has on a file name that arrives from the client (a 32k name took ~700ms to reject, ~1ms after). The set of file names that match is unchanged, only the digits handed back differ, and I kept the change to the pattern so the surrounding fallback stays as it is.

### DIFF
--- a/user/src/com/google/gwt/core/server/StackTraceDeobfuscator.java
+++ b/user/src/com/google/gwt/core/server/StackTraceDeobfuscator.java
@@ -147,7 +147,7 @@ public abstract class StackTraceDeobfuscator {
   }
 
   private static final Pattern JsniRefPattern = Pattern.compile("@?([^:]+)::([^(]+)(\\((.*)\\))?");
-  private static final Pattern fragmentIdPattern = Pattern.compile("(?:.*[^\\d])?(\\d+)\\.js");
+  private static final Pattern fragmentIdPattern = Pattern.compile("(?:.*\\D)?(\\d+)\\.js");
   // Matches ServerSerializationStreamReader: the strong name reaches us straight from the
   // client (X-GWT-Permutation header) and is concatenated into symbol/source map file names.
   private static final Pattern strongNamePattern = Pattern.compile("[a-zA-Z0-9_]+");
--- a/user/src/com/google/gwt/core/server/StackTraceDeobfuscator.java
+++ b/user/src/com/google/gwt/core/server/StackTraceDeobfuscator.java
@@ -147,7 +147,7 @@ public abstract class StackTraceDeobfuscator {
   }
 
   private static final Pattern JsniRefPattern = Pattern.compile("@?([^:]+)::([^(]+)(\\((.*)\\))?");
-  private static final Pattern fragmentIdPattern = Pattern.compile(".*(\\d+)\\.js");
+  private static final Pattern fragmentIdPattern = Pattern.compile("(?:.*[^\\d])?(\\d+)\\.js");
   // Matches ServerSerializationStreamReader: the strong name reaches us straight from the
   // client (X-GWT-Permutation header) and is concatenated into symbol/source map file names.
   private static final Pattern strongNamePattern = Pattern.compile("[a-zA-Z0-9_]+");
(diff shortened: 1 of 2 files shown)


jnehlmeier:
Nice find! I am pretty sure that is the reason why a larger app I work on still produces wrong stack traces even after dodging https://github.com/gwtproject/gwt/issues/9931 . The app has more than 10 split points / fragments.

Samin061:
That lines up with what I'd expect. The fallback only runs when the symbol map has no entry for the method, so it's the anonymous-function frames that get a bad fragment id, and those are exactly the ones that then resolve against whichever source map the truncated digit happens to point at. With 10+ fragments `-12.js` becomes fragment 2, so the frame gets mapped through an unrelated fragment's source map and comes back with a plausible-looking but completely wrong file and line. Frames that do have a symbol map entry take the id from the symbol map and are unaffected, which fits the pattern in #9931 where only some frames in a trace are nonsense.

If you want to sanity check it against your app, the fragment ids that misresolve are the ones >= 10 whose last digit is a different valid fragment; a frame in fragment 12 should show up resolved against fragment 2's source.

user/src/com/google/gwt/core/server/StackTraceDeobfuscator.java
```diff
@@ -147,7 +147,7 @@ Map<String, String> getAll(String strongName, Set<String> symbols) {
   }
 
   private static final Pattern JsniRefPattern = Pattern.compile("@?([^:]+)::([^(]+)(\\((.*)\\))?");
-  private static final Pattern fragmentIdPattern = Pattern.compile(".*(\\d+)\\.js");
+  private static final Pattern fragmentIdPattern = Pattern.compile("(?:.*[^\\d])?(\\d+)\\.js");
```
zbynek:
```suggestion
  private static final Pattern fragmentIdPattern = Pattern.compile("(?:.*\\D)?(\\d+)\\.js");
```

Samin061: Done, pushed. `\D` is the same character class so the captures come out identical, just reads better.

zbynek (dismissed):

user/test/com/google/gwt/core/server/StackTraceDeobfuscatorTest.java
```diff
@@ -53,4 +65,25 @@ public void testValidStrongNameStillLoadsSymbolMap() {
...
+    RecordingDeobfuscator d = new RecordingDeobfuscator();
+    d.resymbolize(traceInFragmentFile("app-5.js"), STRONG_NAME);
+    assertTrue("expected fragment 5 to be requested: " + d.opened,
+        d.opened.contains(STRONG_NAME + "_sourceMap5.json"));
+  }
+
+  public void testMultiDigitFragmentIdIsReadFromFileName() {
+    RecordingDeobfuscator d = new RecordingDeobfuscator();
+    d.resymbolize(traceInFragmentFile("app-12.js"), STRONG_NAME);
+    assertTrue("expected fragment 12, not its last digit: " + d.opened,
```
jnehlmeier: Use the same message style as above. The test is universal but `, not its last digit` references this concrete bug here. If this test fails in the future because of another bug the message might be confusing/wrong if you dont see the last digit in the failing test output. Same with the test below.

jnehlmeier (changes requested):
Just a small wording change. Otherwise its fine.

user/test/com/google/gwt/core/server/StackTraceDeobfuscatorTest.java
Samin061: Good point, done for both. Dropped the "not its last digit" wording so the message doesn't assume this particular cause - the recorded `opened` list is already in the message, so a future failure still shows which fragment actually got requested.

jnehlmeier (approved):

zbynek (approved):

Samin061:
any update?

PR #10377, opened 2026-07-18
reject unquoted angle brackets in SafeStyles style name and value checks

Samin061:
SafeStylesHostedModeUtils.isValidStyleName and isValidStyleValue reject a misplaced `;`, `:`, `-` and unpaired quotes, but not `<` or `>`, even though SafeStylesUtils.verifySafeStylesConstraints lists them alongside the same characters. This makes the two consistent.

Names reject `<` and `>` outright. Values reject them only outside quotes and url(), matching the existing `;` handling, so `list-style-type: ">"` and similar stay valid.

Not a security fix: the `*Trusted*` methods put the burden on the caller, and these checks are dev-mode assertions. Purely a consistency tweak to the existing validation.

Test cases added to the INVALID_STYLE_NAMES / INVALID_STYLE_VALUES and VALID_STYLE_VALUES arrays in GwtSafeStylesUtilsTest, which both GwtSafeStylesUtilsTest and GwtSafeStylesHostedModeUtilsTest iterate.

### DIFF
--- a/user/src/com/google/gwt/safecss/shared/SafeStylesHostedModeUtils.java
+++ b/user/src/com/google/gwt/safecss/shared/SafeStylesHostedModeUtils.java
@@ -95,6 +95,8 @@ public class SafeStylesHostedModeUtils {
       return "Style property names cannot contain a semi-colon: " + name;
     } else if (name.indexOf(':') >= 0) {
       return "Style property names cannot contain a colon: " + name;
+    } else if (name.indexOf('<') >= 0 || name.indexOf('>') >= 0) {
+      return "Style property names cannot contain a bracket (< or >): " + name;
     }
 
     return null;
@@ -120,6 +122,14 @@ public class SafeStylesHostedModeUtils {
     // Empty value.
     if (value == null || value.length() == 0) {
       return "Style property values cannot be null or empty";
+    }
+
+    /*
+     * The HTML tokenizer runs before the CSS parser, so a bracket ends the
+     * enclosing style element even inside a quoted string or a url().
+     */
+    if (value.indexOf('<') >= 0 || value.indexOf('>') >= 0) {
+      return "Style property values cannot contain a bracket (< or >): " + value;
     }
 
     /*
--- a/user/src/com/google/gwt/safecss/shared/SafeStylesHostedModeUtils.java
+++ b/user/src/com/google/gwt/safecss/shared/SafeStylesHostedModeUtils.java
@@ -96,7 +96,7 @@ public class SafeStylesHostedModeUtils {
     } else if (name.indexOf(':') >= 0) {
       return "Style property names cannot contain a colon: " + name;
     } else if (name.indexOf('<') >= 0 || name.indexOf('>') >= 0) {
-      return "Style property names cannot contain a bracket (< or >): " + name;
+      return "Style property names cannot contain an angle bracket: " + name;
     }
 
     return null;
@@ -125,17 +125,9 @@ public class SafeStylesHostedModeUtils {
     }
 
     /*
-     * The HTML tokenizer runs before the CSS parser, so a bracket ends the
-     * enclosing style element even inside a quoted string or a url().
-     */
-    if (value.indexOf('<') >= 0 || value.indexOf('>') >= 0) {
-      return "Style property values cannot contain a bracket (< or >): " + value;
-    }
-
-    /*
      * A value can contain any token, but parenthesis, brackets, braces, and
-     * single/double quotes must be paired. Semi-colons are only allowed in
-     * strings, such as in a url.
+     * single/double quotes must be paired. Semi-colons and angle brackets are
+     * only allowed in strings, such as in a url.
      */
 
     // Create a map of pairable 'open' characters to 'close' characters.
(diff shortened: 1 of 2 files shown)


zbynek:
Angle brackets can be safely banned in property names (AFAIK), but not in values, the simplest example being 

```html
<ul style='list-style-type: ">"'><li>x</li></ul>
```

There is an existing restriction on `;` only being allowed inside quotes, I think `<` and `>` are similar cases.

Samin061:
Fair point, that example is clearly legitimate. Moved the check into the character loop next to the `;` and `:` branches, so `<` and `>` are now only rejected outside quotes and url(). Names still reject them outright.

Dropped the two test cases that relied on the old blanket check (the quoted and url() ones) and added `">"` plus a couple of similar values to VALID_STYLE_VALUES so the quoted case stays covered.

jnehlmeier:
I don't think this change is required and it looks like an AI discovery which did not know about the convention that calling any `*Trusted*` method means that these methods expect the developer to assert that the passed in arguments are safe. If the developer decides that the parameters are not safe then it should not call any `*Trusted*` method. 

Thus these `*Trusted*` code paths only do some minimal checks for common errors and do not verify that all `SafeStyles` requirements are met. It is also clearly documented that `SafeStylesHostedModeUtils` should not be used directly and `isValidStyleName`/`isValidStyleValue` also clearly document that not everything is checked.

So calling `forTrustedColor("red</style><script>alert(1)</script>")` is totally fine to succeed as GWT expects the developer to verify it.

niloc132:
I think Jens is right - these are assertions only, and by design will not run in an application in production. If there is a correctness issue that prevents valid developer-trusted strings from working in these methods, then that should be addressed, but if we're worried about a developer seeing an error at dev time about having attacked themselves... I'm not sure that makes any sense. The developer-attacker should have just written the attack in plain Java rather than obfuscate it against themselves.

@Samin061 can you confirm what the specific issue is we're defending against here?

zbynek:
Framing this as a security improvement is wrong. OTOH if `SafeStylesHostedModeUtils` is spending ~150 lines of code on checking that there is no misplaced `;`, `:`, `-` or `"`, spending extra 6 lines on `<` and `>` sounds like a reasonable consistency improvement. In the long term, native `CSS.supports` could replace those 150 lines, making tweaks like this unnecessary.

Samin061:
You're both right, and to answer @niloc132 directly: there isn't a concrete attack here. I read the `<`/`>` line in verifySafeStylesConstraints as a constraint the validators were meant to enforce and missed that `*Trusted*` means the caller has already vouched for the string. A developer passing an attack into forTrustedColor has already lost, and these checks don't run in production anyway. My security framing was wrong.

What's left is what @zbynek described: the validators already spend a lot of code on a misplaced `;`, `:`, `-` and unpaired quotes, and `<`/`>` are in the same documented list but unchecked. Six lines to make that consistent, catching a typo at dev time rather than defending anything.

I've retitled the PR and rewritten the description to drop the security angle, and removed the `</style><script>` test case and the comment that went with it, since the plain `contains<open-bracket` cases cover the same path without implying a threat model. Happy to close it if the consistency alone isn't worth the lines.

On the tooling question, yes, I use an assistant a bit for research and sanity-checking. The reading of the trusted-input contract was mine and it was wrong.

jnehlmeier:
> Framing this as a security improvement is wrong. OTOH if `SafeStylesHostedModeUtils` is spending ~150 lines of code on checking that there is no misplaced `;`, `:`, `-` or `"`, spending extra 6 lines on `<` and `>` sounds like a reasonable consistency improvement. In the long term, native `CSS.supports` could replace those 150 lines, making tweaks like this unnecessary.

Well these methods are a lite syntax check and thus have a few issues. 

The method about style names does not have any escape logic, thus adding additional disallowed characters will make the check worse. Currently it only produces a false negative with `style\;name` / `style\:name`. The method also currently disallows CSS variables because it was based on CSS 2.1. On the other hand because of the missing escape logic it has no issues with variable length unicode escapes.

In contrast the method which checks style values has a simple escape logic which ignores the next character without checking if it is an unicode escape. It doesn't do any harm but the code should skip more than one character if the escape is an unicode escape. Of course this method also has issues with modern CSS which uses `:` and `;`, e.g.
```css
background-image: if(
    style(--scheme: ice): linear-gradient(#caf0f8, white, #caf0f8);
    style(--scheme: fire): linear-gradient(#ffc971, white, #ffc971);
    else: none;
  );
```

IMHO it is difficult to draw a clear line how far syntax check should go for an input string that a developer is expected to check anyways. Given that CSS is becoming more complex these simple kind of checks can also become too simple and then fail, especially with the value-checking method.

Samin061:
That settles it for me. The name check has no escape handling at all, so every character added to that list is another escaped-but-rejected false negative, and the existing comment already treats the `\;` / `\:` case as a tolerated cost rather than a precedent to extend.

The value side has the same problem from the other direction. Your own `if()` example is the counterexample: `background-image: if(media(width > 400px): red; else: blue)` puts an unquoted `>` in a perfectly valid value, so the check I added would reject it. Range syntax is only going to show up more.

So this is a net negative on both methods and I was patching a doc comment rather than a real defect. I will leave the PR open for a moment in case you want to close it with your own note, otherwise say the word and I will close it.

PR #10383, opened 2026-07-22
Improve duplicate clinit removal

### LINKED ISSUE #9731: dead code not eliminated when two Boolean.parseBoolean in same function

niloc132:
Explore comma expression contents when finding class initializers, and mark them as seen to avoid their inclusion later in method bodies. Also improved try block handling when there are no catches or finallys, and left notes for how this can be later improved when control flow is more structured (#10248, #10381)

Moved this pass into the main JS optimization loop - inlining methods tends to produce grouped clinits, and clinit removal can leave behind dead code that needs to be simplified away. In some cases, clinit removal make also make a method eligible for inlining.

This code previously had no tests, so added a test class, and improved the JS OptimizerTestBase to allow for minor changes (newlines, etc) between input and expected contents, as well as asserting for "no change" when appropriate. This more closely mirrors the Java version of this test base.

Also cleans up some "XO" references ("eXactly Once") in favor of "clinit" as referenced elsewhere.

Fixes #9731

### DIFF
--- a/dev/core/src/com/google/gwt/dev/js/DuplicateClinitRemover.java
+++ b/dev/core/src/com/google/gwt/dev/js/DuplicateClinitRemover.java
@@ -102,7 +102,9 @@ public class DuplicateClinitRemover extends JsModVisitor {
     if (x.getOperator() == JsBinaryOperator.COMMA) {
 
       // This effectively visits any JsInvocation direct child on both sides, so take care to not
-      // encounter any clinit twice when descending further.
+      // encounter any clinit twice when descending further. Important: if the right was a new
+      // clinit, we must not go back and re-check the left or we'll have changed the order of
+      // execution.
       ClinitStatus left = isDuplicateCall(x.getArg1());
       ClinitStatus right = isDuplicateCall(x.getArg2());
 
@@ -147,8 +149,17 @@ public class DuplicateClinitRemover extends JsModVisitor {
         }
         return false;
       }
-      // Descend to both sides only if neither is a clinit at all
-      return right == ClinitStatus.NOT_A_CLINIT && left == ClinitStatus.NOT_A_CLINIT;
+      // Descend to both sides only if neither is a clinit at all - if just the left is a clinit,
+      // we must still descend to the right manually
+      if (right == ClinitStatus.NOT_A_CLINIT) {
+        if (left == ClinitStatus.NOT_A_CLINIT) {
+          // descend into both, neither is a clinit
+          return true;
+        }
+        x.setArg2(accept(x.getArg2()));
+        return false;
+      }
+      return false;
     } else if (x.getOperator().equals(JsBinaryOperator.AND)
         || x.getOperator().equals(JsBinaryOperator.OR)) {
       x.setArg1(accept(x.getArg1()));
--- a/dev/core/src/com/google/gwt/dev/js/DuplicateClinitRemover.java
+++ b/dev/core/src/com/google/gwt/dev/js/DuplicateClinitRemover.java
@@ -24,8 +24,6 @@ import com.google.gwt.dev.js.ast.JsCase;
 import com.google.gwt.dev.js.ast.JsConditional;
 import com.google.gwt.dev.js.ast.JsContext;
 import com.google.gwt.dev.js.ast.JsDefault;
-import com.google.gwt.dev.js.ast.JsEmpty;
-import com.google.gwt.dev.js.ast.JsExprStmt;
 import com.google.gwt.dev.js.ast.JsExpression;
 import com.google.gwt.dev.js.ast.JsFor;
 import com.google.gwt.dev.js.ast.JsForIn;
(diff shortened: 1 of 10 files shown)


niloc132:
Marked as draft for now, some additional tests are required, and some text blocks can be used in place of separate string vararg params. Showcase improves by about 0.1%, or about 300 bytes over 300-400kb permutations.

dev/core/src/com/google/gwt/dev/js/DuplicateClinitRemover.java
```diff
@@ -227,19 +258,46 @@ public boolean visit(JsIf x, JsContext ctx) {
...
+  @Override
+  public boolean visit(JsTry x, JsContext ctx) {
+    if (!x.getCatches().isEmpty()) {
+      // Catch could return control to parent block without completing the try block, so branch
+      // the try block if there is any catch.
+      x.setTryBlock(branch(x.getTryBlock()));
+      List<JsCatch> catches = x.getCatches();
+      for (int i = 0; i < catches.size(); i++) {
+        JsCatch aCatch = catches.get(i);
+        JsCatch c = accept(aCatch);
```
niloc132, line 281: this should almost certainly be a branch(), we can't guarantee a catch runs.

dev/core/test/com/google/gwt/dev/js/JsDuplicateCaseFolderTest.java
```diff
@@ -23,73 +25,68 @@ public class JsDuplicateCaseFolderTest extends OptimizerTestBase {
...
     String input = "function a(x){switch(x){case 0:return 17;case 12:case 1:return 18;default:return 17;case 3:return 18;}}";
     String expected = "function a(x){switch(x){default:case 0:return 17;case 12:case 3:case 1:return 18;}}\n";
-    check(expected, input);
+    optimize(input).into(expected);
   }
 
   public void test2() throws Exception {
     // Don't coalesces cases 0 and 2 since 2 can be reached via fallthrough from 1
     String input = "function a(x){var y;switch(x){case 0:return 17;case 1:y=18;case 2:return 17;case 3:y=18;}return y}";
     String expected = "function a(x){var y;switch(x){case 0:return 17;case 1:y=18;case 2:return 17;case 3:y=18;}return y}\n";
```
niloc132, line 40: In this PR or another, these would all make good text blocks.

dev/core/test/com/google/gwt/dev/js/JsDuplicateClinitRemoverTest.java
```diff
@@ -0,0 +1,402 @@
...
+/**
+ * Need tests for
+ * switch/forin
+ * super clinits:
+ * if/for/while/do/conditional/short-circuit/switch/forin
+ *
+ * Several tests use an expression like "x + (clinit_A(), y) > 0" to avoid short-circuiting but
+ * still allow for side effects so the clinit won't be moved out before the for condition.
+ */
+public class JsDuplicateClinitRemoverTest extends OptimizerTestBase {
```
niloc132, line 33: Possibly rename the DuplicateClinitRemover to have a Js prefix like this does? that seems to be the convention for other passes.

dev/core/test/com/google/gwt/dev/js/JsDuplicateClinitRemoverTest.java
```diff
@@ -0,0 +1,402 @@
...
+    )
+        .into(CLINIT_DECL,
+            "clinit_A();",
+            "alert(cond1 && cond2);");
+    optimize(CLINIT_DECL,
+        "clinit_A();",
+        "alert(cond1 && (clinit_A(), cond2));"
+    )
+        .into(CLINIT_DECL,
+            "clinit_A();",
```
niloc132, line 315: Text blocks all through here for sure.

dev/core/src/com/google/gwt/dev/js/DuplicateClinitRemover.java
niloc132: Fixed, added a test

PR #10386, opened 2026-08-01
Add a compile-time switch for RPC enhanced classes

### LINKED ISSUE #9881: Compile-time configuration to disable rpc.enhancedClasses feature

kamexETH:
Add the single-valued `rpc.enhancedClasses.enabled` configuration
property, defaulting to `true` for compatibility.

When set to `false`, RPC ignores both JPA/JDO enhancement annotations and
the `rpc.enhancedClasses` list. Generated field serializers omit the opaque
server-enhanced payload, serialization policies omit `@ClientFields`, and the
server no longer sees those JPA/JDO fields when clients return them.

Missing, malformed, or multi-valued settings log a warning and retain the
backwards-compatible enabled behavior. The enhanced-class settings are also
included in RPC generator cache validation, preventing field serializers
generated under different settings from being reused.

Tests:

- `ant -f user/build.xml -Dgwt.nongwt.testcase.includes="**/RpcJreSuite.class" test.nongwt`
- `ant -f user/build.xml checkstyle`

Fixes #9881


user/src/com/google/gwt/user/rebind/rpc/Shared.java
```diff
@@ -80,6 +86,32 @@ static Set<String> getEnhancedTypes(PropertyOracle propertyOracle) {
...
+   */
+  static boolean shouldEnableEnhancedClasses(PropertyOracle propertyOracle) {
+    try {
+      ConfigurationProperty prop =
+          propertyOracle.getConfigurationProperty(RPC_ENHANCED_CLASSES_ENABLED);
+      if (prop.getValues().size() == 1) {
+        return Boolean.parseBoolean(prop.getValues().get(0));
+      }
+    } catch (BadPropertyValueException e) {
+      // Preserve the historical behavior when compiling without the new property.
```
niloc132: Why do we need this, given that you also changed RemoteService.gwt.xml?

user/src/com/google/gwt/user/rebind/rpc/Shared.java
```diff
@@ -80,6 +86,32 @@ static Set<String> getEnhancedTypes(PropertyOracle propertyOracle) {
...
+   * @param propertyOracle the property oracle used to access the relevant configuration property
+   * @return whether enhanced class handling is enabled
+   */
+  static boolean shouldEnableEnhancedClasses(PropertyOracle propertyOracle) {
+    try {
+      ConfigurationProperty prop =
+          propertyOracle.getConfigurationProperty(RPC_ENHANCED_CLASSES_ENABLED);
+      if (prop.getValues().size() == 1) {
+        return Boolean.parseBoolean(prop.getValues().get(0));
+      }
```
niloc132, line 107: We _shouldn't_ need this check, as the new property is configured to be single valued. If we're somehow supporting cases where the RemoteService.gwt.xml was replaced, we should warn - the running code doesn't expect this.

So there should be roughly three outcomes
 - single value, parses to true
 - single value, parses to false
 - size != 1 or exception or fail to parse, warn and return true

What do you think?

user/src/com/google/gwt/user/RemoteService.gwt.xml
```diff
@@ -90,6 +90,20 @@
   -->
   <define-configuration-property name="rpc.enhancedClasses" is-multi-valued="true"/>
 
+  <!--
+      Controls whether RPC generates support for server-enhanced classes. Set
+      this to false to ignore both JPA/JDO annotations and the
+      rpc.enhancedClasses list, preventing @ClientFields entries and the
+      corresponding client-side payload handling from being generated.
```
niloc132: Maybe tie this back to the actual server behavior being lost through this?

Something like
```suggestion
      corresponding client-side payload handling from being generated
      and so the server from seeing JPA/JDO fields returned to it from
      client calls
```

user/src/com/google/gwt/user/rebind/rpc/FieldSerializerCreator.java
```diff
@@ -580,7 +580,7 @@ private void writeClassSerializationStatements() {
      * {@link WeakMapping}.
      */
 
-    if (serializableClass.isEnhanced()) {
+    if (Shared.isEnhancedClass(context.getPropertyOracle(), serializableClass)) {
```
niloc132: This seems very late to check this - could we instead just not set isEnhanced as you've already done so that this check fails?

user/src/com/google/gwt/user/rebind/rpc/Shared.java
```diff
@@ -80,6 +86,32 @@ static Set<String> getEnhancedTypes(PropertyOracle propertyOracle) {
...
+      // Preserve the historical behavior when compiling without the new property.
+    }
+    return true;
+  }
+
+  /**
+   * Returns whether the type should be treated as enhanced for the current compilation.
+   */
+  static boolean isEnhancedClass(PropertyOracle propertyOracle, JClassType type) {
+    return shouldEnableEnhancedClasses(propertyOracle) && type.isEnhanced();
```
niloc132: If we do need this, contrary to my comment in FieldSerializerCreator, let's swap the && and do the cheap field check before the map lookups

niloc132 (commented):
Thanks for working on this! A few comments:

user/src/com/google/gwt/user/rebind/rpc/Shared.java
kamexETH: Agreed—the inherited module normally guarantees the property. I kept only a warned, backwards-compatible fallback for custom modules that replace or do not inherit `RemoteService.gwt.xml`: a missing or malformed value now logs a warning and returns `true`.

user/src/com/google/gwt/user/rebind/rpc/Shared.java
kamexETH: Implemented those three outcomes. A single case-insensitive `true` or `false` is parsed directly; a missing, invalid, or multi-valued property logs a warning and returns `true`. The parsing test now covers all five cases.

user/src/com/google/gwt/user/RemoteService.gwt.xml
kamexETH: Updated the module documentation to make the server-side consequence explicit: when disabled, the server cannot see the JPA/JDO fields returned to it from client calls.

user/src/com/google/gwt/user/rebind/rpc/FieldSerializerCreator.java
kamexETH: Done. I removed the late property checks and restored the direct `serializableClass.isEnhanced()` checks. The builder now remains the single place that decides whether to mark a type as enhanced.

user/src/com/google/gwt/user/rebind/rpc/Shared.java
kamexETH: Removed `Shared.isEnhancedClass` entirely, so there is no property lookup at this late stage. `FieldSerializerCreator` and `ProxyCreator` now use the direct `isEnhanced()` checks again.

PR #10390, opened 2026-08-10
Clear cached JS for stale types even when currently unreachable

### LINKED ISSUE #9565: Uncaught ReferenceError: TEk_g$ is not defined SuperDevMode
**GWT version:** **2.8.2** 
**Browser (with version):** **Chrome 61**
**Operating System:**  **Windows 10**

---

##### Description

I am getting this Exception while running the SuperDevMode with Intellij **:**
`P2c_g$	z2c_g$	z1c_g$	initializeModules_0_g$	i1_g$	l1_g$	(anonymous function)	S0c_g$	(anonymous function)	Uncaught ReferenceError: TEk_g$ is not defined
`

in **com.google.gwt.logging.client.LogConfiguration.java** at line 

>  this.root = Logger.getLogger("");

While application is running fine when deployed on tomcat server (without SuperDevMode)

kevinushey:
## Problem

In incremental (Super Dev Mode) compiles, `MinimalRebuildCache.computeAndClearStaleTypesCache()` filters the stale type set down to currently-reachable types **before** clearing cached JS. The filter's comment assumes that an unreachable stale type has no cached JS ("since they're missing JS, they will already be fully traversed when seen in Unify") -- but that invariant does not hold: a type that merely *loses* reachability (without its compilation unit being modified) keeps its cached JS.

When such a type becomes reachable again in a later compile, `UnifyAst.needsNewJs()` sees JS present, marks the type reference-only, and splices the stale chunk into the output. If the type's shape changed in the interim -- for example synthetic lambda types being renumbered because a lambda was added earlier in the file -- freshly generated call sites reference a newly allocated global name while the stale chunk defines the old one:

```
ReferenceError: Widgets$lambda$2$Type_2_g$ is not defined
```

Even when the names happen to line up, the output silently runs outdated code. Because the code server persists the MinimalRebuildCache to disk (`$TMPDIR/gwt-cache-*`), the corruption survives code-server restarts and "clean" rebuilds of the application, which makes it look thoroughly mysterious to users. This is the likely root cause of the long-standing symptom reports in #9565 (see [analysis](https://github.com/gwtproject/gwt/issues/9565#issuecomment-5247014288)), where several commenters found that deleting the code server's tmp files resolved the error.

## Reproduction

Three-step incremental compile sequence (encoded in the included test):

1. Compile with `TestEntryPoint.onModuleLoad()` calling both `Widgets.ping()` and `Widgets.trigger()`; `trigger()` contains several lambdas. Their JS is cached.
2. Modify only the entry point to stop calling `trigger()`. `Widgets` is not modified, so its lambda types keep their cached JS -- but they are now unreachable.
3. Modify the entry point to call `trigger()` again, and modify `Widgets` so an extra lambda is inserted first (renumbering the synthetic lambda types). The lambda types are stale, but the reachability filter (based on the previous compile's reachability) skips clearing them; the output mixes stale definitions with fresh call sites and contains a `new Widgets$lambda$2$Type_2_g$(...)` whose constructor is never defined.

## Fix

Clear cached output for the full stale set before applying the reachability filter. The filter still limits which types are *artificially retraversed* (its stated purpose); clearing the JS of a currently-unreachable stale type costs nothing (it is not emitted) and restores the invariant that a cached JS entry is always current -- when the type becomes reachable again, `needsNewJs()` correctly triggers a fresh traversal.

## Testing

- New `CompilerTest#testIncrementalRecompile_unreachableStaleTypeRegainsReachability` fails on current `main` with `lambda type constructor Widgets$lambda$2$Type_2_g$ is referenced but never defined` and passes with the fix.
- Verified end-to-end against a live CodeServer (GWT 2.12.2): the same three-step sequence produces the dangling reference without the fix and consistent output with it.
- `MinimalRebuildCacheTest` passes unchanged.

Fixes #9565.

### DIFF
--- a/dev/core/src/com/google/gwt/dev/MinimalRebuildCache.java
+++ b/dev/core/src/com/google/gwt/dev/MinimalRebuildCache.java
@@ -324,6 +324,15 @@ public class MinimalRebuildCache implements Serializable {
       staleTypeNames.removeAll(JProgram.SYNTHETIC_TYPE_NAMES);
     }
 
+    // Clear the cached output of every stale type, including those that are not currently
+    // reachable. A stale type that is unreachable now may become reachable again in some later
+    // compile without being remarked stale, and it must not be reused from cache at that point:
+    // its cached JS was generated from an old version of its compilation unit and may reference
+    // (or collide with) global names differently than freshly generated code does.
+    for (String staleTypeName : staleTypeNames) {
+      clearCachedTypeOutput(staleTypeName);
+    }
+
     /*
      * Filter for just those stale types that are actually reachable. Since if they're not reachable
      * we don't want to artificially traverse them and unnecessarily reveal dependency problems. And
@@ -337,11 +346,7 @@ public class MinimalRebuildCache implements Serializable {
       logger.log(TreeLogger.DEBUG, "known modified types = " + modifiedTypeNames);
       logger.log(TreeLogger.DEBUG, "known modified resources = " + modifiedResourcePaths);
       logger.log(TreeLogger.DEBUG,
-          "clearing cached output for resulting stale types = " + staleTypeNames);
-    }
-
-    for (String staleTypeName : staleTypeNames) {
-      clearCachedTypeOutput(staleTypeName);
+          "stale types to retraverse = " + staleTypeNames);
     }
 
     return Sets.newHashSet(staleTypeNames);
--- a/dev/core/src/com/google/gwt/dev/MinimalRebuildCache.java
+++ b/dev/core/src/com/google/gwt/dev/MinimalRebuildCache.java
@@ -278,6 +278,10 @@ public class MinimalRebuildCache implements Serializable {
    * changes in the modified types. For example if the parent of class Foo was changed then the
    * castmaps in all children of Foo need to be recreated.
    * <p>
+   * Note that the returned set is not the same as the set of types whose cache was cleared. Cached
+   * output is cleared for every stale type, while the returned set is narrowed to the stale types
+   * that are currently reachable, since unreachable types should not be artificially retraversed.
+   * <p>
    * In some ways this process is similar to that performed by the CompilationUnitInvalidator but it
    * differs both in what type of cached objects are being cleared (JS versus CompilationUnits) and
    * in what invalidation rules must be applied. CompilationUnitInvalidator is concerned only with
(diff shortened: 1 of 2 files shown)


dev/core/src/com/google/gwt/dev/MinimalRebuildCache.java
```diff
@@ -324,6 +324,15 @@ public Set<String> computeAndClearStaleTypesCache(TreeLogger logger, JTypeOracle
       staleTypeNames.removeAll(JProgram.SYNTHETIC_TYPE_NAMES);
     }
 
+    // Clear the cached output of every stale type, including those that are not currently
```
zbynek, line 331: Maybe the Javadoc for the method should make it clear that the set of types that are removed from cache is not the same as the set returned from this method.

dev/core/test/com/google/gwt/dev/CompilerTest.java
```diff
@@ -1502,6 +1503,127 @@ public void testIncrementalRecompile_dateStampChange() throws UnableToCompleteEx
...
+  // outdated JS is reused and can reference (or collide with) global names inconsistently with
+  // freshly generated code.
+  public void testIncrementalRecompile_unreachableStaleTypeRegainsReachability()
+      throws UnableToCompleteException, IOException, InterruptedException {
+    // Uses PRETTY output so that synthetic lambda type names are recognizable in the JS.
+    JsOutputOption output = JsOutputOption.PRETTY;
+
+    MockJavaResource widgetsWithThreeLambdas =
+        JavaResourceBase.createMockJavaResource("com.foo.Widgets",
+            "package com.foo;",
```
zbynek: nit: Recent PRs are using text blocks, this is one of the places where it could increase readability.

dev/core/src/com/google/gwt/dev/MinimalRebuildCache.java
kevinushey: Good point -- the two sets diverged with this change. Added a paragraph in 756c8a6 noting that cached output is cleared for every stale type while the returned set is narrowed to the reachable ones.

dev/core/test/com/google/gwt/dev/CompilerTest.java
kevinushey: Done in aaf4998, matching the existing text block usage in `JavaResourceBase.OBJECTS`. Makes the two `Widgets` variants much easier to diff by eye.

PR #10392, opened 2026-08-11
Type-check the exemplar in the EnumMap server field serializer

### LINKED ISSUE #10391: EnumMap RPC deserialization throws an uncaught NullPointerException on crafted input
The server custom field serializer for `EnumMap` instantiates the map from an
**untyped** `readObject()`, unlike every other collection serializer in the RPC
core, which read what they instantiate from with a type check.

`user/src/com/google/gwt/user/server/rpc/core/java/util/EnumMap_ServerCustomFieldSerializer.java`
overrides the type-checking server instantiate but discards the type
information it is given and calls the client serializer:

```java
public EnumMap instantiateInstance(ServerSerializationStreamReader streamReader,
    Type[] expectedParameterTypes, DequeMap<TypeVariable<?>, Type> resolvedTypes)
    throws SerializationException {
  return EnumMap_CustomFieldSerializer.instantiate(streamReader); // untyped
}
```

`EnumMap_CustomFieldSerializer.instantiate` then does:

```java
Object exemplar = streamReader.readObject();   // no expected type
Class clazz = exemplar.getClass();
return new EnumMap(clazz);                      // expects an enum class
```

Compare `TreeMap_ServerCustomFieldSerializer` / `TreeSet_ServerCustomFieldSerializer`,
which read their comparator through the typed path
`streamReader.readObject(Comparator.class, resolvedTypes)`.

Because the exemplar is read untyped, the server takes the
`deserialize(signature, null, null)` path, where the
`SerializedTypeViolationException` check (guarded by `expectedType != null`) does
not run. An RPC message can therefore substitute an allow-listed **non-enum**
type for the exemplar, and `new EnumMap(nonEnumClass)` throws
`NullPointerException: Cannot read the array length because "this.keyUniverse"
is null`, which is not a `SerializationException` and is not caught, so the
call fails with an unexpected server error (HTTP 500) instead of a clean
deserialization failure.

The `ServerCustomFieldSerializer` javadoc states this is the serializer's own
responsibility:

> It is this method's responsibility to verify the types of objects that it
> reads. ... In practice, any call to `ServerSerializationStreamReader.readObject()`
> should use the type checking version, passing in the expected type of the
> object to be read. See the built-in GWT server custom field serializers for
> examples.

## Reproduction

Any RPC service with an `EnumMap` parameter is affected. With a service method
`void store(EnumMap settings)` and a serialization policy that allow-lists the
service's ordinary types (`EnumMap`, its enum key type, `Integer`, `String`), an
unauthenticated POST whose EnumMap exemplar is an allow-listed `Integer` instead
of an enum constant returns HTTP 500, and the server log shows the uncaught
`NullPointerException`. With the exemplar read through the type-checking path the
same request is rejected cleanly.

The existing type-check test family
(`RPCTypeCheckTest` / `RPCTypeCheckCollectionsTest` / `RPCTypeCheckArraysTest`)
covers every other collection serializer for exactly this substitution
(`testMapHashSetSpoofingMap`, `testVectorHashSetSpoofingVector`, and so on) but
has no case for `EnumMap`.

I have a fix that mirrors `TreeMap`/`TreeSet` plus a test in that family; I will
open a PR referencing this issue.

Nexory:
Fixes #10391.

`EnumMap_ServerCustomFieldSerializer` overrode the type-checking server
`instantiateInstance` but threw the type information away and called the untyped
client `instantiate`, which reads the exemplar with a bare `readObject()`. Every
other collection serializer reads what it instantiates from through the typed
path; `TreeMap`/`TreeSet` are the direct precedent
(`readObject(Comparator.class, resolvedTypes)`).

Because the exemplar was read untyped, a request could substitute another type
for it, and `new EnumMap(nonEnumClass)` throws an uncaught
`NullPointerException` (`keyUniverse is null`) instead of a clean serialization
error.

This adds a server-side `instantiate(streamReader, expectedParameterTypes,
resolvedTypes)` that reads the exemplar against `expectedParameterTypes[0]`, the
declared key type. A non-enum is rejected with `SerializedTypeViolationException`
the same way the other serializers reject mismatched types, and so is an enum
other than the declared one. A legitimate exemplar still passes and the resulting
`getClass()` is unchanged, so behaviour for valid EnumMaps is identical.

## Tests

Four cases in `RPCTypeCheckTest`, all going through `RPC.decodeRequest`:

- `testEnumMapSpoofingClass` substitutes an `Integer` exemplar on the existing raw
  `EnumMap` parameter. Before this change it fails with an uncaught
  `NullPointerException`; after it, the request is rejected with
  `IncompatibleRemoteServiceException` wrapping `SerializedTypeViolationException`,
  matching the existing spoofing tests for the other collections.
- `testEnumMapSpoofingEnum` substitutes an exemplar of a different enum on a
  parameter declared `EnumMap<NEnum, Integer>` and expects the same rejection.
  This is the case that separates `expectedParameterTypes[0]` from `Enum.class`:
  reverting only that argument leaves the test failing at its `fail()` call,
  because `Enum.class` accepts any enum.
- `testEnumMapTypedValid` decodes a request whose exemplar is the declared key
  type and asserts the resulting `EnumMap`, so the tighter check is shown not to
  reject the legitimate case.
- `testEnumMapValid` decodes an `EnumMap` with an `NEnum` exemplar and one with an
  `OEnum` exemplar on the raw parameter, which declares no key type, and asserts
  both maps. A key type pinned to one enum passes every other case and fails this
  one, on the `OEnum` request.

The raw parameter is kept alongside the typed one, so both paths stay covered.
`RPCTypeCheckTest` passes with 29 tests, and the rest of
`com.google.gwt.user.server.rpc` is unchanged.

## Diff

- `EnumMap_ServerCustomFieldSerializer.java` (+16/-1): the typed instantiate.
- `RPCTypeCheckTest.java` (+158): two enums, a typed test service method, the
  four tests and their request generators.
- `RPCTypeCheckFactory.java` (+46): helpers to write an enum constant and an
  EnumMap message with a chosen exemplar, plus a javadoc that had ended up
  separated from its method.

## Note

The failure is reachable through the normal servlet path, not only through
`RPC.decodeRequest` in a test. `processCall` guards the decode against
`IncompatibleRemoteServiceException` only, so the `NullPointerException`
propagates to `doPost`, which catches `Throwable` and answers through
`RPCServletUtils.writeResponseForUnexpectedFailure`, that is with status 500. I
also confirmed it with an embedded `RemoteServiceServlet` and a strict
serialization policy that allow-lists only the service's own types.


user/src/com/google/gwt/user/server/rpc/core/java/util/EnumMap_ServerCustomFieldSerializer.java
```diff
@@ -68,11 +68,26 @@ public EnumMap instantiateInstance(SerializationStreamReader streamReader)
...
+     * The key type of an {@link EnumMap} is always an enum. Read the exemplar
+     * object with a type check, as the other server custom field serializers
+     * do, so that a non-enum type cannot be substituted for it. Without the
+     * check the substituted type reaches {@code new EnumMap(nonEnumClass)},
+     * which throws an uncaught {@link NullPointerException} rather than a clean
+     * serialization error.
+     */
+    @SuppressWarnings("unused")
+    public static EnumMap instantiate(ServerSerializationStreamReader streamReader,
+        Type[] expectedParameterTypes, DequeMap<TypeVariable< ? >, Type> resolvedTypes)
```
niloc132:
```suggestion
        Type[] expectedParameterTypes, DequeMap<TypeVariable<?>, Type> resolvedTypes)
```

user/test/com/google/gwt/user/server/rpc/RPCTypeCheckTest.java
```diff
@@ -2250,6 +2271,24 @@ public void testEmptyListSpoofingClass() {
    * This checks that a Map generated by Collections.emptyMap correctly reports
    * that it is an incorrect type.
    */
+  /**
```
niloc132: looks like you separated the javadoc above from its method below

user/src/com/google/gwt/user/server/rpc/core/java/util/EnumMap_ServerCustomFieldSerializer.java
```diff
@@ -68,11 +68,26 @@ public EnumMap instantiateInstance(SerializationStreamReader streamReader)
...
 
+    /**
+     * The key type of an {@link EnumMap} is always an enum. Read the exemplar
+     * object with a type check, as the other server custom field serializers
+     * do, so that a non-enum type cannot be substituted for it. Without the
+     * check the substituted type reaches {@code new EnumMap(nonEnumClass)},
+     * which throws an uncaught {@link NullPointerException} rather than a clean
+     * serialization error.
+     */
+    @SuppressWarnings("unused")
```
niloc132:
```suggestion
```
This method is used, you are calling it below.

user/src/com/google/gwt/user/server/rpc/core/java/util/EnumMap_ServerCustomFieldSerializer.java
```diff
@@ -68,11 +68,26 @@ public EnumMap instantiateInstance(SerializationStreamReader streamReader)
...
+    public static EnumMap instantiate(ServerSerializationStreamReader streamReader,
+        Type[] expectedParameterTypes, DequeMap<TypeVariable< ? >, Type> resolvedTypes)
+        throws SerializationException {
+      Object exemplar = streamReader.readObject(Enum.class, resolvedTypes);
```
niloc132: Shouldn't this really be expectedParameterTypes[0]? We don't need to allow _any_ enum here, we can be specific to the actual enum being used.

Tests appear to pass with this additional change, but you might have a more specific use case to consider than the raw test below.

user/test/com/google/gwt/user/server/rpc/RPCTypeCheckTest.java
```diff
@@ -196,6 +196,10 @@ public static class ClassesParamTestClass<T extends List<Integer> & MInterface<S
     public static void testAClass(AClass arg1) {
     }
 
+    @SuppressWarnings({"unused", "rawtypes"})
+    public static void testEnumMap(java.util.EnumMap arg1) {
+    }
+
```
niloc132, line 209: While not technically illegal, using raw generics does limit rpc from doing more accurate typechecking, so we're missing out on testing that. As an example, if `expectedParameterTypes[0]` were used above, we should be able to tell the difference between the right and wrong enum type, even if both are enums - but only if the correct type is actually specified.

niloc132 (commented):
Given that the only bug being fixed here is that bad (malicious?) input results in a slightly less useful error message, this seems low priority.

user/src/com/google/gwt/user/server/rpc/core/java/util/EnumMap_ServerCustomFieldSerializer.java
Nexory: Done in 7f288433. The other three occurrences in this file keep the spaced form
they had before this PR, so the file is mixed now; happy to normalise all four
if you would rather have that.

user/test/com/google/gwt/user/server/rpc/RPCTypeCheckTest.java
Nexory: Fixed in 1d9eaf26. The javadoc belonged to `testEmptyMapSpoofingClass` and my
test went in between it and its method. While I was there I dropped the attack
framing from my own test comment, since what it exercises is the type check.

user/src/com/google/gwt/user/server/rpc/core/java/util/EnumMap_ServerCustomFieldSerializer.java
Nexory: Removed in a7360f10. It was there because `expectedParameterTypes` went unused
inside the method, not because the method is unused, and using it below made
the annotation unnecessary anyway. `TreeMap_ServerCustomFieldSerializer`
ignores the same parameter without one, so it should never have been added.

user/src/com/google/gwt/user/server/rpc/core/java/util/EnumMap_ServerCustomFieldSerializer.java
Nexory: Applied in a7360f10, and you are right about the effect, but it is not
observable in this PR as it stands.

`expectedParameterTypes[0]` here is the type variable `K`, not a class. I
printed it: `len=2 [0]=K`. With the raw `EnumMap` in the test service method,
`K` resolves to its bound, so `readObject` sees `Enum` and behaves exactly as
the old `Enum.class` did.

Measured on this branch: reverting just that one line back to `Enum.class`
leaves `RPCTypeCheckTest` at 26/26 green. Nothing currently covers the change,
which is the same thing you point out below.

user/test/com/google/gwt/user/server/rpc/RPCTypeCheckTest.java
Nexory: Agreed, and it is the reason the change above is untestable right now: with the
raw declaration `K` erases to `Enum`, so right and wrong enum are
indistinguishable.

Parameterising it needs an enum in that test's type universe, and there is not
one anywhere under `user/test/com/google/gwt/user/server/rpc/`. Adding one means
introducing it into the RPC allowlist for `ClassesParamTestClass`, which is more
than a test tweak.

Do you want me to add one, or leave the tighter check in with the raw test and
handle the coverage separately?

niloc132: Yes, the new behavior should be tested. Please ensure there is a positive test as well.

Nexory: Done in a3429a1b, rebased onto current main. The service method now declares `EnumMap<NEnum, Integer>`, so
`expectedParameterTypes[0]` is a class rather than the type variable, and two enums
make the right and the wrong one distinguishable.

Negative: `testEnumMapSpoofingEnum` substitutes an `OEnum` exemplar and expects the
type violation. Positive: `testEnumMapTypedValid` decodes a request whose exemplar is
the declared key type and asserts the resulting map.

Measured both ways out of the same build. As it stands, `RPCTypeCheckTest` is 28/28.
Reverting only the `readObject` argument back to `Enum.class`:

    Tests run: 28,  Failures: 1,  Errors: 0
    1) testEnumMapSpoofingEnum(...RPCTypeCheckTest)junit.framework.AssertionFailedError:
       Expected IncompatibleRemoteServiceException from testEnumMapSpoofingEnum

It fails at the `fail()` call and not as an error, so `decodeRequest` returned
normally: `Enum.class` does accept the other enum. `testEnumMapTypedValid` is green in
both runs, which is the point of having it.

Two details in case they look arbitrary. I measured both rather than assuming them. If
`OEnum` does not implement `IsSerializable`, the reverted run still fails, but on the
serialization policy: `expected:<SerializedTypeViolationException> but
was:<SerializationException>`, which is red for a reason unrelated to the line. If the
negative case carries a map entry instead of an empty body, the reverted run is green:
the entry's key is checked against the declared key type and raises the violation by
itself, so the test would be measuring the map key check rather than the exemplar read.
The empty body is what isolates the exemplar.

The raw `testEnumMap` and its Integer-exemplar test are untouched, so that path stays
covered. Production code is unchanged in this round; both commits are test only.

Separately, d27fb93b moves `writeEnumMapWithSpoofedExemplar` in `RPCTypeCheckFactory`:
I had inserted it between the javadoc of `write(int)` and the method, the same thing you
spotted in the test file.

user/src/com/google/gwt/user/server/rpc/core/java/util/EnumMap_ServerCustomFieldSerializer.java
Nexory: Correction to what I wrote here: `TreeMap_ServerCustomFieldSerializer` does carry
`@SuppressWarnings("unused")` on its `instantiate`, and so does
`TreeSet_ServerCustomFieldSerializer`. I had it backwards.

Removing the annotation was still right, since `expectedParameterTypes[0]` is used now,
but the reason I gave for it was not. The commit message of a7360f10f carries the same
wrong claim.

niloc132: Thanks - commit message isn't important, we'll squash the commits and use the PR description instead (so be sure that is all accurate and ready for merge).

user/src/com/google/gwt/user/server/rpc/core/java/util/EnumMap_ServerCustomFieldSerializer.java
niloc132: We can fix those when the code gets touched, no need up update it all at once.

user/test/com/google/gwt/user/server/rpc/RPCTypeCheckTest.java
```diff
@@ -2246,6 +2321,51 @@ public void testEmptyListSpoofingClass() {
...
+  /**
+   * This checks that an EnumMap whose exemplar is the declared key type is
+   * accepted, and that it is built with that key type.
+   */
+  public void testEnumMapTypedValid() throws Exception {
+    RPCRequest decoded = RPC.decodeRequest(generateEnumMapTypedValid());
+    Object deserializedArg = decoded.getParameters()[0];
+    assertEquals(EnumMap.class, deserializedArg.getClass());
+
+    EnumMap<NEnum, Integer> expected = new EnumMap<NEnum, Integer>(NEnum.class);
```
niloc132:
```suggestion
    EnumMap<NEnum, Integer> expected = new EnumMap<>(NEnum.class);
```

user/test/com/google/gwt/user/server/rpc/RPCTypeCheckTest.java
```diff
@@ -2246,6 +2321,51 @@ public void testEmptyListSpoofingClass() {
...
+      assertEquals(SerializedTypeViolationException.class, e.getCause().getClass());
+      assertTrue(e.getCause().getMessage().matches(".*OEnum.*NEnum.*"));
+    }
+  }
+
+  /**
+   * This checks that an EnumMap whose exemplar is the declared key type is
+   * accepted, and that it is built with that key type.
+   */
+  public void testEnumMapTypedValid() throws Exception {
```
niloc132:
```suggestion
  public void testEnumMapTypedValid() {
```

user/test/com/google/gwt/user/server/rpc/RPCTypeCheckFactory.java
```diff
@@ -405,6 +405,51 @@ public void writeEmptySet() {
...
+  }
+
+  /**
+   * Add data for an EnumMap whose exemplar object (the value from which the
+   * server derives the enum key type) is the given enum constant, and which has
+   * a single entry.
+   */
+  public void writeEnumMapWithEntry(Enum<?> exemplar, Enum<?> key, Integer value)
+      throws SerializationException {
+    writeStringFromTable(generateSerializedClassString(java.util.EnumMap.class));
```
niloc132: import this so it doesnt need to be qualified

```suggestion
    writeStringFromTable(generateSerializedClassString(EnumMap.class));
```

user/test/com/google/gwt/user/server/rpc/RPCTypeCheckTest.java
```diff
@@ -548,6 +574,55 @@ public static void testIntString(int arg1, String arg2) {
     }
   }
 
+  private static String generateEnumMapSpoofingClass() {
+    try {
+      RPCTypeCheckFactory strFactory =
+          new RPCTypeCheckFactory(ClassesParamTestClass.class, "testEnumMap");
```
niloc132, line 580: looks like we still need a postive test for this, probably something that exercises both NEnum and OEnum?

niloc132 (commented):
Thanks for the changes, some feedback below.

user/test/com/google/gwt/user/server/rpc/RPCTypeCheckTest.java
Nexory: Added in 6637070f as `testEnumMapValid`, on the raw `testEnumMap(EnumMap)` parameter
this generator targets. It decodes two requests there, one with an `NEnum` exemplar and
key, one with an `OEnum` exemplar and key, and for each asserts an `EnumMap` equal to the
expected map. Each map carries an entry, and `EnumMap` only treats maps with entries as
equal when their key types match, so a map built for the wrong enum would not pass.

To check that the `OEnum` half earns its place, I pinned the serializer's key type to
`NEnum`, a deliberately too-strict variant, and ran `RPCTypeCheckTest` against it:

    Tests run: 29,  Failures: 0,  Errors: 1
    testEnumMapValid: IncompatibleRemoteServiceException: Attempt to deserialize an
    object of type class ...RPCTypeCheckTest$OEnum when an object of type class
    ...RPCTypeCheckTest$NEnum is expected

The error comes from the `OEnum` request. The `NEnum` half and every other test in the
class pass under that variant, so the second request is what catches a key type pinned
to one enum. With the serializer restored, `RPCTypeCheckTest` is 29/29.

I also updated the PR description for the fourth test, and rebased onto current main so
the tests ran against it.

user/test/com/google/gwt/user/server/rpc/RPCTypeCheckTest.java
Nexory: Applied in 088c3864.

user/test/com/google/gwt/user/server/rpc/RPCTypeCheckTest.java
Nexory: Applied in 088c3864.

user/test/com/google/gwt/user/server/rpc/RPCTypeCheckFactory.java
Nexory: Applied in 088c3864, at all three uses in `RPCTypeCheckFactory`.

user/test/com/google/gwt/user/server/rpc/RPCTypeCheckFactory.java
```diff
@@ -405,6 +406,51 @@ public void writeEmptySet() {
     writeStringFromTable(generateSerializedClassString(emptySet.getClass()));
   }
 
+  /**
+   * Add data for an enum constant. The type written is the declaring class, and
+   * the value is the constant's ordinal.
+   */
+  public void writeEnum(Enum<?> value) {
+    writeStringFromTable(generateSerializedClassString(value.getDeclaringClass()));
+    bodyString += Integer.toString(value.ordinal()) + RPC_SEPARATOR_CHAR;
```
niloc132, line 416: suppress this warning for the method, ordinal is how we write enums to the wire (since the class string will include the crc of the fields)

Nexory: Suppressed for the method in 36f3acb6. With a rebuild of the tests under Error Prone, no warning is left on the lines this PR changes.

Issue #9990, opened 2024-07-31
Provide options for ScriptInjector/CssResource/GssResource to work with CSP headers


niloc132:
CssResource/GssResource inject their contents inline into `<style>` tags using the StyleInjector, and do not presently have any accommodation for CSP rules - that is, if CSP is enabled, `unsafe-inline`/etc must be specified for this to still work.

In theory, we could hash all of the included CssResource blobs of content, and provide them in a generated file, so that a server can provide them and indicate that those inline styles are safe because they match an expected hash. We could also offer a way to pass in a nonce, so that each style tag when created would have a supported nonce for that page load - though I'm a little skeptical of how this would actually be safe if the value is readable by plain JS.

smongia-apptio:
Do we know if someone is considering this request?

niloc132:
To my knowledge, no one is actively working on it - if you're considering picking up the task (or otherwise sponsoring it etc), go for it.

frank70321-hub:
Following the review feedback on #10393, I agree that carrying a nonce onto a generated `<style>` element must not be read as attesting that the stylesheet content itself is safe. The current patch only copies the first existing script nonce so that `StyleInjector` output can pass a nonce-based CSP; it does not validate or sanitize the CSS string.

The main behavior choices seem to be:

1. `ignore`: preserve today's behavior and do not set a nonce.
2. `propagate-nonce`: copy an existing page nonce automatically, as the current patch does.
3. `explicit-nonce`: accept a configured nonce value or add an injection overload that takes one explicitly.

An overload alone would not cover every `CssResource` injection path, while automatic propagation can imply more trust than it provides. My current inclination is an opt-in configuration whose default remains `ignore`, but I would like maintainer direction on the API shape and expected `CssResource` behavior before revising the implementation. I have paused code changes pending that decision.

jnehlmeier:
I feel like the `ScriptInjector` solution is already a mistake and should not have been merged. If you enable CSP `script/style-src` the GWT implementation gives you a false sense of security. 

The code can happily do `ScriptInjector.fromUrl("https://hacked.cdn.com/lib/script.js")`. Same with Css if we would copy the same solution.

I would strongly prefer a hash based solution so the hashes are shipped with the GWT app as the developer has to provide them via a new API that takes a hash as parameter as well. GWT ClientBundle can precompute the hashes and then also use that new API.

niloc132:
I'm worried we might want a more general approach here.

[...]

PR #10393, opened 2026-08-13
Propagate CSP nonce to StyleInjector elements

frank70321-hub:
StyleInjector creates inline `<style>` elements for CssResource and GssResource output. Under a nonce-based `style-src` Content Security Policy, those elements need the page's nonce to be accepted.

This change mirrors the existing ScriptInjector behavior: when the current document contains a `script[nonce]`, StyleInjector copies that nonce to each newly created style element. If no nonce source exists, the injected style element is left unchanged.

The focused HtmlUnit coverage verifies both paths:

- a style element inherits an existing script nonce;
- no nonce attribute is created when the document has no nonce source.

Local verification:

- `StyleInjectorTest`: 6 tests, 0 failures, 0 errors;
- user source and test checkstyle: 0 violations in the changed files;
- `ant buildonly`: successful;
- `git diff --check`: successful.

This addresses the nonce option only; the hash-based option discussed in the issue remains separate.

Partial #9990


niloc132:
I think this looks right - if some other script can read the nonce and assign to its own scripts/stylesheets then by definition that other script _can already run_ and is unsafe to begin with.

This doesn't guarantee the safety of the stylesheet being injected though - StyleInjector will happily take a string from the URL (query params etc) even if entirely unsafe. That's the only thing I would feel nervous about here, that we could be implying that the script is known to be safe. 

Off the top of my head, we could make this configurable (e.g. `ignore`, `propagate-nonce`, `the-nonce-is-this-string`), or add another overload for inject to specify the nonce (doesn't help css resource cases though). I think this could be worth more discussion in the issue prior to implementation.

frank70321-hub:
Agreed. The nonce only lets the generated style pass nonce-based CSP; it does not make an untrusted CSS string safe, and the default behavior should not imply otherwise.

I summarized `ignore`, automatic propagation, and explicit/configured nonce options in #9990, including the `CssResource` limitation of an overload-only approach. I have paused implementation changes until the intended API shape and default behavior are decided there.

PR #10396, opened 2026-08-19
Parallelize CodeSplitter's not-exclusive CFA computation

### LINKED ISSUE #10395: CodeSplitter's not-exclusive CFA computation is quadratic in fragment count
`CodeSplitter.computeNotExclusiveCfaForFragments` builds, for every exclusive fragment, a `ControlFlowAnalyzer` that traverses the run-asyncs of every other fragment — F² traversals.

With 122 split points that is ~15,000 traversals; with 231, ~53,000. On a real application the loop alone is 25.6% of the permutation's CPU, and CodeSplitter is 181s of an 11-minute compile.

The iterations are independent and can run in parallel: the CFA copy constructor deep-copies every mutable set and the traversal only reads the JProgram. Doing so takes CodeSplitter from 181.4s to 71.5s, with byte-identical output.

Note that` -XfragmentCount `does not help here — the merge strategy replaces the F² loop with an O(F) per-run-async CFA loop of roughly the same cost.

rhuanhianc:
`computeNotExclusiveCfaForFragments` is quadratic in the number of exclusive fragments: for each fragment it traverses the run-asyncs of every other fragment. Applications with many split points pay heavily — with 231 split points this loop alone is 25.6% of the permutation's CPU, and CodeSplitter overall is 181s of an 11-minute compile.

The iterations are independent. `ControlFlowAnalyzer`'s copy constructor deep-copies every mutable set, each traversal writes only to its own analyzer, and `ControlFlowAnalyzer` is purely analytical — it contains no setters on AST nodes. This runs them on a fixed thread pool.

Two constraints are respected:

- Dependency-graph recording is stateful and order sensitive, so the parallel path is only taken when no recorder is installed. With `-compileReport` the original serial loop runs unchanged.
- `JProgram.getTypeArray` lazily creates array types in a plain HashMap and is reachable from the traversal, so it is now `synchronized`. It is the only shared mutable state on this path; `getAllArrayTypes`, which already sorts to avoid nondeterminism, is not reachable from `traverseFromRunAsync`.

Happy to derive the pool size from `-localWorkers` instead of `availableProcessors()` if you'd prefer to avoid oversubscription when permutations are already compiled in parallel.

CodeSplitter drops from 38.4s to 17.0s with 122 split points and from 181.4s to 71.5s with 231. Generated JavaScript is byte-for-byte identical across 874 output files from four applications. ant -Dtarget=test dev passes: 1936 tests, 0 failures.

Fixes #10395

### DIFF
--- a/dev/core/src/com/google/gwt/dev/jjs/ast/JProgram.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/ast/JProgram.java
@@ -1023,7 +1023,8 @@ public class JProgram extends JNode implements ArrayTypeCreator {
     return staticImpl;
   }
 
-  public JArrayType getTypeArray(JType elementType) {
+  // Synchronized because CodeSplitter reaches this from concurrent CFA traversals.
+  public synchronized JArrayType getTypeArray(JType elementType) {
     JArrayType arrayType = arrayTypes.get(elementType);
     if (arrayType == null) {
       arrayType = new JArrayType(elementType);
--- a/dev/core/src/com/google/gwt/dev/jjs/impl/codesplitter/CodeSplitter.java
+++ b/dev/core/src/com/google/gwt/dev/jjs/impl/codesplitter/CodeSplitter.java
@@ -43,6 +43,11 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.concurrent.Callable;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 
 /**
  * <p>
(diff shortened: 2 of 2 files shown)


dev/core/src/com/google/gwt/dev/jjs/ast/JProgram.java
```diff
@@ -1023,7 +1023,8 @@ public JMethod getStaticImpl(JMethod method) {
     return staticImpl;
   }
 
-  public JArrayType getTypeArray(JType elementType) {
+  // Synchronized because CodeSplitter reaches this from concurrent CFA traversals.
+  public synchronized JArrayType getTypeArray(JType elementType) {
```
niloc132, line 1027: Instead, perhaps just call `program.getTypeArray(JPrimitiveType.INT)` once before branching so we guarantee the type exists. I think we could even drop that call entirely, since it only exists for an assert (and once it matches a `==` check, we just use the same instance).

niloc132 (commented):
I'm not sure about this - at the very least, tying it to localWorkers will take the sting off of it a bit, but we still risk actually running localWorkers^2 concurrent threads (and running out of memory by doing so much at once, or other issues from too much work scheduled).

For threaded workers, we could at least share a single ExecService with extra localWorker count (e.g. `localWorkers - permutationCount` threads haven't been used yet, let them be scheduled for this work, plus or minus how many permutations are paused waiting for their exclusive fragments). But when we use external processes for workers (e.g. ExternalPermutationWorkerFactory), we don't have as nice of a way to communicate about which process should get the extra threads...

Mechanically I don't much to complain about though, just want to consider this approach carefully before risking breaking anyone's CI.

PR #10397, opened 2026-08-19
Memoize JsInliner.containsNestedFunctions

### LINKED ISSUE #10394: JsInliner re-traverses whole function bodies once per call site
`JsInliner.containsNestedFunctions` traverses a function's entire body, and is called once per candidate call site through
`isInlinable → isVolatile → affectedBySideEffects` ,on every inliner pass.

The predicate is a pure function of the body, so the work is redundant. On two large applications (122 and 231 split points), JFR sampling with full stack depth attributes ~79% of all JsInliner time to this single predicate; JsInliner itself is 28.7% of the permutation's CPU and 36–48% of total compile time.

Memoizing it takes JsInliner from 57.9s to 10.5s and from 328.5s to 22.8s on those two applications, with byte-identical output.

rhuanhianc:
`containsNestedFunctions` is a pure function of a function's body, but it is recomputed — a full traversal of the entire body — once per candidate call site, on every inliner pass. On a large application, JFR sampling attributes about 79% of all JsInliner time to this single predicate, and JsInliner itself accounts for 28.7% of the permutation's CPU.

This memoizes it in a thread-local `IdentityHashMap`, cleared at the start of each execImpl so it never outlives a single run of this optimizer, and invalidated per function wherever the inliner rewrites that function's body.

Invalidation happens at both points where a body actually changes: in` process() `when an inline is committed, and again right after ctx.replaceMe(op). The second is necessary because the intervening op = accept(op) can re-populate the memo for the caller while op is still detached from the tree.

`endVisit(JsExprStmt) `also mutates bodies, but only by removing or restructuring content already present in that statement, so it can only take the answer from true to false. That is the conservative direction: a stale true leaves `safeScope` null and makes the inliner more cautious, never less.

Measured on two real applications, JsInliner drops from 57.9s to 10.5s and from 328.5s to 22.8s. Generated JavaScript is byte-for-byte identical across 874 output files from four applications, including two built with two permutations (4 and 20 local workers). `ant -Dtarget=test dev `passes: 1936 tests, 0 failures.

Fixes #10394

### DIFF
--- a/dev/core/src/com/google/gwt/dev/js/JsInliner.java
+++ b/dev/core/src/com/google/gwt/dev/js/JsInliner.java
@@ -1601,15 +1601,11 @@ public class JsInliner {
    * Examine a JsFunction to determine if it contains nested functions.
    */
   private static boolean containsNestedFunctions(JsFunction func) {
-    Boolean cached = containsNestedFunctionsCache.get().get(func);
-    if (cached != null) {
-      return cached;
-    }
-    NestedFunctionVisitor v = new NestedFunctionVisitor();
-    v.accept(func.getBody());
-    boolean result = v.containsNestedFunctions();
-    containsNestedFunctionsCache.get().put(func, result);
-    return result;
+    return containsNestedFunctionsCache.get().computeIfAbsent(func, function -> {
+      NestedFunctionVisitor v = new NestedFunctionVisitor();
+      v.accept(function.getBody());
+      return v.containsNestedFunctions();
+    });
   }
 
   private static int execImpl(JsProgram program, Collection<JsNode> toInline) {
--- a/dev/core/src/com/google/gwt/dev/js/JsInliner.java
+++ b/dev/core/src/com/google/gwt/dev/js/JsInliner.java
@@ -1024,8 +1024,6 @@ public class JsInliner {
         return x;
       }
 
-      containsNestedFunctionsCache.get().remove(callerFunction);
-
       // We've committed to the inlining, ensure the vars are created
       newLocalVariableStack.peek().addAll(extrudedNames);
 
--- a/dev/core/src/com/google/gwt/dev/js/JsInliner.java
+++ b/dev/core/src/com/google/gwt/dev/js/JsInliner.java
@@ -66,7 +66,6 @@ import com.google.gwt.thirdparty.guava.common.collect.Sets;
 
 import java.util.Collection;
 import java.util.Collections;
-import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.ListIterator;
 import java.util.Map;
@@ -639,6 +638,14 @@ public class JsInliner {
      */
     private JsFunction programFunction;
 
+    /**
+     * Caches {@link #containsNestedFunctions(JsFunction)}, which would otherwise re-traverse a
+     * whole function body at every call site. Invalidated for a function whenever this visitor
+     * rewrites that function's body.
+     */
+    private final Map<JsFunction, Boolean> containsNestedFunctionsCache =
+        Maps.newIdentityHashMap();
+
     public InliningVisitor(JsProgram program, Set<JsNode> whitelist) {
       this.whitelist = whitelist;
       invocationCountingVisitor.accept(program);
@@ -804,7 +811,7 @@ public class JsInliner {
         op = accept(op);
         ctx.replaceMe(op);
         // The accept above may have re-cached the caller while op was still detached.
-        containsNestedFunctionsCache.get().remove(callerFunction);
+        containsNestedFunctionsCache.remove(callerFunction);
       }
 
       if (inlining.pop() != invokedFunction) {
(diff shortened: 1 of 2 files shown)


dev/core/src/com/google/gwt/dev/js/JsInliner.java
```diff
@@ -1589,12 +1601,19 @@ private static int complexity(JsNode toEstimate) {
    * Examine a JsFunction to determine if it contains nested functions.
    */
   private static boolean containsNestedFunctions(JsFunction func) {
+    Boolean cached = containsNestedFunctionsCache.get().get(func);
```
zbynek: You can use `computeIfAbsent` here.

rhuanhianc: Thanks, updated to use computeIfAbsent.

dev/core/src/com/google/gwt/dev/js/JsInliner.java
```diff
@@ -1549,6 +1554,13 @@ public boolean visit(JsObjectLiteral x, JsContext ctx) {
   private static final int INLINING_BIAS = Integer.parseInt(System.getProperty(
       "gwt.jsinlinerInliningBias", "5"));
 
+  /**
+   * Caches {@link #containsNestedFunctions(JsFunction)}, which would otherwise re-traverse a whole
+   * function body at every call site. Thread local because permutations compile concurrently.
+   */
+  private static final ThreadLocal<Map<JsFunction, Boolean>> containsNestedFunctionsCache =
```
niloc132: Wouldn't an instance variable make more sense? This will end up sticking around as long as the thread itself, even if the inliner isn't invoked again.

Granted, this would require other changes - isInlinable, isVolatile, and affectedBySideEffects will need to stop being static (or pass it in as a parameter), but I'm not really sure thats a bad thing - an instance already exists, and these are never called outside the context of having an instance as far as I can tell.

This could also be achieved by moving the clear() to the end of execImpl instead of the start, but I'd mildly prefer the instance var and avoid any concerns about threading.

dev/core/src/com/google/gwt/dev/js/JsInliner.java
```diff
@@ -802,6 +803,8 @@ public void endVisit(JsInvocation x, JsContext ctx) {
          */
         op = accept(op);
         ctx.replaceMe(op);
+        // The accept above may have re-cached the caller while op was still detached.
+        containsNestedFunctionsCache.get().remove(callerFunction);
```
niloc132: I'm not quite understanding this remark - this seems to imply that there's a risk of a function being added or removed during `process`? Probably would be worth temporarily reverting this, writing a (failing) test, and confirming that the test now passes?

dev/core/src/com/google/gwt/dev/js/JsInliner.java
```diff
@@ -1021,6 +1024,8 @@ && isInvokedMoreThanOnce(invokedFunction)) {
         return x;
       }
 
+      containsNestedFunctionsCache.get().remove(callerFunction);
```
niloc132: Without a comment (like the other remove() func has), it seems like this is the same basic idea - invalidate the cache since we're changing the function in question. 

However, this is _before_ the function has been modified - the other site is just after the replaceMe(), so it seems that if accept() might re-add items, we shouldn't bother with this line at all?

niloc132 (commented):
Thanks for spotting and fixing this. I confirmed locally, and was surprised to see that the other uncached values (esp complexity) aren't really worth caching too.

I added a few notes - esp where your comments suggest edge cases, it would be nice to have tests to confirm that they are properly resolved.

dev/core/src/com/google/gwt/dev/js/JsInliner.java
rhuanhianc: Thanks @niloc132 , done. The cache is a field on `InliningVisitor` now, and `isInlinable`, `isVolatile`, `affectedBySideEffects` and `containsNestedFunctions` came along with it as instance methods, bodies untouched.

You were right that nothing outside needs them static: `process` is the only entry into that chain and it has a single call site, so dropping `static` cost nothing. The `ThreadLocal` and the `clear()` in `execImpl` are both gone, and with them the retention you pointed at.

dev/core/src/com/google/gwt/dev/js/JsInliner.java
rhuanhianc: I tried to write that failing test like you suggested and couldn't, and after digging into why, my note there was just wrong. The case can't happen at all.

`convertToExpression` clones every callee statement through `JsSafeCloner`, and that returns null for anything holding a `JsFunction`. So a function with a nested function inside is never inlinable in the first place, and the body that replaces the invocation never carries a function into the caller. The only other thing that moves in is the caller's own arguments, which were already there — and an argument holding a function is always volatile (`AffectedBySideEffectsVisitor.endVisit(JsFunction)`), so it goes into `requiredOrder`, and `EvaluationOrderVisitor.maintainsOrder()` needs it referenced in the inlined body, so it can't get dropped either.

The answer can only go `true → false`, and that's the safe direction.

Rather than just delete the line and leave the reasoning in a comment, I made `containsNestedFunctions` recompute and compare on every cache hit under an `assert`. The dev tests run with `-ea`, so now it's checked on every lookup of every test instead of by one contrived case.

I did check it before removing anything, too. An instrumented build that recomputed and compared on every lookup, with the invalidation disabled, reported 6,683 cache hits and zero mismatches compiling one application, and a full run over a large one with the assert on passes as well — the inliner finishes well before `JavaScriptVerifier` runs.

And the two tests you asked for are in. Both answers of the predicate show up as a difference in the generated output, so forcing it to a constant breaks one of them.

dev/core/src/com/google/gwt/dev/js/JsInliner.java
rhuanhianc: That's right, and it's easy to check: past that line `process` has no `return x` left, and the expression it builds is never the same object as `x`, so the `if (x != op)` branch in `endVisit` always ran and invalidated the caller anyway. Nothing in between touches the caller's body either, since `ctx.replaceMe` hasn't happened yet, so recomputing there gave back the same value. It was a pure re-traversal. Removed.

The other one is gone too, for the reason in the other thread.

niloc132:
I'll take another look as soon as I'm able, thanks for updating this. The diff grew a bit with needing to rearrange the functions, so it might take a little longer.

dev/core/src/com/google/gwt/dev/js/JsInliner.java
```diff
@@ -1058,6 +1060,172 @@ private boolean isTooComplexToInline(JsInvocation x, JsExpression op) {
...
+
+    /**
+     * Examine a JsFunction to determine if it contains nested functions.
+     *
+     * <p>Memoized, since the answer would otherwise be recomputed at every call site. An inlined
+     * body is a {@link JsSafeCloner} clone and that cloner rejects anything holding a function,
+     * so inlining never moves a function into the caller it rewrites.
+     */
+    private boolean containsNestedFunctions(JsFunction func) {
+      Boolean cached = containsNestedFunctionsCache.computeIfAbsent(
```
niloc132: I think this should be safe to make `boolean` - the provided method ref can never return null, so computeIfAbsent is null safe.

dev/core/src/com/google/gwt/dev/js/JsInliner.java
```diff
@@ -1058,6 +1060,172 @@ private boolean isTooComplexToInline(JsInvocation x, JsExpression op) {
...
+          func, InliningVisitor::computeContainsNestedFunctions);
+      assert cached == computeContainsNestedFunctions(func) : "Stale nested function memo";
```
niloc132, line 1096: let's add an explicit "this is expensive and defeats the purpose of memoizing" comment here, so that no one tries to get clever with upgrading it to an IllegalStateException or removing the assert outright without proper thought.

niloc132 (commented):
Two very minor notes, happy for pushback on either one.

dev/core/src/com/google/gwt/dev/js/JsInliner.java
rhuanhianc: Done!  is a primitive `boolean` now

dev/core/src/com/google/gwt/dev/js/JsInliner.java
rhuanhianc: Comment added!

niloc132 (approved):

Issue #9967, opened 2024-06-04
Remove onLoadErrorFn/onPropertyErrorFn, or make them easier to disable


niloc132:
There are [two undocumented hooks in the built-in linkers](https://github.com/gwtproject/gwt/issues/8130), allowing meta tags to contribute JS snippets to be eval'd at runtime and provide error handling functions. 

- In CrossSiteIframeLinker and subtypes, users can disable this feature outright by specifying that `processMetasNull.js` should be loaded instead of `processMetas.js` by overriding getJsProcessMetas. A configuration property could be added to re-enable the old behavior - or even to specify the name of a global function to invoke, or to point at a js file to inline to define that function.
- For the majority of other linkers, this property comes from SelectionScriptLinker.PROCESS_METAS_JS, a protected final field, but this could still have custom logic as described above added.
- In both of the above cases, `__PROCESS_METAS__` is the block to replace in the selection script template.
- Finally, SingleScriptTemplate.js and HostedModeTemplate.js both include these directly, and they cannot be disabled without replacing those template scripts.

[As discussed in the mailing list](https://groups.google.com/g/google-web-toolkit/c/Srwv_QRonXI), CSP rules can prevent the eval() calls from running at all, but it should still be removed.

Additionally, HostedModeLinker is entirely unused (and not even registered as a linker that can be used in module files), and is probably safe to remove as part of https://github.com/gwtproject/gwt/issues/9923.

PR #10405, opened 2026-09-08
Disable legacy metadata error hooks by default

cameroncrippen:
Add `gwt.enableMetaErrorHandlers`, defaulting to `false`. The built-in selection scripts omit the complete `gwt:onLoadErrorFn` and `gwt:onPropertyErrorFn` metadata callback branches before JavaScript optimization. Ordinary `gwt:property` processing remains available. Applications that still use the legacy callbacks can set the option to `true`; invalid values fail generation.

This revises the initial Boolean-guard implementation in response to maintainer feedback. It removes the inactive callback code from generated output instead of relying on an optimizer to eliminate guarded branches. The default change affects applications that currently use those legacy callback tags, including pages shared with older compiled modules; explicit opt-in restores the callbacks for the newly compiled module.

Current local validation on Windows with Java 17: the rebuilt `SelectionScriptJavaScriptTest` suite passed 15 tests with zero failures or errors. The generated-helper cases cover five linkers and exercise ordinary metadata, disabled/default callbacks, explicit opt-in callbacks, module-specific tags and invalid settings. They also assert that the disabled/default helper omits both callback names and their `eval` calls. Contributor compilation, API and style checks completed with exit 0; the API checker retains optional type-resolution diagnostics.

Three real minimal modules also compiled successfully with default, explicit-false and explicit-true configuration using the modern linker, one browser permutation and PRETTY output. Default and false bootstrap files are identical and omit the legacy callback branches. Explicit true reproduces the original unchanged bootstrap exactly.

This targets the legacy metadata callback handling. It does not remove all uses of `eval` or establish full CSP compatibility. Full browser startup with the disabled configuration, native hosted-mode execution and the upstream CI matrix remain unverified. The earlier 36-test LinkerSuite result belongs to the original revision and is not claimed for this update.

AI disclosure: patch preparation, validation and this description used OpenAI Codex assistance. No prior human technical review is claimed. Maintainer review of the default change and coverage is requested.

Partial #9967


niloc132:
Thank you for working on this - can you help me understand why this is needed in this form? As I read/debug the code in question, each of those branches fail anyway, since the meta's name doesn't match those keys. Even if you are controlling property values from meta tags, you are specifying `gwt:property`, so these branches are already disabled. This patch then is just "double disabling" them - CSP won't trigger on these anyway since they won't run. Or are you specifying these in your meta tags, and you need to disable the feature _without_ removing the meta tags?

Am I missing something? As written, this would just prefix the existing checks with `true &&` or `false &&` - better to optionally remove the entire check. A new config prop, like what you added, could drive that - enabled for a release, disabled (with a warning to reenable) for a release, then remove outright.

EDIT: actually, in re-reading the issue I originally wrote, this is undocumented anyway - maybe we skip past the "enabled by default" stage and just disable them.

cameroncrippen:
You're right: the runtime guards left the callback checks and eval source in the generated bootstrap. The revision removes those entire callback branches during selection-script generation. I also changed the default to disabled, with `gwt.enableMetaErrorHandlers=true` retaining an explicit compatibility opt-in. Ordinary `gwt:property` processing is unchanged.

The 15-test offline helper suite passes across all five linker classes, including generated-source absence checks and ordinary metadata/opt-in callback behavior. Three real default, false and true module compilations confirm the modern bootstrap output: default and false are identical, and explicit true reproduces the original unchanged bootstrap. Contributor compilation, API and style checks completed; the API checker retains optional type-resolution diagnostics. Full browser/hosted runtime and upstream CI remain unverified. This is limited to the two metadata callback paths and does not establish complete CSP support. This revision and reply were prepared with OpenAI Codex assistance.

niloc132:
EDIT: I missed that there was a new patch, disregard this message.

niloc132 (approved):

PR #10406, opened 2026-09-09
Inline JS assignment, reference to just assignment

### LINKED ISSUE #10399: JS should optimize `(a=b, a)` expressions, rewrite as `a=b`
The GWT AST JMultiExpression and JS AST JsBinaryOperation with COMMA are used to condense simple statements into an expression, enabling further inlining work. However, from time to time these can emit code like `(a=<expr>, a)` in the JS output, which could be optimized to simply `(a=<expr>)`. The value of both expressions is the same, with a few bytes saved.

This is a fairly easy change for JsStaticEval.trySimplifyComma, with a few simple improvements as a result.

As an example, SimpleEventBus gets this small improvement in the showcase:
```diff
   if (!event) {
     throw toJs(new NullPointerException('Cannot fire null event'));
   }
   try {
     ++this$static.firingDepth;
-    handlers = (directHandlers = $getHandlerList(this$static, event.getAssociatedType(), null) , directHandlers);
+    handlers = directHandlers = $getHandlerList(this$static, event.getAssociatedType(), null);
     causes = null;
     it = this$static.isReverseOrder?handlers.listIterator(handlers.size()):handlers.listIterator();
     while (this$static.isReverseOrder?it.hasPrevious():it.hasNext()) {
       handler = this$static.isReverseOrder?it.previous():it.next();
       try {
```

niloc132:
Cleans up a pattern left by both inliners, while still retaining the assignment for any later usage. Technically this would be better fixed with a variable inlining pass, but GWT doesn't have one of those at this time, and this is a very simple pattern match to use instead.

Adds the requirement that symbols are resolved before running JsStaticEval. This should already be done in most circumstances, added that pass to tests as well.

Partially fixes a size regression from #10398.

Fixes #10399

### DIFF
--- a/dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
+++ b/dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
@@ -716,17 +716,23 @@ public class JsStaticEval {
       return arg1;
     }
 
-    // Likewise, simplify (name++, name) to (name++)
+    // Likewise, simplify (name++, name) or (++name, name) to (++name)
     if (arg1 instanceof JsUnaryOperation unaryOp
         && arg2 instanceof JsNameRef resultRef
-        && unaryOp.getOperator().isModifying()
-        && unaryOp.getArg() instanceof JsNameRef assignRef
+        && (unaryOp.getOperator() == JsUnaryOperator.INC
+            || unaryOp.getOperator() == JsUnaryOperator.DEC)
+        && unaryOp.getArg() instanceof JsNameRef unaryArgRef
         && resultRef.getQualifier() == null
-        && assignRef.getQualifier() == null
-        && assignRef.getName() == resultRef.getName()) {
-      assert assignRef.isResolved() : "assignRef was not resolved: " + assignRef;
+        && unaryArgRef.getQualifier() == null
+        && unaryArgRef.getName() == resultRef.getName()) {
+      assert unaryArgRef.isResolved() : "unaryArgRef was not resolved: " + unaryArgRef;
       assert resultRef.isResolved() : "resultRef was not resolved: " + resultRef;
-      return arg1;
+      if (unaryOp instanceof JsPrefixOperation) {
+        return unaryOp;
+      }
+      // Rewrite the discarded postfix operation into the equivalent prefix operation.
+      return new JsPrefixOperation(unaryOp.getSourceInfo(), unaryOp.getOperator(),
+          unaryOp.getArg());
     }
 
     return expr;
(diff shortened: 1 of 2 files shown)


dev/core/src/com/google/gwt/dev/js/JsStaticEval.java
```diff
@@ -700,6 +701,21 @@ protected static JsExpression trySimplifyComma(JsBinaryOperation expr) {
     if (!arg1.hasSideEffects()) {
       return arg2;
     }
+
+    // Simplify (name = expr, name) to (name = expr), a pattern produced by both java and js method
+    // inlining passes.
+    if (arg1 instanceof JsBinaryOperation op
+        && arg2 instanceof JsNameRef resultRef
+        && op.getOperator() == JsBinaryOperator.ASG
```
zbynek: Should this allow all assigning operators? `(name += expr, name)` is the same as `(name += expr)`, though I guess the number of occurrences is negligible compared to the plain assignment case.

niloc132: That's a good point - at least I never saw regressions related to this, but it would be easy for me to check that.

For that matter, `(a++, a)` etc could have the same thing apply to them.

I'm slightly leaning towards "fix the regression and let's get a real local var inliner", but it would be quite easy to just add a few more checks here.

zbynek: What I meant to suggest was just `op.getOperator().isAssignment()` instead of `op.getOperator() == JsBinaryOperator.ASG`, but you're right that unary cases with `++` and `--` can also be optimized.

niloc132: Yep, I'm in agreement, just think that if it worth doing, its worth expanding to that case too. 

I thought I found a few cases, but updating to latest JsStaticEval improved how StringBuilder is inlined, and that confused my diff at first - that will improve inlining for builders in other cases too.

I found no cases where any assignment or ++/--/delete was improved by this change. I'm going to add tests to be sure there is nothing sketchy about this enhancement, then I'll add them to the patch. It is a cheap change, so I think it is still worth doing, if we can be confident that nothing else will break.

zbynek (approved):
If my numbers are correct, the regression from https://github.com/gwtproject/gwt/pull/10398 on a ~4.7MB codebase was just 89 bytes and this alone saves 184 bytes 👍

niloc132:
Yeah, this also happens to catch a previously ignored issue from java method inlining - my change for 10398 only took away the cleanup for the js inliner, but this is a more general fix that solves both (plus its nice and cheap to check, at least until we put in a more expensive "this is the only write, that's the only read, and there are no conflicting side effects" inliner).

PR #10407, opened 2026-09-10
Prune JS vars that are written but never read

### LINKED ISSUE #10400: JS passes should remove local vars that are written but not read
On the Java side, ControlFlowAnalyzer looks at local variable refs across the program, finding writes with no reads, and removing them. Nothing seems to do this today on the JS side though, which can result in extra local vars that are assigned but unread or never even assigned.

In the example given by https://github.com/gwtproject/gwt/issues/10399, the inlined local variable `directHandlers` is only assigned but never read, and should be entirely removed.

niloc132:
This is somewhat similar to the Java Pruner pass, but specific at this time to local variables in functions. Requires multiple passes of both the local function and any nested functions to ensure we only clean up those that are never read.

With the information this pass is collecting, it could be a good place to perform other future optimizations as well.

Partially fixes a size regression from #10398.

Fixes #10400

### DIFF
--- a/dev/core/src/com/google/gwt/dev/js/JsUnusedVarPruner.java
+++ b/dev/core/src/com/google/gwt/dev/js/JsUnusedVarPruner.java
@@ -1,0 +1,182 @@
+/*
+ * Copyright 2026 GWT Project Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.gwt.dev.js;
+
+import com.google.gwt.dev.jjs.impl.OptimizerStats;
+import com.google.gwt.dev.js.ast.JsBinaryOperation;
+import com.google.gwt.dev.js.ast.JsBinaryOperator;
+import com.google.gwt.dev.js.ast.JsContext;
+import com.google.gwt.dev.js.ast.JsFunction;
+import com.google.gwt.dev.js.ast.JsModVisitor;
+import com.google.gwt.dev.js.ast.JsName;
+import com.google.gwt.dev.js.ast.JsNameRef;
+import com.google.gwt.dev.js.ast.JsProgram;
+import com.google.gwt.dev.js.ast.JsVars;
+import com.google.gwt.dev.js.ast.JsVars.JsVar;
+import com.google.gwt.dev.js.ast.JsVisitor;
+import com.google.gwt.dev.util.collect.IdentityHashSet;
+
+import java.util.Set;
+import java.util.Stack;
+
+/**
+ * Removes variable declarations and write-only assignments for local variables that are never read
+ * within their enclosing function (including nested functions).
+ * <p>
+ * Future work here could regroup all var decls at the top of the function, and collect any
+ * assignment statements into init expressions (reordering as needed to preserve side effects).
+ */
+public class JsUnusedVarPruner {
+  public static final String NAME = JsUnusedVarPruner.class.getSimpleName();
+
+  public static int exec(JsProgram program) {
+    return new JsUnusedVarPruner(program).execImpl();
+  }
+
+  private final JsProgram program;
+
+  private JsUnusedVarPruner(JsProgram program) {
+    this.program = program;
+  }
+
+  private int execImpl() {
+    try (OptimizerStats stats = OptimizerStats.optimization(NAME)) {
+      FunctionVisitor functionVisitor = new FunctionVisitor();
+      functionVisitor.accept(program);
+
+      stats.recordModified(functionVisitor.totalMods);
+
+      return stats.getNumMods();
+    }
+  }
+
+  /**
+   * Finds functions and collects their vars, revisiting to find reads and prune unread vars.
+   * Due to var hoisting and closures, we scan each function twice, once in this visitor to collect
+   * vars in the function itself, then another visitor to find reads of those vars.
+   * <p>
+   * A third pass then removes any writes from all vars that are not read, and attempts to remove
+   * the var declaration itself.
+   */
+  private static class FunctionVisitor extends JsVisitor {
+    private int totalMods = 0;
+    private final Stack<Set<JsName>> namesStack = new Stack<>();
+
+    @Override
+    public boolean visit(JsFunction x, JsContext ctx) {
+      namesStack.push(new IdentityHashSet<>());
+      return super.visit(x, ctx);
+    }
+
+    @Override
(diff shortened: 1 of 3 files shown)


dev/core/test/com/google/gwt/dev/js/JsUnusedVarPrunerTest.java
```diff
@@ -0,0 +1,102 @@
...
+  public void testPreserveVarWithSideEffectInit() throws Exception {
+    // Can't remove because the initializer has side effects, could be improved in the future
+    assertEquals("function f(){var a=foo()}\n",
+        optimize("function f() { var a = foo(); }"));
+  }
+
+  public void testPreserveCompoundAssignment() throws Exception {
+    // Compound assignment reads the var, so it must be preserved.
+    assertEquals("function f(){var a=0;a+=1}\n",
+        optimize("function f() { var a = 0; a += 1; }"));
```
zbynek, line 63: Trivial cases like this can be simplified to `function(){}`, it might be nice to have a test that shows why compound assignments are not safe to remove.
```
function f() { var a = 0; alert(a += 1); alert(a += 1); }
```

niloc132: Even your case, the second alert should become `alert(a+1)` and drop the write.

This is a shortcoming of both java and js DCE steps, see https://github.com/gwtproject/gwt/issues/10286.

I can add your more complex case, but we still want the trivial case too, at least until we solve the issue - the pass must not eliminate writes if there are still reads, and we don't presently work out if a read serves a purpose other than to write.

zbynek:
Compiling GeoGebra with this [gives](https://github.com/zbynek/terser-check/actions/runs/34540237688/job/103081023600)
```
Caused by: java.util.EmptyStackException
	at java.base/java.util.Stack.peek(Stack.java:101)
	at com.google.gwt.dev.js.JsUnusedVarPruner$FunctionVisitor.endVisit(JsUnusedVarPruner.java:86)
	at com.google.gwt.dev.js.ast.JsVars$JsVar.traverse(JsVars.java:67)
	at com.google.gwt.dev.js.ast.JsVisitor.doTraverse(JsVisitor.java:463)
	... 28 more
         [ERROR] at com.google.gwt.dev.js.ast.JsProgram(0): _
            com.google.gwt.dev.js.ast.JsVars$JsVar
         [ERROR] at com.google.gwt.dev.js.ast.JsProgram(0): var _, prototypesByTypeId, initFnList, permutationId = -1
            com.google.gwt.dev.js.ast.JsVars
```

dev/core/src/com/google/gwt/dev/js/JsUnusedVarPruner.java
```diff
@@ -0,0 +1,182 @@
...
+   * Finds functions and collects their vars, revisiting to find reads and prune unread vars.
+   * Due to var hoisting and closures, we scan each function twice, once in this visitor to collect
+   * vars in the function itself, then another visitor to find reads of those vars.
+   * <p>
+   * A third pass then removes any writes from all vars that are not read, and attempts to remove
+   * the var declaration itself.
+   */
+  private static class FunctionVisitor extends JsVisitor {
+    private int totalMods = 0;
+    private final Stack<Set<JsName>> namesStack = new Stack<>();
```
zbynek, line 76: `Stack` is used across the codebase, so it's no big deal either way, but tools including Errorprone suggest to use `Deque` instead https://errorprone.info/bugpattern/JdkObsolete

zbynek (changes requested):
Compiling GeoGebra with this [gives](https://github.com/zbynek/terser-check/actions/runs/34540237688/job/103081023600)
```
Caused by: java.util.EmptyStackException
	at java.base/java.util.Stack.peek(Stack.java:101)
	at com.google.gwt.dev.js.JsUnusedVarPruner$FunctionVisitor.endVisit(JsUnusedVarPruner.java:86)
	at com.google.gwt.dev.js.ast.JsVars$JsVar.traverse(JsVars.java:67)
	at com.google.gwt.dev.js.ast.JsVisitor.doTraverse(JsVisitor.java:463)
	... 28 more
         [ERROR] at com.google.gwt.dev.js.ast.JsProgram(0): _
            com.google.gwt.dev.js.ast.JsVars$JsVar
         [ERROR] at com.google.gwt.dev.js.ast.JsProgram(0): var _, prototypesByTypeId, initFnList, permutationId = -1
            com.google.gwt.dev.js.ast.JsVars
```

niloc132:
Thanks, the error makes sense, though I'm confused about how I failed to hit it with tests and samples. Which linker do you use?

zbynek:
>Which linker do you use?

XSI, with [some overrides](https://github.com/geogebra/geogebra/blob/4f1aeb42aed04c3a13a3e122a9dac0a36e871eac/source/web/web-dev/src/main/java/org/geogebra/web/linker/StandardIframeLinker.java) -- mostly to produce .mjs in parallel to .js

niloc132:
I dont immediately see why that linker would have had this issue, but I put a fix up last night, can you try it out at your convenience?

I think a better fix would be to also prune top-level vars, but I picked a more conservative approach for the moment. I have other enhancements I want to do to this class later, like merging different `var` statements (mostly from JSNI) to save a few bytes per method, and merging early assignments into their var initialization where possible (two bytes per assignment saved, and there are a lot of locals that get written early in methods, so it adds up). We should also be able to rearrange statements to prune vars that have initializers (only JSNI can have those today though).

zbynek:
With the latest state of this branch the codebase I tested compiles and size goes from 4,729,840  to 4,729,728 bytes, so the improvement is again slightly bigger than the regression in #10398.

dev/core/test/com/google/gwt/dev/js/JsUnusedVarPrunerTest.java
zbynek: Having both tests would be nice (one where the copound assignment is read by parent expression, one where it isn't).

zbynek (commented):
With the latest state of this branch the codebase I tested compiles and size goes from 4,729,840  to 4,729,728 bytes, so the improvement is again slightly bigger than the regression in #10398. 

The code looks good, two non-blocking suggestions in other comments.

zbynek:
> I dont immediately see why that linker would have had this issue

The variables causing this problem are either static fields, enum valaues, or class literal holders, e.g.
`java_lang_Boolean_TRUE`
`GLOBAL_VAR org_geogebra_common_awt_GColor_GREEN`
`com_google_gwt_lang_ClassLiteralHolder_Ljava_1lang_1Number_12_1classLit`

niloc132:
> > I dont immediately see why that linker would have had this issue
> 
> The variables causing this problem are either static fields, enum valaues, or class literal holders, e.g. `java_lang_Boolean_TRUE` `GLOBAL_VAR org_geogebra_common_awt_GColor_GREEN` `com_google_gwt_lang_ClassLiteralHolder_Ljava_1lang_1Number_12_1classLit`

Agreed sorry, I should have been more clear - what I mean is that I don't see why that linker should be substantially different than the default linker that the showcase uses, or that the various `user` tests use. I'm still unclear, but the fix should handle any top-level vars now.

zbynek (approved):
Looks good. The failure with empty stack (that's now fixed) would have been found by a test that runs the optimization loop on somewhat reallistic code -- I proposed one in https://github.com/niloc132/gwt/pull/7 , but that's maybe better suited for a separate PR.

